PR #22, opened 2017-05-22
Add support and testing for Python 3.6

jd:


jd:
I just realized we can't do that since Python 3.6 is not in Ubuntu and therefore not in the docker image.

PR #23, opened 2017-05-22
storage: introduce add_measures_batch for Ceph

jd:
This replaces the thread based approach with a faster write operation for the
driver that will support it.
This seems to improve performance by a large magnitude.

(cherry picked from commit d9d472ed70e327f8affeefbd26135e1db2df376c)

### DIFF
--- a/gnocchi/rest/__init__.py
+++ b/gnocchi/rest/__init__.py
@@ -18,7 +18,6 @@ import functools
 import itertools
 import uuid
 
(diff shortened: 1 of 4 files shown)


jd:
@chungg @sileht since it's a big improvement I've been requested to backport this to 3.1. I don't see any issue with it, you're all good too?

sileht:
It's ok for me.

PR #27, opened 2017-05-22
Bump Cradox requirement to 1.2.0

jd:
The `operate_aio_write_op` method is unavailable before that version.


setup.cfg
```diff
@@ -51,7 +51,7 @@ ceph =
     lz4>=0.9.0
     tooz>=1.38
 ceph_recommended_lib =
-    cradox>=1.0.9
+    cradox>=1.2.0
```
apevec: Why not just go for latest 1.3.2 ? Is there any distro stuck at 1.2.0 in particular?

jd: There's no reason to require 1.3.2 if 1.2.0 is API compatible.

PR #31, opened 2017-05-23
Merge request for jd/add-incoming-in-doc

jd:
doc: add incoming configuration option note

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -122,6 +122,10 @@ options you want to change and configure:
 |                     | if you use the Redis storage driver.              |
 +---------------------+---------------------------------------------------+
 
+The same options are also available as `incoming.<drivername>_*` for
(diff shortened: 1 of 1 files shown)


doc/source/install.rst
```diff
@@ -122,6 +122,10 @@ options you want to change and configure:
 |                     | if you use the Redis storage driver.              |
 +---------------------+---------------------------------------------------+
 
+The same options are also available as `incoming.<drivername>_` for configuring
```
chungg: what's the trailing '_' for?

jd: I meant _*, as whatver starts with drivername-underscore. Maybe I should add a * to make it more clear?

chungg: sure, yeah, i thought it just a bad link or something (and i didn't realise we prefix everything). seems best to follow format above and add _*

PR #36, opened 2017-05-23
ceph: simplify reporting

chungg:
backlog is not grouped by common metric keys so it can have hundreds
of thousands to millions of keys to loop through. we shouldn't waste
time by checking details millions of times.

### DIFF
--- a/gnocchi/storage/incoming/ceph.py
+++ b/gnocchi/storage/incoming/ceph.py
@@ -107,6 +107,10 @@ class CephStorage(_carbonara.CarbonaraBasedStorage):
             op.wait_for_complete()
 
     def _build_report(self, details):
+        def inc_details(details, metric):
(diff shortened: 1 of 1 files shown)


jd:
This seems to really do the same thing without an `if'. Did you profile that to be sure it's way faster? It's not obvious and I'm afraid it'll go away first time somebody touches that code.

chungg:
it's probably not obvious. i'm indifferent on whether we merge this. i'm also looking at avoiding building full details so we save memory but that's probably also non-obvious... *shrugs*

your call.

jd:
If you are sure it's improves perf in anyway and if you add a note in the code that this was done on purpose, I'm ok to merge.
Otherwise I'd say let's not do it. :)

chungg:
negligible. less than half a percent improvement. closing it.

PR #37, opened 2017-05-23
reduce memory for slim report

chungg:
- don't store all metrics and counts in memory if not building detailed
report.
- leverage pipelines in redis
- remove s3 FIXME since i don't think it's true.

### DIFF
--- a/gnocchi/storage/incoming/file.py
+++ b/gnocchi/storage/incoming/file.py
@@ -98,13 +98,27 @@ class FileStorage(_carbonara.CarbonaraBasedStorage):
                         raise
 
     def _build_report(self, details):
-        metric_details = {}
+        report_vars = {'metrics': 0, 'measures': 0, 'metric_details': {}}
+        if details:
+            def build_metric_report(metric, sack):
+                report_vars['metric_details'][metric] = len(
+                    self._list_measures_container_for_metric_id_str(sack,
+                                                                    metric))
+        else:
+            def build_metric_report(metric, sack):
+                report_vars['metrics'] += 1
+                report_vars['measures'] += len(
+                    self._list_measures_container_for_metric_id_str(sack,
+                                                                    metric))
+
         for i in six.moves.range(self.NUM_SACKS):
             for metric in self.list_metric_with_measures_to_process(i):
-                metric_details[metric] = len(
-                    self._list_measures_container_for_metric_id_str(i, metric))
-        return (len(metric_details.keys()), sum(metric_details.values()),
-                metric_details if details else None)
+                build_metric_report(metric, i)
+        return (report_vars['metrics'] or
+                len(report_vars['metric_details'].keys()),
+                report_vars['measures'] or
+                sum(report_vars['metric_details'].values()),
+                report_vars['metric_details'] if details else None)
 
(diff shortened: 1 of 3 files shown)


gnocchi/storage/incoming/redis.py
```diff
@@ -49,13 +48,27 @@ def _store_new_measures(self, metric, data):
...
+        measures = 0
         match = redis.SEP.join([self.get_sack_name("*"), "*"])
-        metric_details = collections.defaultdict(int)
-        for key in self._client.scan_iter(match=match, count=1000):
-            metric = key.decode('utf8').split(redis.SEP)[1]
-            metric_details[metric] = self._client.llen(key)
-        return (len(metric_details.keys()), sum(metric_details.values()),
-                metric_details if details else None)
+        metric_details = {}
+        cursor = '0'
```
jd: '0' ?

gnocchi/storage/incoming/redis.py
```diff
@@ -49,13 +48,27 @@ def _store_new_measures(self, metric, data):
...
+        while cursor != 0:
```
jd: any reason you replace `scan_iter`? I found it more Pythonic. :)

chungg: i wanted to leverage pipeline and i don't want to pipe 1 million commands to redis and then get a million item list in return. also, didn't want to add a check every loop to decide whether to chunk or not, although i guess this relates back to other patch and can test how much a million single if checks adds to workload

gnocchi/storage/incoming/redis.py
chungg: stolen from source[1] but basically redis returns '0' as cursor when it hits end of list[2]

[1] https://github.com/andymccurdy/redis-py/blob/d6c300b39ab4c81cd97b81fcd8d25a9cde9476e7/redis/client.py#L1469-L1482
[2] https://redis.io/commands/scan

gnocchi/storage/incoming/redis.py
chungg: i'll add check instead... need to figure out why swift report is wrong also.

gnocchi/storage/incoming/file.py
```diff
@@ -98,13 +98,26 @@ def _store_new_measures(self, metric, data):
...
+        def detailed_report(metric, sack):
+            report_vars['metric_details'][metric] = len(
+                self._list_measures_container_for_metric_id_str(sack, metric))
+
+        def short_report(metric, sack):
+            report_vars['metrics'] += 1
+            report_vars['measures'] += len(
+                self._list_measures_container_for_metric_id_str(sack, metric))
+
+        build_metric_report = detailed_report if details else short_report
```
jd: you could avoid an entire `def` here:

```python
if details:
    def build_metric_report(metric, sack):
        report_vars['metrics'] += 1
        report_vars['measures'] += len(self._list_measures_container_for_metric_id_str(sack, metric))
else:
    def build_metric_report(metric, sac):
        report_vars['metric_details'][metric] = len(self._list_measures_container_for_metric_id_str(sack, metric))
```

PR #46, opened 2017-05-25
catch error trying to change sack on pre-gnocchi4 storage

chungg:
if storage hasn't been upgraded to gnocchiv4, it will fail with
obscure message. a TypeError is because storage hasn't been upgraded.


gnocchi/cli.py
```diff
@@ -83,7 +83,16 @@ def change_sack_size():
     ])
     conf = service.prepare_service(conf=conf)
     s = storage.get_driver(conf)
-    report = s.incoming.measures_report(details=False)
+    try:
+        report = s.incoming.measures_report(details=False)
+    except TypeError:
```
jd: Good idea, but the implementation is too vague. We need a proper exception caught defined by gnocchi.storage. TypeError could be raised by more or less anything, especially when we'll modify the code.

chungg: done, i think

gnocchi/cli.py
```diff
@@ -86,7 +86,11 @@ def change_sack_size():
     ])
     conf = service.prepare_service(conf=conf)
     s = storage.get_incoming_driver(conf.incoming)
-    report = s.measures_report(details=False)
+    try:
+        report = s.measures_report(details=False)
+    except Exception:
+        # issue is already logged by measures_report
```
jd: where? I skimmed through the code but I see nowhere it is handled. This should be handled here to me with a proper exception raised by the underlying driver(s).

chungg: here: https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/storage/incoming/_carbonara.py#L47-L48

basically what's happening is if we didn't upgrade, it can't find NUM_SACKS value. all the drivers will return back None here https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/storage/incoming/_carbonara.py#L45 which causes int(None) failure.

i'm not sure what you want me to catch here or do you want me to raise an error instead of having drivers return None? and if so, do you want me to still log here: https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/storage/incoming/_carbonara.py#L45 or let it raise and have to catch it everywhere it's used?

jd: On https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/storage/incoming/_carbonara.py#L49

Change by `raise NumSackReadingError(e)` (or any better name, sorry)

and in this patch, line 91 just change `except Exception` by `except NumSackReadingError`. That way you're sure you're only ignoring this for a good reason.

WDYT?

chungg: sure

PR #48, opened 2017-05-29
archive_policy: Raise Error if calculated points is < 0

sum12:
In case granularity is larger than timespan, then the calculated points
are < zero, and since there was no validation of this value an invalid
policy would be created.
This commit fixes that behaviour.


sum12:
fixes #48

sjamgade:
@jd  done. :)

sum12:
@jd done :)

PR #49, opened 2017-05-29
doc: fox uwsgi example

jd:
doc: fox uwsgi example

### DIFF
--- a/doc/source/running.rst
+++ b/doc/source/running.rst
@@ -23,7 +23,7 @@ machines.
 The following uwsgi configuration file can be used::
 
   [uwsgi]
-  http = localhost:8041
+  http-socket = localhost:8041
   # Set the correct path depending on your installation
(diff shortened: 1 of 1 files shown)


doc/source/running.rst
```diff
@@ -23,7 +23,7 @@ machines.
 The following uwsgi configuration file can be used::
 
   [uwsgi]
-  http = localhost:8041
```
chungg: this is actually valid... it's what i have in my deployment.

jd: what version of uwsgi do you have?
got 2.0.15 here

chungg: 2.0.14, according to docs, http makes uwsgi act similar to nginx/apache? http://uwsgi-docs.readthedocs.io/en/latest/ThingsToKnow.html

jd: Ok, so I guess it's a compile option missing on my install or something. I'll abandon.

PR #54, opened 2017-05-30
auth: Add support for REMOTE_USER env

sileht:


gnocchi/opts.py
```diff
@@ -41,6 +41,7 @@ def __getitem__(self, key):
             return ''
         return value
 
+
```
jd: okay?

sileht: My editor fixes it ! I wonder why my pep8 plugin detect that and not tox...

sileht:
pecan.request is a really fancy interface to the environment variables, it's the webob.Request object, is it ok to pass this to the plugin ? I really don't want to deal with environ[XXXX], it's a trap ! 

Each var have a different RFC/encoding/sometime bytes/sometime str, Webob hides all of this crap. Even if we switch to another wsgi framework or move to an non wsgi (asyncio) framework. nobody will reinvent Webob and just use it.

chungg:
stupid question, but what's a REMOTE_USER auth for?

jd:
@chungg when web servers e.g. Apache HTTPD authenticate someone with an external module (e.g. mod_ldap) they set this env var to the user that has been authenticated.

sileht:
I have forget git add...

sileht:
Pull-request updated, HEAD is now 9360c523d8acee263f2436fea50b355ac1c489bc

sileht:
Pull-request updated, HEAD is now a213a4132d86efea643b748cf9b133082643d8bb

jd:
@sileht I've merged master but tests actually failed for probably a good reason.

sileht:
Pull-request updated, HEAD is now 8e8084a560a1bd21517ed6dff353d2efb7eb21e2

sileht:
Pull-request updated, HEAD is now 8e8084a560a1bd21517ed6dff353d2efb7eb21e2

jd:
You can do this @sileht

sileht:
Pull-request updated, HEAD is now 176745cbf8c4a005e782d1581659c919bca2cbb3

PR #55, opened 2017-05-31
Return new metrics when creating metrics for a resource

### LINKED ISSUE #12: rest call for appending metrics to a resource does not return any result
See https://bugs.launchpad.net/gnocchi/+bug/1640942

BigChief45:
Fixes #12

### DIFF
--- a/gnocchi/rest/__init__.py
+++ b/gnocchi/rest/__init__.py
@@ -641,7 +641,7 @@ class NamedMetricController(rest.RestController):
         else:
             abort(404, indexer.NoSuchResource(self.resource_id))
 
-    @pecan.expose()
+    @pecan.expose('json')
     def post(self):
         resource = pecan.request.indexer.get_resource(
             self.resource_type, self.resource_id)
@@ -650,8 +650,10 @@ class NamedMetricController(rest.RestController):
         enforce("update resource", resource)
         metrics = deserialize_and_validate(MetricsSchema)
         try:
-            pecan.request.indexer.update_resource(
-                self.resource_type, self.resource_id, metrics=metrics,
+            r = pecan.request.indexer.update_resource(
+                self.resource_type,
+                self.resource_id,
+                metrics=metrics,
                 append_metrics=True,
                 create_revision=False)
         except (indexer.NoSuchMetric,
@@ -662,6 +664,8 @@ class NamedMetricController(rest.RestController):
             abort(409, e)
         except indexer.NoSuchResource as e:
             abort(404, e)
+
+        return r.metrics
 
     @pecan.expose('json')
     def get_all(self):
--- a/gnocchi/tests/test_rest.py
+++ b/gnocchi/tests/test_rest.py
@@ -978,7 +978,7 @@ class ResourceTest(RestTest):
         metrics = {'foo': {'archive_policy_name': "high"}}
         self.app.post_json("/v1/resource/" + self.resource_type
                            + "/" + self.attributes['id'] + "/metric",
-                           params=metrics, status=204)
+                           params=metrics, status=200)
         metrics = {'foo': {'archive_policy_name': "low"}}
         self.app.post_json("/v1/resource/" + self.resource_type
                            + "/" + self.attributes['id']
@@ -999,7 +999,7 @@ class ResourceTest(RestTest):
         metrics = {'foo': {'archive_policy_name': "high"}}
         self.app.post_json("/v1/resource/" + self.resource_type
                            + "/" + self.attributes['id'] + "/metric",
-                           params=metrics, status=204)
+                           params=metrics, status=200)
         result = self.app.get("/v1/resource/"
                               + self.resource_type + "/"
                               + self.attributes['id'])
--- a/gnocchi/tests/functional/gabbits/history.yaml
+++ b/gnocchi/tests/functional/gabbits/history.yaml
@@ -139,7 +139,10 @@ tests:
       data:
         foobar:
           archive_policy_name: low
-      status: 204
+      status: 200
+      response_json_paths:
+          $[/name][1].name: foobar
+          $[/name][1].resource_id: f93450f2-d8a5-4d67-9985-02511241e7d1
 
     - name: list all resources with history no change after metrics creation
       GET: /v1/resource/generic
--- a/gnocchi/tests/functional/gabbits/resource.yaml
+++ b/gnocchi/tests/functional/gabbits/resource.yaml
@@ -667,11 +667,14 @@ tests:
           x-user-id: 0fbb231484614b1a80131fc22f6afc9c
           x-project-id: f3d41b770cc14f0bb94a1d5be9c0e3ea
           content-type: application/json
-      status: 204
+      status: 200
       data:
           electron.spin:
               archive_policy_name: medium
       response_headers:
+      response_json_paths:
+          $[/name][1].name: electron.spin
+          $[/name][1].resource_id: 85c44741-cc60-4033-804e-2d3098c7d2e9
 
     - name: post metric at generic with empty definition
       POST: $LAST_URL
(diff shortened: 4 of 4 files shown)


gnocchi/rest/__init__.py
```diff
@@ -665,6 +665,8 @@ def post(self):
         except indexer.NoSuchResource as e:
             abort(404, e)
 
+        return metrics
```
sileht: We should not return the user input, but the values that come from the database. Also we update a resource, so we should return a resource description, not the metrics (the resource description should have the complete metrics info).

Maybe just adding 'return' at the beginning of line 655 is enough.

BigChief45:
Are you referring to the [`test_post_append_metrics`](https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/tests/test_rest.py#L985) test?

sileht:
On Wed, May 31, 2017 at 02:01:00AM -0700, Andrés Álvarez wrote:
>Are you referring to the [`test_post_append_metrics`](https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/tests/test_rest.py#L985) test?

Not only, you can looks at travis-ci job results to see all the failing
tests.

-- 
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

gnocchi/rest/__init__.py
```diff
@@ -665,6 +665,9 @@ def post(self):
         except indexer.NoSuchResource as e:
             abort(404, e)
 
+        return pecan.request.indexer.get_resource(
```
jd: If the user does not need this output it's a whole request for nothing.
It might  be better to leave the user request it if it needs it after, no?

gnocchi/tests/test_rest.py
```diff
@@ -989,7 +989,7 @@ def test_post_append_metrics(self):
         metrics = {'foo': {'archive_policy_name': "high"}}
         self.app.post_json("/v1/resource/" + self.resource_type
                            + "/" + self.attributes['id'] + "/metric",
-                           params=metrics, status=204)
+                           params=metrics, status=200)
```
jd: This needs to be in the previous patch IIUC.

gnocchi/rest/__init__.py
BigChief45: Do you mean that the user should somehow specify if the output is needed? Could you show me an example of this?

jd: Not really. What I mean is that if the user wants the new updated resource he just have to do `GET /v1/resource/generic/<id>`.

And actually returning the entire resource on the `/v1/resource/generic/<id>/metric` seems wrong to me. This is not the resource endpoint, it's the metric of the resource endpoint. So if anything would be returned it's the new list of metric only – which would be a different SQL request, not sure there's such a call in the indexer API.


Also I did understand at first read you were trying to fix #12. I'd like to have @chungg opionion since he is the one that opened the bug, if this is really a requirement since it will cost an indexer request on each call and I'm not sure it's really worth it.

BigChief45: I agree. 

On my first commit before rebasing I was only returning the newly created metrics. But @sileht mentioned something about returning the whole resource's description, which would return the new metrics as well.

Waiting on more information to know how to proceed with this.

sileht:
>On my first commit before rebasing I was only returning the newly created metrics. But @sileht mentioned something about returning the whole resource's description, which would return the new metrics as well.

Sorry for that, I have thought this API endpoint was about resource, but it's not.

-- 
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

BigChief45:
Any consensus on what the response content should be? Or if this change should be abandoned.

jd:
Re-reading this, I think your first patch was correct – @sileht misread the problem as he said after.

You should return the newly metrics that are created, so "return metrics" was the right choice.

But, as @sileht said you can't blindly return "metrics" because it's user-supplied definitions, and not metrics objects from the database.

`update_resource()` returns the resource with all its metric loaded. So I'd suggest actually returning `r.metrics` at the end of the function. That'll report all the metrics attached to this resource, which considering the endpoint URL, sounds like a good thing.

BigChief45:
@jd Understood. I have made the changes. Will squash if everything is ok.

sileht:
The solution looks good, maybe we should check the returned json with gabbi in additional to the 204 -> 200 change to make this PR perfect.

BigChief45:
@sileht I have updated the Gabbi tests to check for the JSON response. Please let me know if these are correct and if they are enough.

jd:
That's cool, though we need to squash all of that in a single commit. Can you do that @BigChief45? Otherwise I'll try that big UI button in GitHub. :)

BigChief45:
@jd No problem. But I was thinking maybe you could try that big button this time? I've actually never seen it in action myself hehe.

jd:
@BigChief45 Maybe next time, since you'll need to edit that branch to fix the Travis failures. ;)

BigChief45:
@jd No problem. Fixing this soon.

By the way, I am taking a look at the documentation regarding this:

image

I cannot find the response content/scenario anywhere in the docs. Is this content generated into the docs from an actual API call?

jd:
@BigChief45 Yes, the whole API doc is generated from real API calls. Check out https://github.com/gnocchixyz/gnocchi/blob/master/doc/source/rest.yaml and https://github.com/gnocchixyz/gnocchi/blob/master/doc/source/rest.j2

BigChief45:
```
: Unable to match $[-1].resource_id as 85C44741-CC60-4033-804E-2D3098C7D2E9, got 85c44741-cc60-4033-804e-2d3098c7d2e9
```

Hm why are resource IDs being passed in uppercase in the Gabbi tests? Seems a bit strange.

jd:
@BigChief45 Likely because whoever wrote that put them that way. It does not matter and the case should be preserved, so just upper case your changes and you should be all set.

gnocchi/rest/__init__.py
```diff
@@ -665,6 +667,8 @@ def post(self):
         except indexer.NoSuchResource as e:
             abort(404, e)
 
+        return r.metrics
```
chungg: i think this will return all metrics in resource not just the new metrics you added. it's failing in gate in gate because the new metric is not necessarily the first item returned.

if you're testing this locally, you can add verbose: true to gabbi tests, so check what order metrics are returned in (i'm guessing alphabetical). something like:

     - name: create new metrics
       verbose: True
       POST: /v1/resource/generic/f93450f2-d8a5-4d67-9985-02511241e7d1/metric
       request_headers:
           x-user-id: 0fbb231484614b1a80131fc22f6afc9c
           x-project-id: f3d41b770cc14f0bb94a1d5be9c0e3ea
           content-type: application/json
        data:
          foobar:
            archive_policy_name: low
       status: 200
       response_json_paths:
           $[0].name: foobar
           $[0].resource_id: f93450f2-d8a5-4d67-9985-02511241e7d1

BigChief45: This issue is driving me nuts. Locally, all the gabbi tests pass, assuming that the metric being created in the test will be the first one to be returned in the list.

When testing using the Gnocchi API, metrics are not returned in alphabetical order, but in the order of creation (last created, first in the list). But I don't get why it is not the case in the gate.

BigChief45:
So it seems that the order of the returned metrics differs between Postgresql and MySQL?

gnocchi/rest/__init__.py
jd: There's no order set by default. So, *by chance* it might always been the same order, but it also can depends on whetever it runs on MySQL or PostgreSQL. You can't assume in the test that the order is going to be always the same. Especially because maybe the next release of $RDBMS will break that assumption.

Now, what you can do is sort in the `response_json_paths`. This is documented here:

http://gabbi.readthedocs.io/en/latest/jsonpath.html

Basically you can do this:

```
response_json_paths:
  $[/name][1].name: foobar
```

If your metric list is "foobar, cpu_util" for example (you'll need to check what the full list is, then "foobar" will be alphabetically after "cpu_util" when sorting by name (the `[/name]` in the JSON path expression.

BigChief45:
Seems like there is some issue with Docker in the CI.

jd:
That should be fixed now.

BigChief45:
I have squashed all the commits and all seems to pass at the gate. I think this is ready to be merged? :smiley:

jd:
Thanks @BigChief45 !

PR #62, opened 2017-06-02
Remove catching of non existent ArchivePolicyInUse exception

BigChief45:
A non existent `ArchivePolicyRuleInUse` exception is catched when deleting an archive policy rule in the `ArchivePolicyRulesController` controller:

```python
    @pecan.expose()
    def delete(self, name):
        # NOTE(jd) I don't think there's any point in fetching and passing the
        # archive policy rule here, as the rule is probably checking the actual
        # role of the user, not the content of the AP rule.
        enforce("delete archive policy rule", {})
        try:
            pecan.request.indexer.delete_archive_policy_rule(name)
        except indexer.NoSuchArchivePolicyRule as e:
            abort(404, e)
        except indexer.ArchivePolicyRuleInUse as e:
            abort(400, e)
```

This PR adds this missing exception.

### DIFF
--- a/gnocchi/rest/__init__.py
+++ b/gnocchi/rest/__init__.py
@@ -355,8 +355,6 @@ class ArchivePolicyRulesController(rest.RestController):
             pecan.request.indexer.delete_archive_policy_rule(name)
         except indexer.NoSuchArchivePolicyRule as e:
             abort(404, e)
-        except indexer.ArchivePolicyRuleInUse as e:
-            abort(400, e)
 
 
 def MeasuresListSchema(measures):


BigChief45:
In that case ,should the change be squashed into one commit?

jd:
@BigChief45 yeah this is just one change, so just amend your commit :)

BigChief45:
Done :)

By the way, can anyone explain why the indexer cannot raise this exception?

jd:
@BigChief45 This was copy/pasted from the archive policy CRUD operation I think, that's why it was there. When you delete an archive policy, it can be "in use" by metrics or archive policy rules, so it's not deletable.

The archive policy rules are never referenced by anything, they're just rules applied on the fly when creating metrics, so raising such an error make no sense. You can always delete an archive policy rule, without this kind of error.

BigChief45:
@jd Understood. Thanks for the explanation.

PR #65, opened 2017-06-02
Fix doc build

jd:
Fix doc build


jd:
This patch backported without issue in stable/3.1 so what's going on? Any idea @sileht?

jd:
Ok what happens is that the git tag 3.1.0 was not fetched by git. Why? I've no clue. I rechecked the job and it worked.

PR #66, opened 2017-06-02
reuse storage coordinator

chungg:
the processing worker creates it's own connection to coordinator
to handle hashring distribution. this is unnecessary as the storage
driver already has a coordinator for sack_locking which is exactly
what the processing worker's coordinator is ultimately for. reuse
the same coordinator since if the storage coordinator is down, the
processing worker is ultimately down as it's the exact same target.


chungg:
hmm... that creates another issue regarding who stops the coordinator? if we pass in a coordinator we will need to assume storage stops it? or do we close both places and catch the toozerror if we stop an already stopped coordinator?

jd:
That was just a suggestion, I didn't think it completely through. I don't think catching the error is a good design lol.

If the storage driver gets an external coordinator, it can just store that fact and not take care of starting/stopping it?

jd:
pep8 failure

chungg:
fffffffffffffffffffffffffffffffff

PR #72, opened 2017-06-05
incoming-only connection

chungg:
for changing sacks and reporter, there is no need to connect to main
storage and initialise a new coordinator. change so we only connect
to incoming driver and do not create a coordinator connection.

### DIFF
--- a/gnocchi/storage/__init__.py
+++ b/gnocchi/storage/__init__.py
@@ -146,6 +146,14 @@ def get_driver_class(namespace, conf):
                                 conf.driver).driver
 
 
+def get_incoming_driver(conf):
+    """Return configured incoming driver only
+
+    :param conf: incoming configuration only (not global)
+    """
+    return get_driver_class('gnocchi.incoming', conf)(conf)
+
+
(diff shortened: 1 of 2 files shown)


gnocchi/storage/__init__.py
```diff
@@ -146,6 +146,14 @@ def get_driver_class(namespace, conf):
                                 conf.driver).driver
 
 
+def get_incoming_driver(conf):
+    """Return configured incoming driver only
+
+    :param conf: incoming configuring only (not global)
+    """
+    return get_driver_class('gnocchi.incoming', conf)(conf)
```
sileht: Shouldn't be conf.incoming ?

chungg: hmm.. i was going to to make conf param already be conf.incoming but it seems i forgot. i can either pass in whole conf and use conf.incoming here or pass in just conf.incoming from cli... any preference?

jd: Nah I think it's good as it is.

PR #74, opened 2017-06-05
storage: fix resample on empty metric

### LINKED ISSUE #69: measures aggregation: fill and granularity does not work as expected
Hi,

When we try to extract aggregated metrics, from resources that do not have any metrics for a given time range, gnocchi returns an  Internal Server Error  (HTTP 500)

 - [Tracebug](https://gist.github.com/weehal/3e8a4b76ffce8f972c8d880671e597d0)
 - [Steps to reproduce](https://gist.github.com/weehal/878853ac51a633d474ff68b6ae5d54c6)
 - [Some additional tests](https://gist.github.com/weehal/438bf73c86a287ee04a2e379555711be)

jd:
If a metric is empty, carbonara raises an IndexError when checking for the data
on grouping since the Pandas timeseries is empty.

Fixes #69

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -246,7 +246,7 @@ class TimeSerie(object):
         # NOTE(jd) Our whole serialization system is based on Epoch, and we
         # store unsigned integer, so we can't store anything before Epoch.
         # Sorry!
-        if self.ts.index[0].value < 0:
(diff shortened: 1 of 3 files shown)


jd:
Pull-request updated, HEAD is now 5fb48769440cc405d375d7721fecf82e0dd5610f

PR #78, opened 2017-06-06
sqlalchemy: be explicit about the PostgreSQL driver

### LINKED ISSUE #56: PostgreSQL driver should set the driver
```
WARNING oslo_db.sqlalchemy.engines: URL postgresql://localhost/postgres?host=/var/folders/7k/pwdhb_mj2cv4zyr0kyrlzjx40000gq/T/tmpzbm8cE&port=9824 does not contain a '+drivername' portion, and will make use of a default driver.  A full dbname+drivername:// protocol is recommended.
```

jd:
SQLAlchemy now logs a warning if the driver is not correctly set in the URL.

Fixes #56

### DIFF
--- a/gnocchi/indexer/sqlalchemy.py
+++ b/gnocchi/indexer/sqlalchemy.py
@@ -275,6 +275,10 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
             url = sqlalchemy_url.make_url(url)
             url.drivername = "mysql+pymysql"
             return str(url)
+        if url.startswith("postgresql://"):
+            url = sqlalchemy_url.make_url(url)
+            url.drivername = "postgresql+psycopg2"
+            return str(url)
         return url
 
     def __init__(self, conf):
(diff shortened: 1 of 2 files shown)


jd:
Pull-request updated, HEAD is now f351f63ee5fb92f70c823af17dd37ec5db94be7c

jd:
@chungg No we don't, unless we want people to be able to specify their own drivers – which is a bad idea. The `mysql+pymysql` entry points was added because in the OpenStack world (and devstack IIRC) people are used to use it as an entry since most OpenStack projects don't specify the mysql driver by default.

Now that SQLAlchemy wants the app to specify the driver, we do it, but there's no need to allow the user to specify it – especially because it's always going to be our pick, we only support psycopg2 :)

chungg:
@jd word. well fix up the tests then :)

jd:
Pull-request updated, HEAD is now b15a27c5d4a442ab414f8a05a8479a02d6e5fca5

jd:
Pull-request updated, HEAD is now 14a6b6a2a6b30bea11de438153784a8692002141

jd:
Pull-request updated, HEAD is now ce7a12f5dc440eebdf6b0b5c832b6ab60af46c4a

jd:
Pull-request updated, HEAD is now ce7a12f5dc440eebdf6b0b5c832b6ab60af46c4a

jd:
Pull-request updated, HEAD is now 375df41a89da16bc1e166c7fda7a1b7d9e18eb61

jd:
I finally gave up. It's simpler to add the psycopg2 entry in setup.cfg for the tests. It won't change much for the end user anyway, since we never asked them to pick a SQLAlchemy driver. :)

PR #80, opened 2017-06-06
Remove devstack gate support

jd:
This is not maintained nor gate against anymore. No OpenStack project can
consume this since Gnocchi is now hosted on GitHub. It does not make much sense
to keep this code laying around.


jd:
@chungg Hm, considering Gnocchi is mainly used by Ceilometer and that Ceilometer install Gnocchi in devstack, I don't really see the point of keeping this "dead code" here?

chungg:
i don't necessarily require it to be in this repo but i would like to be able to add to my local.conf "enable_plugin gnocchi ..." and it works. i use devstack to do profiling and i tell others to use it when they have trouble getting ceilometer+gnocchi to work. ie. look at the configuration devstack does, compare it to what you have in your environment.

jd:
1. If you want to setup Gnocchi itself to do profiling, I don't think you need devstack (which means OpenStack), you just need a 5 lines script that setups Gnocchi?

2. I don't get the ceilometer+gnocchi troubleshooting. Ceilometer is setting up Gnocchi so that's where they are supposed to look for, not there, no?

chungg:
@jd well i usually do it as a whole to simulate an openstack workflow which is rather periodic. regarding troubleshooting, there's ceilometer connection to gnocchi but also gnocchi connection to keystone and just basic configuration for apache/uwsgi. regardless if it's a worse path, the main point is we don't shouldn't make things more difficult for openstack users/devs. i'm more in touch with the openstack workflow than you guys :p

jd:
Okay, I don't really mind leaving it. I'd suggest adding a Travis job though because I don't think it's currently tested anywhere.
I'll update the PR to only delete the devstack gate stuff.

jd:
Pull-request updated, HEAD is now 8c2096372b55ef179c9d1a6f0e559c528b6ec546

jd:
Pull-request updated, HEAD is now e4f4ac0a4380c4c10e042b05401862d1fcf057ac

jd:
Pull-request updated, HEAD is now 1b5993f25269291315c6641165a7b8f86282feea

jd:
Pull-request updated, HEAD is now 41ed02c0e3b45286523ed141e1a094f70fc6a21d

jd:
Pull-request updated, HEAD is now 36f2b695206333defb7b718c7968c71b4651e36f

jd:
Pull-request updated, HEAD is now 60be1b141a780f3af19ac30f6bd3a6f1a806aff7

PR #85, opened 2017-06-06
travis: fetch all refs for docs

sileht:
Since 464cfc9f8469fc7cea4d6987603775009f6cab16 we change the fetcher
configuration to build the doc of stable branch from master.

But that doesn't work when we build the doc from stable branch.

This change should fix that by fetching all refs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,11 +25,13 @@ before_script:
   # Travis We need to fetch all tags/branches for documentation target
   - case $TARGET in
       docs*)
-        git config --add remote.origin.fetch +refs/heads/stable/*:refs/remotes/origin/stable/*;
+        git config --get-all remote.origin.fetch;
+        git config --unset-all remote.origin.fetch;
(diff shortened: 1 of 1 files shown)


sileht:
Pull-request updated, HEAD is now 9daefb785d577c362cc27b68a7a360399891290e

sileht:
Pull-request updated, HEAD is now 716eca24cd0b13567629cc1360c32ace8ebb3c35

sileht:
Pull-request updated, HEAD is now 5d7bdbbe6f1e8b4279dd1c8f7f37788aad882596

chungg:
i see this in doc logs:

2017-06-07 14:41:31.555 180 INFO reno.config [-] no configuration file in /home/tester/src/releasenotes/config.yaml: IOError: [Errno 2] No such file or directory: '/home/tester/src/releasenotes/config.yaml'

it's a INFO level log so i assume it's safe?

jd:
Yeah I don't think that's new (on our side) we never had any config for reno. It works out of the box!

PR #89, opened 2017-06-07
remove ignore_too_old_timestamps path

chungg:
there is no workflow that actually accepts too old timestamps. the
only workflow that uses the alternate path is test case. i'm not
sure what the use case is that we would allow someone to bypass
the archive policy definition.


chungg:
hm... I think I deleted a good test. let's pause for now.

chungg:
thanks. added back the ignore timestamp test.

PR #94, opened 2017-06-08
doc: add git base override for devstack installation

jd:


jd:
@chungg IIUC this is needed to clone from github.com rather than opentack.org as the URL in the `enable_plugin` is useless. So either this patch is phony, either you're using the old Gnocchi repository.

But I've no clue I don't use devstack. I'll @sileht confirm or not. :p

chungg:
i think i manually cloned gnocchi from github and then added enable_plugin... maybe this is useful for gnocchi clone step

jd:
yeah it's for cloning, you cheater!

chungg:
i deleted local clone and it tried stack.sh with `enable_plugin gnocchi https://github.com/chungg/gnocchi
` and it seemed to clone fine... i'm using devstack with commit 9b335e2f71650536217d4e1cabe093fd354c2756

sileht:
the url for enable_plugin is useless only in gate because gate checkout gits in /opt/stack before running devstack.

jd:
Cool, not needed, thanks guys for testing and confirming! ❤️

PR #95, opened 2017-06-08
tests: remove half of tests

sileht:
We are running gabbi tests twice with two differents URL
just to check header Location.

This change just keeps the one with a prefix.


sileht:
Pull-request updated, HEAD is now c3c6eb7c44ee2457cfb711ca828fea654de40613

sileht:
It's not often I have the occasion to be able to have such title, so I jump :)

PR #100, opened 2017-06-09
tests: speed-up data generation

jd:
Calling 28800 times `date' is pretty slow. Let's generate that data in a snap
using Python.

### DIFF
--- a/run-upgrade-tests.sh
+++ b/run-upgrade-tests.sh
@@ -3,8 +3,6 @@ set -e
 
 export GNOCCHI_DATA=$(mktemp -d -t gnocchi.XXXX)
 
-GDATE=$((which gdate >/dev/null && echo gdate) || echo date)
-
 old_version=$(pip freeze | sed -n '/gnocchi==/s/.*==\(.*\)/\1/p')
 
 RESOURCE_IDS=(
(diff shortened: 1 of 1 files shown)


jd:
Pull-request updated, HEAD is now f7511912ab8b41fc76ecb8868aa32eaba06433fc

jd:
Pull-request updated, HEAD is now d84069e0166299e5db0f1c16cd2a584e72dacd43

tox.ini
```diff
@@ -48,7 +48,7 @@ commands =
     {toxinidir}/run-tests.sh {posargs}
     {toxinidir}/run-func-tests.sh {posargs}
 
-[testenv:py35-postgresql-file-upgrade-from-3.1]
+[testenv:py36-postgresql-file-upgrade-from-3.1]
```
chungg: i should ask why you change to 3.6 right? i'll ask why... why you change to 3.6 if everything else is 3.5? :)

jd:
Pull-request updated, HEAD is now cc1a6a1db7dbf455898692cc779c9cdc7524df6d

jd:
Pull-request updated, HEAD is now 64e9f56c0a62d6e327e19de5203ffeaa2792ccf9

tox.ini
jd: Good catch. I needed to change it to test in my env as I don't have Python 3.5 anymore. But I did not check while committing!

Thanks @chungg !

PR #102, opened 2017-06-12
Remove oslotest dependency

jd:
Oslotest is OpenStack specific and we actually don't use any of its feature
except stdout redirection and log silencing, which is really provided by
`fixtures' and `daiquiri'.
Remove it and use testtools directly. This avoids having OS_* variables used
everywhere and makes sure we don't pull dependencies we don't need (mox3,
os-client-config, etc).

### DIFF
--- a/gnocchi/tests/base.py
+++ b/gnocchi/tests/base.py
@@ -20,10 +20,8 @@ import subprocess
 import threading
 import uuid
 
+import daiquiri
 import fixtures
-from oslotest import base
-from oslotest import log
-from oslotest import output
 import six
 from six.moves.urllib.parse import unquote
 try:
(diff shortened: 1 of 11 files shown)


jd:
Pull-request updated, HEAD is now 6558528e4104f80bf17433f55e96429e2b538ed8

jd:
@sileht would be glad to have your review too so we're sure I did not miss anything :)

gnocchi/tests/base.py
```diff
@@ -175,8 +173,40 @@ def post_account(self, headers, query_string=None, data=None,
...
+        self._stderr_fixture = fixtures.StringStream('stderr')
+        self.stderr = self.useFixture(self._stderr_fixture).stream
+        self.useFixture(fixtures.MonkeyPatch('sys.stderr', self.stderr))
+
+
+class BaseTestCase(testcase.TestCase):
+    def setUp(self):
+        super(BaseTestCase, self).setUp()
+        self.useFixture(CaptureOutput())
+        daiquiri.setup(outputs=[])
```
sileht: I use OS_XXX var to debug, so having an environment variable to disbale CaptureOuput() and setup
daiquiri with stdout or stderr on debug level would be cool.

Like GNOCCHI_DEBUG_TESTS

gnocchi/tests/base.py
```diff
@@ -234,14 +264,6 @@ class TestCase(base.BaseTestCase):
     @classmethod
     def setUpClass(self):
         super(TestCase, self).setUpClass()
-
-        # NOTE(sileht): oslotest does this in setUp() but we
-        # need it here
```
sileht: I haven't test but I think all the logs/print done during service.prepare_service and indexer setup will be printed.

jd:
Update for sileht comments

Issue #87, opened 2017-06-07
bulk GET metrics support


chungg:
it'd be nice to be able to support retrieving multiple metrics at once if we want to minimise API requests. there are multiple dimensions we can do this on.

1. get all metrics for resource A
2. get all 'cpu_util' for all resources
3. get any metric on any resource

i think (1) makes sense, but do we need to consider the other two cases?  if just the first case, we can have a payload like {'cpu_util': [(time, value), ...], 'memory':  [(time, value), ...]}. if we need to handle cross resources, the payload would probably need another level for resource id (or we use metric_id, but i don't think that's a good idea).

BigChief45:
I thought number 1 was already supported, but I cannot find it anywhere in the docs. I think this one would be a good starting point.

BigChief45:
I think this can be closed now.

PR #104, opened 2017-06-12
batch GET metric interface

chungg:
support ability to get multiple metrics based on id. returns result
with format matching:

{'id1': [(timestamp, granularity, value)],
 'id2': [(timestamp, granularity, value)]}

Fixes: #87

### DIFF
--- a/gnocchi/rest/__init__.py
+++ b/gnocchi/rest/__init__.py
@@ -1497,6 +1497,65 @@ class MetricsMeasuresBatchController(rest.RestController):
 
         pecan.response.status = 202
 
+    @staticmethod
+    @pecan.expose('json')
+    def get_all(**kwargs):
+        # Check RBAC policy
+        metric_ids = arg_to_list(kwargs.get('metric', []))
+        metrics = pecan.request.indexer.list_metrics(ids=metric_ids)
+        missing_metric_ids = (set(metric_ids)
+                              - set(six.text_type(m.id) for m in metrics))
+        if missing_metric_ids:
+            abort(400, {"cause": "Unknown metrics",
+                        "detail": list(missing_metric_ids)})
+
+        for metric in metrics:
+            enforce("get metric", metric)
+
+        start = kwargs.get('start')
+        if start is not None:
+            try:
+                start = utils.to_datetime(start)
+            except Exception:
+                abort(400, "Invalid value for start")
+
+        stop = kwargs.get('stop')
+        if stop is not None:
+            try:
+                stop = utils.to_datetime(stop)
+            except Exception:
+                abort(400, "Invalid value for stop")
+
+        aggregation = kwargs.get('aggregation', 'mean')
+        if (aggregation
+           not in archive_policy.ArchivePolicy.VALID_AGGREGATION_METHODS):
+            abort(
+                400,
+                'Invalid aggregation value %s, must be one of %s'
+                % (aggregation,
+                   archive_policy.ArchivePolicy.VALID_AGGREGATION_METHODS))
+
+        granularity = kwargs.get('granularity')
+        if granularity is not None:
+            try:
+                granularity = Timespan(granularity)
+            except ValueError as e:
+                abort(400, e)
+
+        metric_batch = {}
+        try:
+            for metric in metrics:
+                measures = pecan.request.storage.get_measures(
+                    metric, start, stop, aggregation, granularity)
+                metric_batch[str(metric.id)] = [
+                    (timestamp.isoformat(), offset, v)
+                    for timestamp, offset, v in measures]
+        except (storage.GranularityDoesNotExist,
+                storage.AggregationDoesNotExist) as e:
+            abort(404, e)
+
+        return metric_batch
+
 
 class SearchController(object):
     resource = SearchResourceController()
--- a/gnocchi/tests/functional/gabbits/batch-measures.yaml
+++ b/gnocchi/tests/functional/gabbits/batch-measures.yaml
@@ -181,6 +181,100 @@ tests:
         $:
           - ["2015-03-06T14:34:12+00:00", 1.0, 12.0]
 
+    - name: get multiple metrics
+      GET: /v1/batch/metrics/measures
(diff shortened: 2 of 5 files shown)


chungg:
strange, when i run tox -epy27-postgresql-redis i get the following:

Exception TypeError: "'NoneType' object is not callable" in <function _removeHandlerRef at 0x7fbf4f1ab050> ignored
Exception TypeError: "'NoneType' object is not callable" in <function _removeHandlerRef at 0x7fbf4f1ab050> ignored

 but it's not a failure

jd:
@chungg at which stage do you see this?

chungg:
hmmm.. this might be something relating to packages i have. if i run the same gnocchi code on a different driver (py27-postgresql-file), it doesn't seem to show up.

gnocchi/rest/__init__.py
```diff
@@ -1497,6 +1497,65 @@ def post(self):
...
 
+    @pecan.expose('json')
+    def get_all(self, **kwargs):
+        # Check RBAC policy
+        metric_ids = arg_to_list(kwargs.get('metric', []))
+        metrics = pecan.request.indexer.list_metrics(ids=metric_ids)
+        missing_metric_ids = (set(metric_ids)
+                              - set(six.text_type(m.id) for m in metrics))
+        if missing_metric_ids:
+            # Return one of the missing one in the error
```
jd: What about returning all of them? :)

gnocchi/rest/__init__.py
```diff
@@ -1497,6 +1497,65 @@ def post(self):
...
+                abort(400, "Invalid value for start")
+
+        stop = kwargs.get('stop')
+        if stop is not None:
+            try:
+                stop = utils.to_datetime(stop)
+            except Exception:
+                abort(400, "Invalid value for stop")
+
+        aggregation = kwargs.get('aggregation', 'mean')
```
jd: you could just write that in the signature of the function with `aggregation='mean'` by default. There's too much kwargs going here on. :)

gnocchi/rest/__init__.py
```diff
@@ -1497,6 +1497,65 @@ def post(self):
...
+                measures = pecan.request.storage.get_measures(
+                    metric, start, stop, aggregation, granularity)
+                metric_batch[str(metric.id)] = [
+                    (timestamp.isoformat(), offset, v)
+                    for timestamp, offset, v in measures]
+        except (storage.GranularityDoesNotExist,
+                storage.AggregationDoesNotExist) as e:
+            abort(404, e)
+
+        return metric_batch
```
jd: I'm worried that this could create a huge list that uses a lot of memory.

IIRC there's a way to return an iterable here rather than a finished list. That'd allow the Web server to iterate on it as the client is reading.

gnocchi/tests/functional/gabbits/batch-measures.yaml
```diff
@@ -181,6 +181,89 @@ tests:
...
+      status: 404
+
+    - name: get multiple metrics with irrelevant aggregation
+      GET: /v1/batch/metrics/measures
+      query_parameters:
+        metric:
+          - $HISTORY['list metrics'].$RESPONSE['$[0].id']
+          - $HISTORY['list metrics'].$RESPONSE['$[1].id']
+        aggregation: 'last'
+      status: 404
```
jd: would be cool to check the error message too

gnocchi/tests/functional/gabbits/batch-measures.yaml
```diff
@@ -181,6 +181,89 @@ tests:
...
+      response_json_paths:
+        {}
+
+    - name: get multiple metrics with unknown metric
+      GET: /v1/batch/metrics/measures
+      query_parameters:
+        metric: 
+          - $HISTORY['list metrics'].$RESPONSE['$[0].id']
+          - badbadba-d63b-4cdd-be89-111111111111
+      status: 404
```
jd: check the error message too

gnocchi/rest/__init__.py
chungg: let me try. but it seemed like get_all doesn't work if you use named args.

chungg: yeah, i tried. this doesn't work if you use named args. it only works if you use kwargs... i don't know why.

gnocchi/rest/__init__.py
chungg: sure.

gnocchi/rest/__init__.py
chungg: how do you stream a http response?

gnocchi/tests/functional/gabbits/batch-measures.yaml
chungg: ok

gnocchi/tests/functional/gabbits/batch-measures.yaml
chungg: ok

gnocchi/rest/__init__.py
jd: So I knew this was possible, but it's actually with Flask (sigh). I'm not sure it works with Pecan, and especially with JSON output. So forget it.

gnocchi/rest/__init__.py
```diff
@@ -1497,6 +1497,64 @@ def post(self):
...
+    @staticmethod
+    @pecan.expose('json')
+    def get_all(**kwargs):
+        # Check RBAC policy
+        metric_ids = arg_to_list(kwargs.get('metric', []))
+        metrics = pecan.request.indexer.list_metrics(ids=metric_ids)
+        missing_metric_ids = (set(metric_ids)
+                              - set(six.text_type(m.id) for m in metrics))
+        if missing_metric_ids:
+            abort(404, "Unknown metrics: %s" % ", ".join(missing_metric_ids))
```
jd: This is impossible to parse for clients. Can you use something like

```python
abort(404, {"cause": "Unknown metrics", "detail": missing_metric_ids})
```

PR #105, opened 2017-06-13
archive_policy: lighten the default archive policies

jd:
This reduces by one or two thirds the default archive policies CPU
consumption for 'low' and 'medium' by removing one or two definitions.
Also make it clearer that using low and medium is going to be fasten in
terms of CPU as the consumption depends on the number of definition in
an archive policy.

Change-Id: Iaba3b2ef88858ad777147d2859180d9a27658f1c
(cherry picked from commit 246b7873b413418bc244daba6a424f27f5f26e01)

### DIFF
--- a/doc/source/architecture.rst
+++ b/doc/source/architecture.rst
@@ -84,14 +84,14 @@ the 8 default aggregation methods (mean, min, max, sum, std, median, count,
 used will go up to a maximum of 8 × 4.1 MiB = 32.8 MiB.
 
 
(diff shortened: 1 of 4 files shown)


jd:
Default experience on Gnocchi 3.0 is not that great so I think it'd be great to backport this patch.

sileht:
I have fixed the pastamaker bot bug: https://github.com/sileht/pastamaker/commit/485058f123e93fd1a388864f275d7defebe1a23d

PR #106, opened 2017-06-13
tests: don't use wait without pid

sileht:
Without pid, wait always return 0.

This change tracks pids to ensure we have the return code
of the background jobs.


sileht:
Pull-request updated, HEAD is now 73ec19e1bf32a583733f3f04a5a5fc39227194fc

PR #110, opened 2017-06-14
ceph: change default timeout

sileht:
The default ceph timeout is unlimited. For metricd this
is not really an issue. But for API, that makes connection
stuck until ceph comeback.

This change uses a default to ensure we return an error 500 by default,
and not hang.


gnocchi/storage/ceph.py
```diff
@@ -29,7 +29,9 @@
                help='Ceph username (ie: admin without "client." prefix).'),
     cfg.StrOpt('ceph_secret', help='Ceph key', secret=True),
     cfg.StrOpt('ceph_keyring', help='Ceph keyring path.'),
-    cfg.IntOpt('ceph_timeout', help='Ceph connection timeout'),
+    cfg.IntOpt('ceph_timeout',
+               default=30,
```
chungg: is this seconds? maybe add that too help?

PR #111, opened 2017-06-14
Pull request for sileht/ceph-log-hide

sileht:
test: don't print the whole config file
tests: hide useless rados output

### DIFF
--- a/gnocchi/tests/base.py
+++ b/gnocchi/tests/base.py
@@ -336,8 +336,10 @@ class TestCase(BaseTestCase):
                                    'storage')
         elif self.conf.storage.driver == 'ceph':
             pool_name = uuid.uuid4().hex
-            subprocess.call("rados -c %s mkpool %s" % (
-                os.getenv("CEPH_CONF"), pool_name), shell=True)
+            with open(os.devnull, 'w') as f:
+                subprocess.call("rados -c %s mkpool %s" % (
+                    os.getenv("CEPH_CONF"), pool_name), shell=True,
+                    stdout=f, stderr=subprocess.STDOUT)
             self.conf.set_override('ceph_pool', pool_name, 'storage')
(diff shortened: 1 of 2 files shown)


gnocchi/tests/base.py
```diff
@@ -337,7 +337,8 @@ def setUp(self):
         elif self.conf.storage.driver == 'ceph':
             pool_name = uuid.uuid4().hex
             subprocess.call("rados -c %s mkpool %s" % (
-                os.getenv("CEPH_CONF"), pool_name), shell=True)
+                os.getenv("CEPH_CONF"), pool_name), shell=True,
+                stdout=open(os.devnull, 'w'), stderr=subprocess.STDOUT)
```
jd: this looks like it'll never be closed, so I'd go `subprocess.PIPE`, no?

sileht:
Pull-request updated, HEAD is now 52229a8e9fcb66d75c479acb09734ce6284c519e

gnocchi/tests/base.py
sileht: subprocess.PIPE can hang if don't read the pipe

gnocchi/tests/base.py
```diff
@@ -336,8 +336,10 @@ def setUp(self):
                                    'storage')
         elif self.conf.storage.driver == 'ceph':
             pool_name = uuid.uuid4().hex
-            subprocess.call("rados -c %s mkpool %s" % (
-                os.getenv("CEPH_CONF"), pool_name), shell=True)
+            with open(os.devnull) as f:
```
chungg: this defaults to 'r'... it should be 'w'?

sileht:
Pull-request updated, HEAD is now edd394460992c04cfff243dd9e04c53996250569

PR #112, opened 2017-06-14
rest: make sure we use an absolute path for api-paste file

jd:
Add Docker compose support
rest: make sure we use an absolute path for api-paste file


sileht:
I wonder if this is a good place to put this.
That adds a job that most change will never break.

jd:
Pull-request updated, HEAD is now 2f0ecfe1056ee908d73c196065f6012d47533541

jd:
Pull-request updated, HEAD is now a538ed39ca490c08f1c0742fa8c822d1203d5de6

jd:
@sileht Makes sense. I might build a different repo then.

jd:
Pull-request updated, HEAD is now daf3c67f5547353f65db44b561b8aff005911a90

PR #115, opened 2017-06-15
Metrics: Allow to copy a metric with a different archive policy

### LINKED ISSUE #4: Allow to copy a metric with a different archive policy
See https://bugs.launchpad.net/gnocchi/+bug/1501352

BigChief45:
Fixes #4

**_Still a work in progress._** 

The current approach I am using to copy a metric is to do a POST request like the following:

```http
POST v1/metric?source=67d95308-8034-4489-b9de-24b717c69c71

{
  "archive_policy_name": "high"
}
```

Where `source` stands for the UUID of the metric we want to copy. The request body will contain the archive policy that the metric copy will have.

Pending stuff:
- [ ] Add/update unit tests
- [ ] Add/update Gabbi tests
- [ ] Validate source and new archive policy definitions.

### DIFF
--- a/gnocchi/indexer/sqlalchemy.py
+++ b/gnocchi/indexer/sqlalchemy.py
@@ -664,6 +664,35 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
             raise
         return m
 
+    def copy_metric(self, id, creator, archive_policy_name, source_metric):
+        m = Metric(id=id,
+                    creator=creator,
+                    archive_policy_name=archive_policy_name,
+                    name=source_metric.name,
+                    unit=source_metric.unit,
+                    resource_id=source_metric.resource_id)
+
+        source_ap = source_metric.archive_policy
+        new_ap = self.get_archive_policy(archive_policy_name)
+
+        # TODO: Validate archive policy definitions
+        # "We are not going to be able to copy to an archive policy that is
+        # smaller grained in its definition (e.g. 1h -> 5min)"
+
+        try:
+            with self.facade.writer() as session:
+                session.add(m)
+        except exception.DBDuplicateEntry:
+            raise indexer.NamedMetricAlreadyExists(source_metric.name)
+        except exception.DBReferenceError as e:
+            if e.constraint == 'fk_metric_ap_name_ap_name':
+                raise indexer.NoSuchArchivePolicy(archive_policy_name)
+            if e.constraint == 'fk_metric_resource_id_resource_id':
+                raise indexer.NoSuchResource(source_metric.resource_id)
+            raise
+
+        return m
+
     @retry_on_deadlock
     def list_metrics(self, names=None, ids=None, details=False,
                      status='active', limit=None, marker=None, sorts=None,
--- a/gnocchi/rest/__init__.py
+++ b/gnocchi/rest/__init__.py
@@ -535,21 +535,41 @@ class MetricsController(rest.RestController):
         return definition
 
     @pecan.expose('json')
-    def post(self):
+    def post(self, source=None):
         creator = pecan.request.auth_helper.get_current_user(
             pecan.request)
         body = deserialize_and_validate(self.MetricSchema)
-        try:
-            m = pecan.request.indexer.create_metric(
-                uuid.uuid4(),
-                creator,
-                name=body.get('name'),
-                unit=body.get('unit'),
-                archive_policy_name=body['archive_policy_name'])
-        except indexer.NoSuchArchivePolicy as e:
-            abort(400, e)
+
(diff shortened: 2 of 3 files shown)


gnocchi/indexer/sqlalchemy.py
```diff
@@ -664,6 +664,35 @@ def create_metric(self, id, creator, archive_policy_name,
             raise
         return m
 
+    def copy_metric(self, id, creator, archive_policy_name, source_metric):
```
jd: This really looks like `create_metric()` actually. I'm not sure you really need this function in the indexer, the job can be done by `gnocchi.rest` directly.

At least if you need it, just use `create_metric()` here.

gnocchi/rest/__init__.py
```diff
@@ -535,21 +535,41 @@ def MetricSchema(cls, definition):
...
-            m = pecan.request.indexer.create_metric(
-                uuid.uuid4(),
-                creator,
-                name=body.get('name'),
-                unit=body.get('unit'),
-                archive_policy_name=body['archive_policy_name'])
-        except indexer.NoSuchArchivePolicy as e:
-            abort(400, e)
+
+        if not source:
```
jd: details, but in general I prefer:

```python
if <true condition>:
 <stuff>
else:
 <stuff>
```

So more `if source:` :)

gnocchi/rest/__init__.py
```diff
@@ -535,21 +535,41 @@ def MetricSchema(cls, definition):
...
+                    uuid.uuid4(),
+                    creator,
+                    name=body.get('name'),
+                    unit=body.get('unit'),
+                    archive_policy_name=body['archive_policy_name'])
+            except indexer.NoSuchArchivePolicy as e:
+                abort(400, e)
+        else:
+            source_metric = pecan.request.indexer.list_metrics(
+                id=source, details=True)[0]
```
jd: watch out, doing [0] on `None` or an empty list is a bad idea

gnocchi/rest/__init__.py
```diff
@@ -535,21 +535,41 @@ def MetricSchema(cls, definition):
...
+            # Copy the metric
+            try:
+                m = pecan.request.indexer.copy_metric(
+                    uuid.uuid4(),
+                    creator,
+                    archive_policy_name=body['archive_policy_name'],
+                    source_metric=source_metric)
+            except indexer.NoSuchArchivePolicy as e:
+                abort(400, e)
+
```
jd: yeah you also need to copy the data. whereas I'm not sure you need a `copy_metric` method in the indexer, I sense you'll need one in the storage driver. :)

gnocchi/rest/__init__.py
BigChief45: Is there a better way for obtaining a single metric in this case? I do not like the idea of using `list_metrics`. I couldn't find a method for this in the Metric controller, I am thinking the `_lookup` method is handling that somehow.

gnocchi/rest/__init__.py
BigChief45: By _data_ you mean all the associated stored measures for this metric? I suppose this would have to be added to all the available storage drivers?

jd: By data yeah I do mean the associated store measures. You suppose well.

gnocchi/rest/__init__.py
jd: No, list_metrics with `id=` is the good call. It just does a `SELECT` with the right where, so it's equivalent to what would be a `get_metric`.

jd:
@BigChief45 Check my latest comment on the original bug report too. I closed it because I don't think it's achievable in the end. :(

BigChief45:
@jd Is [this](https://bugs.launchpad.net/gnocchi/+bug/1501352) the original bug report? 

I am a bit confused. You recently wrote:

> I think the guy that opened the bug 2 years ago did not think that through. So no, I don't think that's possible. In the original bug report, the suggestion was made to rather change the archive policy. I think it's smarter.

But that Launchpad bug was opened by you. Unless that Launchpad bug is not the original bug report...

jd:
@BigChief45 I see, you're new around here. That was sarcasm. ;-)

BigChief45:
@jd Haha yes.

So I guess that if this is not achievable, this pull request can be closed now.

jd:
Ack, closing. Thanks @BigChief45 for trying!

chungg:
i think this might be possible, but only if you're mapping it to another archive policy with exact same granularities and aggregation_methods and only change the timespan... which is very limited in functionality so i agree with abandoning

PR #120, opened 2017-06-15
doc: mention gnocchi-openshift as an install solution

jd:


jd:
@chungg it's not in this repo so I'd not add a release note.

We don't have any polling daemon so I can't see how we'd add OpenShift support that way. But I'm sure writing something for e.g. collectd could do it. :)

jd:
Pull-request updated, HEAD is now b260b7553ed0bb996bb77c3ec7bef89e2e9e3378

jd:
Pull-request updated, HEAD is now 7a74c11b8a5a01b0400ea7805a407affa610cc66

PR #127, opened 2017-06-16
doc: improve various doc and add comparison table

jd:
- Rename running to operating
- Mention aggregation methods in archive policies
- Add a comparison table


jd:
Pull-request updated, HEAD is now c4fbccc6bb7e00ae506106bd02dd5d000b91f7ac

doc/source/operating.rst
```diff
@@ -48,7 +48,10 @@ Once written to `/etc/gnocchi/uwsgi.ini`, it can be launched this way::
 How to define archive policies
 ==============================
 
-In Gnocchi, the archive policy definitions are expressed in number of points.
+In Gnocchi, the archive policies define how the metrics are aggregated and how
+long they are stored. Each archive policy definition is expressed in a number
```
chungg: s/in a number/as the number/

s/points other/points over/

tox.ini
```diff
@@ -139,3 +139,6 @@ deps = {[testenv:docs]deps}
        oslosphinx
 commands =
     pifpaf -g GNOCCHI_INDEXER_URL run postgresql -- sphinx-versioning build doc/source doc/build/html
+
+[doc8]
+ignore-path = doc/source/rest.rst,doc/source/comparison-table.rst
```
chungg: what's this for?

jd: This is to ignore doc8 on rest.rst (autogenerated from the j2 template, as we do currently) and comparison-table, which is a very large table that doc8 does not like because lines are too long.

jd:
Updated with gordc hints

PR #128, opened 2017-06-19
doc: do not build 2.0.x tags

jd:
We don't build the stable/2.0 branch so don't build the tags neither.


jd:
Pull-request updated, HEAD is now 51c3631f4c01859b7d660de35594de63525f4913

jd:
Yeah let's close for #129 now that it is ready.

PR #129, opened 2017-06-19
doc: disable all tags building in versioning

jd:
There is little sense in building docs for all tags. The stable branches ought
to be enough anyway. Let's skip them entirely.


jd:
Closing so it does not get merged. @sileht the banner is wrong with this patch and I'm not sure how to fix it. It says on stable/* that this is the dev version and that "master" should be checked. Any hint?

jd:
Reopening and dismissing reviews, simpler :)

doc/source/conf.py
```diff
@@ -183,7 +183,7 @@
 scv_banner_greatest_tag = True
```
sileht: I think we should replace:

scv_banner_greatest_tag = True

by 

scv_banner_main_ref = "master"

jd: That does not solve it because it does *not* show the banner in master. That's where it should be. :/

sileht: So we need:

-scv_banner_greatest_tag = True
+scv_banner_main_ref = 'stable/4.0'

jd: Ok, we need to find a way so that's dynamically computed later then… :)

jd:
Pull-request updated, HEAD is now 5f3c70ad1b56949f5af248fab8c9f71d442afe6e

jd:
Pull-request updated, HEAD is now 49196a37b0062180ec71c4b7444eee3277f59bcc

PR #133, opened 2017-06-20
utils: use ASCII bytes as member id

### LINKED ISSUE #130: Coordination member_id should be string according to tooz source
https://github.com/gnocchixyz/gnocchi/commit/a8eb25e04cdcec26161ae31e65a26b8a3f326daf introduced a change to align with tooz docs but according to the source code tooz expects the member_id to be a string (https://github.com/openstack/tooz/blame/master/tooz/coordination.py#L754-L755).

```
    :param member_id: the id of the member
    :type member_id: str
```

When using bytes as per tooz documentation the library fails to initialise due to a utf8 conversion.

Some drivers of tooz are still not ready, memcache text protocol has an issue and so does the consul driver which still expects a string which it decodes to form a session_name.

I propose that the above change be rolled back until tooz fully updates its drivers.

jd:
Tooz actually wants ASCII bytes and not random bytes.

Fixes #130


jd:
Pull-request updated, HEAD is now a7ce50fc6c8de41085a9eab43464a71954e20d5e

jd:
Pull-request updated, HEAD is now 74dd0a033e241c875b63bcfd71c3f834af830eb2

PR #136, opened 2017-06-22
rest: move _carbonara.SackLockTimeoutError to storage

jd:
gnocchi.rest actually expects storage.SackLockTimeoutError so it currently is
not caught.


jd:
Pull-request updated, HEAD is now 64e1b81d5590ce2b9ff3f99d3e6725c532a2b8e9

PR #138, opened 2017-06-23
tox: don't specify envdir

jd:
Putting `envdir = upgrade` makes tox create an upgrade in the middle of the
source directory, which is not convenient/standard and gets scanned by e.g.
flake8. Let's each job use its own default directory.


jd:
Pull-request updated, HEAD is now 341545cf017c875c3394c9a7980c418f7d5db4c3

jd:
Pull-request updated, HEAD is now 0ff791ed0aca6f03503ab24e12ae9246c71253c6

PR #141, opened 2017-06-26
Install gnocchi flavor of pifpaf

jd:
Otherwise we miss uwsgi.


jd:
Pull-request updated, HEAD is now 3adfb0e51f51b100e5658f7ed07ed12eb83cc425

jd:
Not needed, will install uwsgi in the docker image so it has the right plugins and it's faster.

PR #144, opened 2017-06-26
Remove deprecated noauth auth mode

jd:
Sem-Ver: api-break


jd:
Pull-request updated, HEAD is now f2ec9aa81acbe81649c5b533a3b4cbe8e5175d2c

PR #145, opened 2017-06-26
metricd: use a better number of partitions for the hashring

jd:
The current hashring distributing the sacks across the metricd workers is 200.
That means the the smallest number of sacks that a worker can have is
(number_of_sacks/200), being 0.64 on a default installation.

That's weird ratio.

If the number of sacks created is 4096 and the number of workers is 4096, then
each worker manages at least 4096/200=20,48 sacks. There's no good reason for
that, each worker should be responsible for one sack (+ replicas) in such a
setup.

Let's use the number of sacks as the partition number so the minimum amount of
sack that a worker will always be able to manage is 1 (+ replicas).

### DIFF
--- a/gnocchi/cli.py
+++ b/gnocchi/cli.py
@@ -194,7 +194,7 @@ class MetricProcessor(MetricProcessBase):
             six.moves.range(self.store.incoming.NUM_SACKS))
         try:
             self.partitioner = self.coord.join_partitioned_group(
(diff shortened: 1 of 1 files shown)


jd:
Pull-request updated, HEAD is now dc7e9839c674404d68d15545387f38225f04fcd8

jd:
This patch does not seem good after all and we never hit an issue yet, so we'll see later :)

PR #150, opened 2017-06-27
improve moving average data retrieval

chungg:
don't unnecessarily pull in all granularities, search through
all series for min granularity, and rebuild.

just detect if valid, pull in granularity that's valid, and continue.


chungg:
*shrugs* i was going to replace it completely with pandas rolling function which does exactly this... i just haven't tested it out.

i guess bigger question is what do we want in gnocchi? right now we focus on storing timeseries but we can also do a resampling and moving windows... and if we keep going down this timeseries path we can do ARIMA, Kalman filtering, etc... (i'm "trying" to learn this stuff currently). depends how far the analytic rabbit hole we want to go.

jd:
I'm cool with that. If you got interesting use cases to add in doc, that'd be cool!

chungg:
i'm going to deprecate this. i just took a closer look and this isn't even correct. the rolling average in the test for some reason looks for a left-aligned rolling average which is opposite of standard rolling average which lags (right-aligned)... also, the unused parameters don't seem right either.

PR #152, opened 2017-06-27
DN: add some debug to see why uwsgi might be long to start

jd:


jd:
Pull-request updated, HEAD is now 7c59f82fb3ee95dd09746220c4f85b72c6037f34

PR #153, opened 2017-06-27
tests: exit when an error happens with storage setup

jd:
In run-func-tests, the storage set-up is made using "eval $(pifpaf)". So pifpaf
is run in a subprocess and its output is evaluated. But the return value of
pifpaf is never checked, the -x setting is not taken into account here.

Example:

  $ eval $(ls /foobar)
  ls: /foobar: No such file or directory
  $ echo $?
  0

Let's add an explicit check later to see that everything has been started
correctly.


jd:
Pull-request updated, HEAD is now 31e579967b96b584e6039b7ddb145f463924354d

jd:
Pull-request updated, HEAD is now 31e579967b96b584e6039b7ddb145f463924354d

PR #155, opened 2017-06-28
statsd: remove deprecated user/project ids opts

jd:


gnocchi/opts.py
```diff
@@ -176,16 +176,7 @@ def list_opts():
...
-                'user_id',
-                deprecated_for_removal=True,
-                help='User ID to use to identify statsd in Gnocchi'),
-            cfg.StrOpt(
-                'project_id',
-                deprecated_for_removal=True,
-                help='Project ID to use to identify statsd in Gnocchi'),
-            cfg.StrOpt(
                 'creator',
-                default="${statsd.user_id}:${statsd.project_id}",
```
chungg: where are these values coming from?

jd: the config file. it's `[statsd]user_id`. oslo.config supports variable substitution

chungg: does the default become ':' if they're not there then?

jd: Yes.
(though you realize your commenting the *current* code, not my change right? :))

chungg: oh. word. i saw green for some reason.lol

PR #158, opened 2017-06-29
Allow metric aggregation with a POST

sileht:
When we aggregate metric with a GET we have a limitation with the number
of metric we can aggregate.


gnocchi/tests/functional/gabbits/aggregation.yaml
```diff
@@ -70,13 +70,29 @@ tests:
...
       GET: /v1/aggregation/metric?metric=$HISTORY['get metric list'].$RESPONSE['$[0].id']&metric=$HISTORY['get metric list'].$RESPONSE['$[1].id']&granularity=1&refresh=true
       response_json_paths:
         $:
           - ['2015-03-06T14:33:57+00:00', 1.0, 23.1]
           - ['2015-03-06T14:34:12+00:00', 1.0, 7.0]
 
+    - name: POST measure aggregates by granularity with refresh
+      POST: /v1/aggregation/metric?granularity=1&refresh=true
+      request_headers:
+          content-type: application/json
```
chungg: this is default already.

chungg:
i'm curious, what's the limitation?

sileht:
@chungg each webserver, wsgi stack, webbrowser have it own limitation, for example the apache default limit is 8000 bytes. So maybe something like 150

PR #172, opened 2017-06-30
tests: We must ensure variable are reinit

sileht:


run-func-tests.sh
```diff
@@ -19,6 +19,8 @@ GNOCCHI_TEST_STORAGE_DRIVERS=${GNOCCHI_TEST_STORAGE_DRIVERS:-file}
 GNOCCHI_TEST_INDEXER_DRIVERS=${GNOCCHI_TEST_INDEXER_DRIVERS:-postgresql}
 for storage in ${GNOCCHI_TEST_STORAGE_DRIVERS}; do
     for indexer in ${GNOCCHI_TEST_INDEXER_DRIVERS}; do
+        STORAGE_URL=
```
jd: unset?

sileht:
Pull-request updated, HEAD is now 8760e33c76573c210db95fa02b16ba5d363b2810

jd:
Seems like Swift does not start but I don't see how it's related?

sileht:
Pull-request updated, HEAD is now c631ae63d78050d2685debf7be0e4e744a1fcea0

sileht:
Yes swift doesn't work since a while, we was testing file driver instead. With this change we fail as expected...

If that related to xattr, I'm guessing ceph driver is not tested correctly too :(

sileht:
Ceph tests looks ok, so it's not a xattr issue.

sileht:
Pull-request updated, HEAD is now 8760e33c76573c210db95fa02b16ba5d363b2810

sileht:
Pull-request updated, HEAD is now f861a104f74e91da3669c9605a83d67aa39b1268

sileht:
Pull-request updated, HEAD is now 31a155d0947d99cb0b9f67f5a51e303f564b62b0

PR #173, opened 2017-06-30
DNM: debug uwsgi

sileht:


sileht:
Pull-request updated, HEAD is now f9e18bff1ad533976c531781181eab07abf4db25

sileht:
Pull-request updated, HEAD is now dc68366341594c3ca6dca9ca6918fe14af8b0536

jd:
@sileht I've a log that shows that uwsgi receive the kill signal but one of the worker is not killed yet when the new uwsgi starts:

https://api.travis-ci.org/jobs/248805407/log.txt?deansi=true

I think we need to change `--worker-reload-mercy` to 0 or something like that. I did not try yet.

sileht:
My last pifpaf PR should have fix the issue, it sigkill all uwsgi.

sileht:
Also each time swift doesn't work correctly so uswgi i maybe stuck because it can't join swift

PR #174, opened 2017-06-30
Allow to derive timeseries

sileht:
This is useful to convert cumulative metric into
non cumulative one.

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -490,6 +490,11 @@ class AggregatedTimeSerie(TimeSerie):
         return AggregatedTimeSerie.from_grouped_serie(
             self.group_serie(sampling), sampling, self.aggregation_method)
 
+    def derive(self):
+        return AggregatedTimeSerie.from_data(
+            self.sampling, self.aggregation_method,
+            self.ts.index[1:], numpy.diff(self.ts.values))
+
     @classmethod
     def from_data(cls, sampling, aggregation_method, timestamps=None,
                   values=None, max_size=None):
--- a/gnocchi/rest/__init__.py
+++ b/gnocchi/rest/__init__.py
@@ -406,7 +406,8 @@ class MetricController(rest.RestController):
 
     @pecan.expose('json')
     def get_measures(self, start=None, stop=None, aggregation='mean',
-                     granularity=None, resample=None, refresh=False, **param):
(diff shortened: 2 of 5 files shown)


sileht:
I just would like to confirm about the chosen solution. It looks simple as first side, but...

For example, collectd sends to Gnocchi the number of packets send since the server boot. By default Gnocchi record the mean/min/max/sum/... but not first and last. max is close to last most of the times. But when the server reboot, we got one point wrong during one aggregation period.

So deriving a metric make sense only with first and last method.  My initial plan is :

* Write some archive policy rules in collectd-gnocchi to add last/first and remove min/max/mean/sum/.... for cumulative metrics
* Document in Gnocchi that derive make sense only for last and first.

Is it good ?

**/me switches devil advocate**

With the current implementation, you can't have mean/max/min of packet/s. That looks bad for a timeseries dabatase, no ?

**/me switches the other proposal guy**

* We do that at archive policy layer with a flag like derive=True
* We store somewhere only the last datapoint (or the first one ?)
  * I imagine storing the last datapoint as-is in each timeserie during the serialization 
* We store we compute the next derivated value and aggregate it with all methods of the archive policy

This solution is heavier to implement but should fix the devil advocate point.

jd:
If you have a timeseries where values is a counter that resets, the best thing is to store only `last`. As you say, `mean` or `min` and `max` makes no sense in a time range where the counter resets to 0.

If you have `last`, it's easy to compute the derive. You can also add another option with `derive` that would be something like `derive=true&absolute=true` where you'd return the `abs` difference between the points. That'd solve the problem of a counter going back to 0.

sileht:
I'm more worred about the fact we can't do mean/max/min of packet/s.

jd:
So then we'd need a solution you like described, with a different archive policy rules that states "compute the diff between points and then aggregate".
Why not.

sileht:
The other approach: https://github.com/gnocchixyz/gnocchi/pull/180

chungg:
i think you're right, it doesn't allow us to compute mean of rate.

getting the diff  is useful in case where you want to eliminate the trend of a time-series... but you generally diff the log so maybe this doesn't help exactly.

i am indifferent to this solution. are we declaring /aggregates endpoint to be the place where we allow users to play with the data that is stored? i do think we should have an endpoint(s) for requests that are beyond retrieval of stored data.

PR #179, opened 2017-07-04
swift: we must use the number of default sack

sileht:
We currently set the number of sack already configured instead of the
asked number. Making upgrade fail when the number of sack is not yet
set.

This change fixes that.

### DIFF
--- a/run-func-tests.sh
+++ b/run-func-tests.sh
@@ -15,9 +15,15 @@ check_empty_var() {
     fi
 }
 
+PYTHON_VERSION_MAJOR=$(python -c 'import sys; print(sys.version_info.major)')
+
 GNOCCHI_TEST_STORAGE_DRIVERS=${GNOCCHI_TEST_STORAGE_DRIVERS:-file}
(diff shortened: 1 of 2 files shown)


sileht:
The swift driver is currently not tested because of a bug in run-func-tests.sh, ...

It's almost fixed. I hope I will not discover other bugs.

sileht:
Pull-request updated, HEAD is now f5bf82d75142d39a83545a9bd1e3b1c6d1614c2c

sileht:
Pull-request updated, HEAD is now 2f6abb9534cd34e0a0659c46f55cfc7f9efd96fe

run-func-tests.sh
```diff
@@ -15,9 +15,15 @@ check_empty_var() {
     fi
 }
 
+PYTHON_VERSION_MAJOR=$(python --version 2>&1 | awk '{print $2}' | awk -F. '{print $1}')
```
jd: python -c 'import sys; print(sys.version_info.major)'

sileht: done

sileht:
Pull-request updated, HEAD is now cbc98d0ae6b8e8fe4061aab6255071f4e90c76e1

PR #180, opened 2017-07-04
Add rate of change aggregation methods

sileht:
This change allows to compute the rate of change before applying
the aggregation method. To do so, aggregation method need to be
prefixed by "rate:"

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -79,13 +79,23 @@ def round_timestamp(ts, freq):
 
 
 class GroupedTimeSeries(object):
-    def __init__(self, ts, granularity):
+    def __init__(self, ts, granularity, start=None):
         # NOTE(sileht): The whole class assumes ts is ordered and don't have
         # duplicate timestamps, it uses numpy.unique that sorted list, but
         # we always assume the orderd to be the same as the input.
+        self.granularity = granularity
         freq = granularity * 10e8
-        self._ts = ts
-        self.indexes = (numpy.array(ts.index, numpy.float) // freq) * freq
+        self.start = start
+        if start is None:
+            self._ts = ts
+            self._ts_for_derive = ts
+        else:
+            self._ts = ts[start:]
+            start_derive = start - pandas.Timedelta(freq, unit='ns')
+            self._ts_for_derive = ts[start_derive:]
+
+        self.indexes = (numpy.array(self._ts.index,
+                                    numpy.float) // freq) * freq
         self.tstamps, self.counts = numpy.unique(self.indexes,
                                                  return_counts=True)
 
@@ -125,6 +135,7 @@ class GroupedTimeSeries(object):
         counts, timestamps = self._count()
         cumcounts = numpy.cumsum(counts) - 1
         values = self._ts.values[cumcounts]
+
         return pandas.Series(values, pandas.to_datetime(timestamps))
 
     def first(self):
@@ -156,6 +167,14 @@ class GroupedTimeSeries(object):
                         *args, **kwargs)
         timestamps = tstamps.astype('datetime64[ns]', copy=False)
         return pandas.Series(values, pandas.to_datetime(timestamps))
+
+    def derived(self):
+        timestamps = self._ts_for_derive.index[1:]
+        values = numpy.diff(self._ts_for_derive.values)
+        # FIXME(sileht): create some alternative __init__ to avoid creating
+        # useless Pandas object, recounting, timestamps convertion, ...
+        return GroupedTimeSeries(pandas.Series(values, timestamps),
+                                 self.granularity, self.start)
 
 
 class TimeSerie(object):
@@ -220,14 +239,14 @@ class TimeSerie(object):
         except IndexError:
             return
 
-    def group_serie(self, granularity, start=0):
+    def group_serie(self, granularity, start=None):
         # NOTE(jd) Our whole serialization system is based on Epoch, and we
         # store unsigned integer, so we can't store anything before Epoch.
         # Sorry!
         if not self.ts.empty and self.ts.index[0].value < 0:
             raise BeforeEpochError(self.ts.index[0])
 
-        return GroupedTimeSeries(self.ts[start:], granularity)
+        return GroupedTimeSeries(self.ts, granularity, start)
 
     @staticmethod
     def _compress(payload):
--- a/gnocchi/storage/_carbonara.py
+++ b/gnocchi/storage/_carbonara.py
@@ -258,9 +258,16 @@ class CarbonaraBasedStorage(storage.StorageDriver):
                       metric, grouped_serie,
                       previous_oldest_mutable_timestamp,
                       oldest_mutable_timestamp):
+
+        if aggregation.startswith("rate:"):
+            grouped_serie = grouped_serie.derived()
+            aggregation_to_compute = aggregation[5:]
+        else:
+            aggregation_to_compute = aggregation
+
         ts = carbonara.AggregatedTimeSerie.from_grouped_serie(
             grouped_serie, archive_policy_def.granularity,
-            aggregation, max_size=archive_policy_def.points)
+            aggregation_to_compute, max_size=archive_policy_def.points)
 
         # Don't do anything if the timeserie is empty
         if not ts:
@@ -389,6 +396,10 @@ class CarbonaraBasedStorage(storage.StorageDriver):
         block_size = metric.archive_policy.max_block_size
         back_window = metric.archive_policy.back_window
         definition = metric.archive_policy.definition
+        # NOTE(sileht): We keep one more blocks to calculate rate of change
+        # correctly
+        if any(filter(lambda x: x.startswith("rate:"), agg_methods)):
+            back_window += 1
 
         try:
             ts = self._get_unaggregated_timeserie_and_unserialize(
--- a/gnocchi/tests/functional/gabbits/metric-derived.yaml
+++ b/gnocchi/tests/functional/gabbits/metric-derived.yaml
@@ -1,0 +1,199 @@
+fixtures:
+    - ConfigFixture
+
+defaults:
+  request_headers:
+    # User foobar
+    authorization: "basic Zm9vYmFyOg=="
+    content-type: application/json
+
+tests:
+    - name: create archive policy
+      desc: for later use
+      POST: /v1/archive_policy
+      request_headers:
+        # User admin
+        authorization: "basic YWRtaW46"
+      data:
+          name: carrot-cake
+          aggregation_methods:
+              - rate:mean
+              - rate:max
+              - rate:95pct
+              - max
+          definition:
+              - granularity: 1 minute
+      status: 201
+
+    - name: create valid metric
+      POST: /v1/metric
+      data:
+          archive_policy_name: carrot-cake
+      status: 201
+
+    - name: push measurements to metric
+      POST: /v1/metric/$RESPONSE['$.id']/measures
+      data:
+          - timestamp: "2015-03-06T14:33:00"
+            value: 10
+          - timestamp: "2015-03-06T14:34:10"
+            value: 13
+          - timestamp: "2015-03-06T14:34:20"
+            value: 13
+          - timestamp: "2015-03-06T14:34:30"
+            value: 15
+          - timestamp: "2015-03-06T14:34:40"
+            value: 18
+          - timestamp: "2015-03-06T14:34:50"
+            value: 20
+          - timestamp: "2015-03-06T14:35:00"
+            value: 22
+          - timestamp: "2015-03-06T14:35:10"
+            value: 26
+          - timestamp: "2015-03-06T14:35:20"
+            value: 30
+          - timestamp: "2015-03-06T14:35:30"
+            value: 31
+          - timestamp: "2015-03-06T14:35:40"
+            value: 37
+          - timestamp: "2015-03-06T14:35:50"
+            value: 55
+          - timestamp: "2015-03-06T14:36:00"
+            value: 62
+          - timestamp: "2015-03-06T14:36:10"
+            value: 100
+          - timestamp: "2015-03-06T14:36:20"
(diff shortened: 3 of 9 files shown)


gnocchi/carbonara.py
```diff
@@ -157,6 +159,18 @@ def _scipy_aggregate(self, method, remove_unique=False, *args, **kwargs):
         timestamps = tstamps.astype('datetime64[ns]', copy=False)
         return pandas.Series(values, pandas.to_datetime(timestamps))
 
+    def derived(self):
+        timestamps = self._ts.index[1:]
+        # NOTE(sileht): we use absolute() as best-effort for cumulative counter
+        # reset, what ever the solution we use, the first point after the
+        # counter reset will always be wrong, because we can't known when
+        # the counter have been reseted between two points.
+        values = numpy.absolute(numpy.diff(self._ts.values))
```
jd: I'd say do not use absolute here. If you use absolute you lose information. It'd be better to store the relative diff, and then, with an API option, apply `abs` on values _if_ needed.

sileht:
Pull-request updated, HEAD is now 2176cf1e034f1d02d0045fbf02415778b54e92c3

sileht:
Pull-request updated, HEAD is now af5ce4bac6489e950d96ddb47623e4cd6c579233

gnocchi/tests/functional/gabbits/metric-derived.yaml
```diff
@@ -0,0 +1,114 @@
...
+          - timestamp: "2015-03-06T14:34:10"
+            value: 13
+          - timestamp: "2015-03-06T14:34:20"
+            value: 13
+          - timestamp: "2015-03-06T14:34:30"
+            value: 15
+          - timestamp: "2015-03-06T14:34:40"
+            value: 18
+          - timestamp: "2015-03-06T14:34:50"
+            value: 20
```
chungg: just to clarify, when we do rate:mean, how does it compute 14:34:00?

(13-10 + 13-13 + 15-13 + 18-15 + 20-18) / 5?

how does it get the backwindow before the current one if it's not part of the same post?

sileht: We have point at 33

chungg: but if we didn't... say we pushed 33 in one post. and then pushed min34 metrics in another post... do we always have access to the previous window? maybe add a test to verify this?

disclaimer: i'm not sure how it works but i imagine it needs to previous window to build current window?

sileht: everything is recomputed with the raw measures, so if we are out the backwindow we don't have the point.

The archive policy for this tests is 1 point per minutes while the input rate is 1 point every 10 seconds.

I will retry to add a tests

sileht: Looks like my new tests don't behave as expected.

sileht: the raw timeseries doesn't have the last point of the previous block so the first value of the current block is often wrong. To solve this I need to store one additional point.

gnocchi/tests/test_carbonara.py
```diff
@@ -158,11 +158,63 @@ def test_before_epoch(self):
...
+        ts = self._resample(ts, 60, 'last', derived=True)
+
+        self.assertEqual(4, len(ts))
+        self.assertEqual(
+            [(datetime.datetime(2014, 1, 1, 12, 0, 0), 60, 5),
+             (datetime.datetime(2014, 1, 1, 12, 1, 0), 60, 4),
+             (datetime.datetime(2014, 1, 1, 12, 3, 0), 60, 92),
+             (datetime.datetime(2014, 1, 1, 12, 4, 0), 60, 2)],
+            list(ts.fetch(
+                from_timestamp=iso8601.parse_date(
```
jd: Please don't use iso8601 here.

(that will avoid me having to remove it in a future patch :)

gnocchi/tests/test_carbonara.py
```diff
@@ -158,11 +158,63 @@ def test_before_epoch(self):
...
+
+        self.assertEqual(5, len(ts))
+        self.assertEqual(
+            [(datetime.datetime(2014, 1, 1, 12, 0, 0), 60, 5),
+             (datetime.datetime(2014, 1, 1, 12, 1, 0), 60, 5),
+             (datetime.datetime(2014, 1, 1, 12, 2, 0), 60, 11),
+             (datetime.datetime(2014, 1, 1, 12, 3, 0), 60, -32),
+             (datetime.datetime(2014, 1, 1, 12, 4, 0), 60, 16)],
+            list(ts.fetch(
+                from_timestamp=iso8601.parse_date(
```
jd: Please don't use iso8601 here.

(that will avoid me having to remove it in a future patch :)

gnocchi/tests/test_carbonara.py
```diff
@@ -158,11 +158,63 @@ def test_before_epoch(self):
...
     @staticmethod
-    def _resample(ts, sampling, agg, max_size=None):
+    def _resample(ts, sampling, agg, max_size=None, derived=False):
         grouped = ts.group_serie(sampling)
+        if derived:
+            grouped = grouped.derived()
         return carbonara.AggregatedTimeSerie.from_grouped_serie(
             grouped, sampling, agg, max_size=max_size)
 
+    def test_derived_last(self):
```
jd: this test is called last but it uses mean?

gnocchi/tests/test_carbonara.py
```diff
@@ -158,11 +158,63 @@ def test_before_epoch(self):
...
+             (datetime.datetime(2014, 1, 1, 12, 3, 2), 105),
+             (datetime.datetime(2014, 1, 1, 12, 3, 22), 5),  # Counter reset
+             (datetime.datetime(2014, 1, 1, 12, 3, 42), 7),
+             (datetime.datetime(2014, 1, 1, 12, 4, 9), 23)])
+        ts = self._resample(ts, 60, 'mean', derived=True)
+
+        self.assertEqual(5, len(ts))
+        self.assertEqual(
+            [(datetime.datetime(2014, 1, 1, 12, 0, 0), 60, 5),
+             (datetime.datetime(2014, 1, 1, 12, 1, 0), 60, 5),
```
jd: I don't get why the result is 5. We have 3 points during this minute, 65, 66 and 70.

mean of them would be 67.
mean of ((70-66)+(66-65)) = 2.5

No?

jd: I don't understand why 5 is the result.

releasenotes/notes/rate-archive-policy-74888634f90a81e3.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    New aggregation methods are available for archive policy; rate:mean, rate:last, .... These new methods
+    allow to compute the timeseries rate of change before applying the XXX aggregation method.
```
jd: s/XXX/selected/

sileht: done

gnocchi/tests/test_carbonara.py
sileht: done

gnocchi/tests/test_carbonara.py
sileht: done

gnocchi/tests/test_carbonara.py
sileht: done

gnocchi/tests/test_carbonara.py
sileht: We first do the diff and then the mean of each block

> 2014-1-1 12:00:00 50
> 2014-1-1 12:00:04 55  5
>  -> 5
>
> 2014-1-1 12:01:02 65  10
> 2014-1-1 12:01:14 66  1
> 2014-1-1 12:01:24 70  4
>  -> 5
>
> 2014-1-1 12:02:04 83  13
> 2014-1-1 12:02:35 92  9
> 2014-1-1 12:02:42 103 11
>  -> 11
>
> 2014-1-1 12:03:02 105 2
> 2014-1-1 12:03:22 5   -110
> 2014-1-1 12:03:42 7   2
>  -> -32
>
> 2014-1-1 12:04:09 23  16
>  -> 16

sileht: Currently when the diff is built, we keep the ending timestamp. If we keep the starting timestamp instead, we result would be:
 
> (Timestamp('2014-01-01 12:00:00'), 60.0, 7.5)
> (Timestamp('2014-01-01 12:01:00'), 60.0, 6.0)
> (Timestamp('2014-01-01 12:02:00'), 60.0, 7.333333333333333)
> (Timestamp('2014-01-01 12:03:00'), 60.0, -27.333333333333332)

I wonder which one make more sense or not.

sileht:
Pull-request updated, HEAD is now 8decf76e42b773291b6946fe5069cfeff74664fd

sileht:
Pull-request updated, HEAD is now 2db31e5ec87e788ada6304d5820dcf5241c4cade

sileht:
Pull-request updated, HEAD is now cc50772ec2c811d323800df7c733e5ec6f2b36cb

sileht:
Pull-request updated, HEAD is now a629a78ae6ad26e85816db13cb0f346dfaca1eed

gnocchi/tests/test_carbonara.py
jd: OK thanks for the explanation, I forgot to compute `2014-1-1 12:01:02 65 10` in my mind :) That all makes sense. I think keeping the ending timestamp makes more sense. Let's go that way for now, we can add yet-another-agg-method if needed later.

gnocchi/storage/_carbonara.py
```diff
@@ -396,8 +403,10 @@ def _compute_and_store_timeseries(self, metric, measures):
 
         agg_methods = list(metric.archive_policy.aggregation_methods)
         block_size = metric.archive_policy.max_block_size
-        back_window = metric.archive_policy.back_window
         definition = metric.archive_policy.definition
+        # NOTE(sileht): We keep one more blocks to calculate rate of change
+        # correctly
+        back_window = metric.archive_policy.back_window + 1
```
jd: So this is done for all timeseries? Looks to me we should limit it to `rate:`, no?

chungg: :( can we do this only if rate is being computed? i don't think it makes sense to double the back_window size unnecessarily.

chungg: damn. you post before me

sileht:
Pull-request updated, HEAD is now 9a56226c0b8b8282c1fd58a2467365217eaaf4d9

gnocchi/storage/_carbonara.py
sileht: done

sileht:
Pull-request updated, HEAD is now 1a2b58f10336c0916ed1dbe5ac194e6aa4f739db

gnocchi/tests/functional/gabbits/metric-derived.yaml
```diff
@@ -0,0 +1,199 @@
...
+      status: 202
+
+    - name: get measurements rate:mean
+      GET: /v1/metric/$HISTORY['create valid metric'].$RESPONSE['id']/measures?aggregation=rate:mean&refresh=true
+      status: 200
+      response_json_paths:
+          $:
+            - ['2015-03-06T14:34:00+00:00', 60.0, 2.0]
+            - ['2015-03-06T14:35:00+00:00', 60.0, 5.833333333333333]
+            - ['2015-03-06T14:36:00+00:00', 60.0, 9.166666666666666]
```
chungg: stupid question:

- 14:35:00 is from "2015-03-06T14:34:50" to "2015-03-06T14:35:50"? which is diff of 35 over 6 points and gives us 5.8333...
- 14:34:00 is from "2015-03-06T14:33:00" to "2015-03-06T14:34:50"? which is diff of 10 over 6 points... how does that give us 2.0?

gnocchi/storage/_carbonara.py
```diff
@@ -398,6 +405,10 @@ def _compute_and_store_timeseries(self, metric, measures):
         block_size = metric.archive_policy.max_block_size
         back_window = metric.archive_policy.back_window
         definition = metric.archive_policy.definition
+        # NOTE(sileht): We keep one more blocks to calculate rate of change
+        # correctly
+        if list(filter(lambda x: x.startswith("rate:"), agg_methods)):
```
jd: Good point. I WITHDRAW MY APPROVAL!!11!

gnocchi/tests/functional/gabbits/metric-derived.yaml
jd:
```
 +          - timestamp: "2015-03-06T14:33:00"
 +            value: 10
 +          - timestamp: "2015-03-06T14:34:10"
 +            value: 13
 +          - timestamp: "2015-03-06T14:34:20"
 +            value: 13
 +          - timestamp: "2015-03-06T14:34:30"
 +            value: 15
 +          - timestamp: "2015-03-06T14:34:40"
 +            value: 18
 +          - timestamp: "2015-03-06T14:34:50"
 +            value: 20
```
((13-10) + (13-13) + (15-13) + (18-13) + (20-18)) / 6
= (3 + 0 + 2 + 5 + 2) / 6
= 12 / 6
= 2

chungg: oh  nevermind. there's no 14:34:00 point so it's only 5 points.

gnocchi/storage/_carbonara.py
sileht: done

sileht:
Pull-request updated, HEAD is now 727a4bd6837862557e39e8256a94e8e1ddc325d6

PR #188, opened 2017-07-06
Add a notifier subsystem

jd:
This will allow metricd to be notifier as soon as a sack is ready to be
processed.


jd:
Pull-request updated, HEAD is now 5bee9cc45ef671180d860a50f595885b4d621b45

chungg:
what happens if metricd loses the message from notifier? does that mean we potentially can lose work?

jd:
@chungg The `metric_processing_delay` is still there, so if no messages are received in this time frame, metricd will go for checking all the sacks old-school-style.

jd:
Pull-request updated, HEAD is now 94f72dc1087055215c5810ce27b84aa6abca6b6e

jd:
How do you handle a queue with any of our current storage backend?
Do we want us to pull oslo.messaging? lol

chungg:
yeah, all the exisitng incoming drivers would not work (except redis which really isn't a queue.)

my concern is that i'm not sure how much we're getting for adding all this extra stuff (2x writes). we now have 2 routes that overlap (a lot imo)

gnocchi/notifier/redis.py
```diff
@@ -0,0 +1,51 @@
...
+        return "%s: %s channel %s" % (
+            self.__class__.__name__, self._client, self.channel
+        )
+
+    def iter_on_sacks_to_process(self):
+        p = self._client.pubsub()
+        p.subscribe(self.channel)
+        for message in p.listen():
+            if (message['type'] == 'message' and
+               message['channel'] == self.NOTIFY_CHANNEL_NAME):
```
chungg: NOTIFY_CHANNEL_NAME doesn't exist anywhere when i ctrl+f

gnocchi/notifier/redis.py
```diff
@@ -0,0 +1,51 @@
...
+                data = message['data']
+                nb_sack = data / self._SERIALIZATION_FORMAT_LEN
+                for sack in struct.unpack("<" + "H" * nb_sack):
+                    yield sack
+
+    def notify_new_measures_for_sacks(self, sacks):
+        self._client.publish(self.channel,
+                             struct.pack("<" + "H" * len(sacks), *sacks))
```
chungg: it's not entirely clear but it seems like all the workers will receive notification and then race to lock sack? is there a way we can get around this? https://redis.io/topics/notifications maybe?

chungg:
or use blpop? https://redis.io/commands/blpop

gnocchi/notifier/redis.py
jd: Right. The first iteration had a hard coded channel name. Fixed.

gnocchi/notifier/redis.py
jd: All the replicas will race, yes.

A solution for that could be to have only one of the replica react on notification, where the others would only try to grab the sack based at polling.

https://redis.io/topics/notifications has no link with that.

jd:
Pull-request updated, HEAD is now c2b0c7092f95b0ecd489cbb2cd134199cf6cb876

jd:
* Why do want to use `BLPOP` for?
* The complexity of the pub/sub here is O(number of metricd workers). There's only one channel being used in this implementation. There is obviously alternatives, but they all have different trade-offs.

gnocchi/notifier/redis.py
chungg: i guess it's two-fold. there's the race of having it published to everyone. and we're also publishing sacks redundantly as well so one message might contain sack2 , and the next might contain sack2 as well so we have 100s of metricd racing to lock sack2 on first message and the same 100s of metricd racing to lock sack2 again on second go.

it's that race which leads me to believe to think we're not saving much vs polling since the subsequent attempts have a very high chance of polling and discovering that some previous worker already just handled entire sack

jd: > 100s of metricd

No, just 3 by default, **if** they are not busy doing other things. Otherwise they will just try later as they will do with current polling.

We're not necessary saving if polling schedule is the same, but with notifier it could be drastically reduced to a several minutes.

jd:
So `BLPOP` does not help. `BRPOPLPUSH` is designed for that https://redis.io/commands/brpoplpush, but has some serious drawback, such as having to handle recovery. If we had `BLTRIM` that'd be great, but it does not exist. :(

gnocchi/notifier/redis.py
chungg: huh? all the workers subscribe to a single channel no? therefore they'd all get the notification to "process sacks x, y, z"

jd: They will all get notifications, but they are filtered based on the which sack they are responsible for:

https://github.com/gnocchixyz/gnocchi/pull/188/commits/976466086d0e9e889d125458f0a0c5ef88c5ccf9#diff-38284cdc2b2cddcaef8202cd07eb3060R258

chungg: is there a reason we have this filtering at all? why do we care which of the workers handles the sack?

gnocchi/cli.py
```diff
@@ -228,7 +240,35 @@ def run_watchers():
...
+            return
+
+        try:
+            for sack in n.iter_on_sacks_to_process():
+                if sack in self._get_sacks_to_process():
+                    LOG.debug(
+                        "Got notification for sack %s, waking up processing",
+                        sack)
+                    self.sacks_with_measures_to_process.add(sack)
+                    self.wakeup()
```
chungg: curious, but isn't it simpler to just have this trigger it's on code to process sack rather than hijacking the periodic polling job. seems like we end up with a lot of conditionals in _run_job to discover what triggered it.

gnocchi/cli.py
```diff
@@ -228,7 +240,35 @@ def run_watchers():
...
+            # Terminate the thread
+            return
+
+        try:
+            for sack in n.iter_on_sacks_to_process():
+                if sack in self._get_sacks_to_process():
+                    LOG.debug(
+                        "Got notification for sack %s, waking up processing",
+                        sack)
+                    self.sacks_with_measures_to_process.add(sack)
```
chungg: curious, but isn't it simpler to just have this trigger it's on code to process sack rather than hijacking the periodic polling job. seems like we end up with a lot of conditionals in _run_job to discover what triggered it.  we wouldn't have to worry about ensuring sacks_with_measures_to_process is thread safe... 

also, when we run self.sacks_with_measures_to_process.discard(s) below, that doesn't seem like it's safe since we can easily just process a sack with 100metrics but  during processing a bunch of new measures for metrics processed earlier in sack our added.

gnocchi/notifier/redis.py
jd: So we don't have 100s of workers trying to lock the same sack, as you just suggested?

gnocchi/cli.py
jd: It is thread-safe because the only shared data structure is a set. Waking-up the main thread is the easiest and "safest" solution.

The discard call is safe, but you're right that another notification can come in the mean time. There's no way to be sure which stage of processing might. There is already a possibility that a sack with N measures is being processed while more measures are being written (remember writing measures is lock-free).

That's why the polling system is kept in place. Notification is just a good way to cut-down the processing interval. There's no good reason to have metricd being sleeping for `metric_processing_delay` when you can know there's jobs to do.

gnocchi/notifier/redis.py
chungg: or we have channels per sack or for range of sacks so we aren't wasting time broadcasting to subscribers we know don't need. as you mentioned earlier (i think), publish sequentially publishes to all subscribers so the more subscribers, the slower publish is, and the slower our API post is.

not saying this is better, just asking if it's considered. i don't want to see us slowing down API anymore as that is what 'user experience' is.

jd: Haha, it took you time to get on board, but here you come! :)

So yeah that is another solution to have one channel per sack, so that'd avoid spamming all the metricd and then filter. The only downsides that I see is that 1. it is trickier to implement because the sack list changes each time a worker join/leaves the tooz group and 2. it can therefore miss a little bit more notification than the current solution.

That means the subscription patterns must be regularly checked and updated as soon as tooz group member list changes.

I am currently not sure it is worth it – or not.

As for the API, `NOTIFY` should not slow anything down AFAIU. There might be also a way to leverage https://redis.io/topics/notifications which would avoid doing the `NOTIFY` call at all. I did not pursue this yet as 1. it's an optimization 2. at first glance it was not obvious it is something that you can control from the Redis client.

doc/source/architecture.rst
```diff
@@ -35,6 +40,12 @@ The *indexer* is responsible for storing the index of all resources, archive
 policies and metrics, along with their definitions, types and properties. The
 indexer is also responsible for linking resources with metrics.
 
+The *coordinator* is responsible of the scheduling of jobs between *metricd*
```
chungg: s/responsible of the scheduling/responsible for the division/

doc/source/architecture.rst
```diff
@@ -35,6 +40,12 @@ The *indexer* is responsible for storing the index of all resources, archive
...
+workers.
+
+The *notifier* is responsible of letting me the worker that they have metrics
+to compute.
```
chungg: s/responsible of letting me the worker that they have metrics to compute/responsible for event-driven processing of incoming metrics/

i think we should be more explicit that this is an alternative to polling. for example, we should say if you enable notificatier, you should set polling to be less.

gnocchi/cli.py
```diff
@@ -140,11 +147,12 @@ def run(self):
...
             with utils.StopWatch() as timer:
                 self._run_job()
-            self._shutdown.wait(max(0, self.interval_delay - timer.elapsed()))
+            self._wake_up.wait(max(0, self.interval_delay - timer.elapsed()))
+            self._wake_up.clear()
         self._shutdown_done.set()
 
     def terminate(self):
-        self._shutdown.set()
+        self.wakeup()
```
chungg: i don't understand why you are calling wakeup on terminate?

jd: To stop the waiting on L150.

doc/source/architecture.rst
```diff
@@ -20,9 +20,14 @@ are scalable. Additional workers can be added depending on load.
...
-Gnocchi uses three different back-ends for storing data: one for storing new
-incoming measures (the incoming driver), one for storing the time series (the
-storage driver) and one for indexing the data (the index driver).
+Gnocchi uses five different back-ends that you will have to configure. They are
+used to:
+
+- Store incoming measures (incoming section)
+- Store archived measures (storage section)
+- Store resource index (indexer section)
+- Coordinate workers (coordination_url in storage/incoming sections)
```
jd: fixed

doc/source/architecture.rst
jd: fixed

doc/source/architecture.rst
jd: fixed

jd:
Pull-request updated, HEAD is now 8f9bdcc781f96f39656aaa6cfdefecef3b5cd137

jd:
@chungg After re-thinking about it, I'd suggest to modify this design to keep incoming+notifier merged.

This has the downside to not have a notifier for e.g. Swift as incoming. But it has the upside of not complicating the architecture to much with yet-another-thing-to-configure. It can also make an optimization of not having a `PUBLISH` call on Redis but listening to keys changes.

WDYT?

chungg:
@jd seems like a good idea. does that mean for redis we don't poll at all? or only on start?

jd:
@chungg No, it won't be synchronous, and Redis is not a queue system, so it'll be the same functioning that it is in this patch. You'll be able to lower the polling though since it'll be only useful when a measure has been missed while the sack being processed.

Making it synchronous and removing polling would require locking the sack before sending any notification (to be sure no processing is being done), and it's not something we want.

But it'd be a first ground for a synchronous incoming driver like e.g. Kafka.

chungg:
synchronous? if we just use redis notifications, we can listen against all sack keys, we can just trigger processing like it does now no? i guess polling is because redis pubsub does not guarantee delivery?

jd:
> synchronous? if we just use redis notifications, we can listen against all sack keys, we can just trigger processing like it does now no? i guess polling is because redis pubsub does not guarantee delivery?

Yeah sorry for the term _synchronous_ which was probably mischosen. This is exactly what I meant.

I'll rework the branch this way then.

jd:
Pull-request updated, HEAD is now fe9ec621fa396014ece093ec376421ac974a90c8

jd:
Closing for now, I'll reopen a new one when I've implemented what I want. Which will not be a new subsystem.

PR #189, opened 2017-07-07
tests: don't use wait without pid

sileht:
Without pid, wait always return 0.

This change tracks pids to ensure we have the return code
of the background jobs.


sileht:
Merged with #190 to ensure I fix the ceph driver.

PR #190, opened 2017-07-07
ceph: Remove the async read of incoming measures

sileht:
We don't have clue for now, but when rados_osd_op_timeout is set,
aio_read seems to return corrupted data or the add_measures callback
does not work as expected.

This change moves back to the synchronous read of the measures like in
3.0.

This code is gone in 4.0 and later, because we use omap attributes
instead.

### DIFF
--- a/gnocchi/storage/incoming/ceph.py
+++ b/gnocchi/storage/incoming/ceph.py
@@ -15,7 +15,6 @@ from collections import defaultdict
 import contextlib
 import datetime
 import errno
-import functools
 import itertools
 import uuid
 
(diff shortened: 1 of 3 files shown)


sileht:
Pull-request updated, HEAD is now c2a499972b195d91d17d1d3277dc0db26c0e073d

sileht:
Pull-request updated, HEAD is now 5e01311093649eb9390ed40ed4e22bb7f463f865

jd:
@chungg  TL;DR we really need this to get merged

We spent a delightful afternoon with @sileht debugging that. The current CI for `stable/3.1` is broken (because of the non-backported wait patch) and left the patch setting the ceph_timeout default value being merged.

The timeout bug is present since 3.1.0 anyway, we just never met it before. @sileht will report the bug to Ceph next week.

sileht:
Pull-request updated, HEAD is now 0742a1b8f93ec70ddc8bb12430443d418a5c9291

sileht:
Pull-request updated, HEAD is now 50ae4fb58270108972247f0c08b1e9ed712fd684

PR #192, opened 2017-07-08
doc: do not specify just `pip install`

jd:
Nothing's gonna work with just "pip install gnocchi" anyway.


jd:
Pull-request updated, HEAD is now 4a48ba3e7cfd0d352f08b9f7ad09b330a36893d2

PR #193, opened 2017-07-08
utils: rely on numpy rather than Pandas to handle timestamps

jd:
This leverages numpy and datetime64 types rather than pandas.
Let's make this part of the ongoing process of getting rid of Pandas at
somepoint.

### DIFF
--- a/gnocchi/utils.py
+++ b/gnocchi/utils.py
@@ -19,7 +19,6 @@ import distutils.util
 import errno
 import itertools
 import multiprocessing
-import numbers
 import os
 import uuid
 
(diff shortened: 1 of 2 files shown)


gnocchi/utils.py
```diff
@@ -103,22 +102,26 @@ def get_coordinator_and_start(url):
...
-              is_valid_timestamp(values[0])):
-            times = pd.to_datetime(values, utc=True, box=False)
+        if isinstance(values[0], (numpy.datetime64, datetime.datetime)):
+            times = numpy.array(values)
         else:
             try:
                 float(values[0])
             except ValueError:
-                times = (utcnow() + pd.to_timedelta(values)).values
+                if is_valid_timestamp(values[0]):
```
chungg: can you add a note to explain what each of these conditions is trying to handle? it seems non-trivial aside from line105

jd: Sure, it was not obvious to me too when hacking it :))

jd:
Pull-request updated, HEAD is now dc4bf9faafcee030dcd46c35a9f9d97873588237

Issue #195, opened 2017-07-12
Do not return "Invalid input: not a valid value for dictionary value" in REST API


jd:
This error is the generic voluptuous error. A custom error should be returned.

PR #197, opened 2017-07-13
Mimimize Pandas usage and use Numpy types

jd:
carbonara: use numpy native types for computing
tests: fix gabbi test title
carbonara: simplify unserialize
carbonara: use sampling from SplitKey
storage: return SplitKey objects when listing split keys
utils: fix to_timestamps when value list is empty

### DIFF
--- a/gnocchi/archive_policy.py
+++ b/gnocchi/archive_policy.py
@@ -18,9 +18,12 @@ import collections
 import datetime
 import operator
 
+import numpy
 from oslo_config import cfg
 from oslo_config import types
 import six
+
+from gnocchi import utils
 
 
 class ArchivePolicy(object):
@@ -71,7 +74,7 @@ class ArchivePolicy(object):
             raise ValueError(
                 "More than one archive policy "
                 "uses granularity `%s'"
-                % duplicate_granularities[0]
+                % utils.timespan_total_seconds(duplicate_granularities[0])
             )
 
         if aggregation_methods is None:
@@ -155,20 +158,25 @@ class ArchivePolicyItem(dict):
                 raise ValueError(
                     u"timespan ≠ granularity × points")
 
-        if granularity is not None and granularity <= 0:
-            raise ValueError("Granularity should be > 0")
+        if granularity is not None:
+            if not isinstance(granularity, numpy.timedelta64):
+                granularity = numpy.timedelta64(int(granularity * 10e8), 'ns')
+            if granularity <= numpy.timedelta64(0, 'ns'):
+                raise ValueError("Granularity should be > 0")
 
         if points is not None and points <= 0:
             raise ValueError("Number of points should be > 0")
+
+        if (timespan is not None
+           and not isinstance(timespan, numpy.timedelta64)):
+            timespan = numpy.timedelta64(int(timespan * 10e8), 'ns')
 
         if granularity is None:
             if points is None or timespan is None:
                 raise ValueError(
                     "At least two of granularity/points/timespan "
                     "must be provided")
-            granularity = round(timespan / float(points))
-        else:
-            granularity = float(granularity)
+            granularity = timespan / float(points)
 
         if points is None:
             if timespan is None:
@@ -201,11 +209,25 @@ class ArchivePolicyItem(dict):
         """Return a dict representation with human readable values."""
         return {
             'timespan': six.text_type(
-                datetime.timedelta(seconds=self.timespan))
+                datetime.timedelta(
+                    seconds=utils.timespan_total_seconds(
+                        self.timespan)))
             if self.timespan is not None
             else None,
             'granularity': six.text_type(
-                datetime.timedelta(seconds=self.granularity)),
+                datetime.timedelta(
+                    seconds=utils.timespan_total_seconds(
+                        self.granularity))),
+            'points': self.points,
+        }
+
+    def serialize(self):
+        return {
+            'timespan': None
+            if self.timespan is None
+            else float(utils.timespan_total_seconds(self.timespan)),
+            'granularity': float(
+                utils.timespan_total_seconds(self.granularity)),
             'points': self.points,
         }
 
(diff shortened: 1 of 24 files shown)


jd:
Pull-request updated, HEAD is now d83d4cdf8eea1b96d0ba4561e560afe35bb9d8dd

jd:
Pull-request updated, HEAD is now 1dd53e251003d2766312e9f4cf42d06e2dccbd4c

jd:
Pull-request updated, HEAD is now 0a22420228a8606922d9dc0dee0dc62fd7ee2dff

jd:
Pull-request updated, HEAD is now d1f2e2cc928671061f2967b7a811b9f5d155ac04

jd:
Pull-request updated, HEAD is now a17db4114c80333a86ca80d6cf52b689f3cd4f3e

gnocchi/rest/__init__.py
```diff
@@ -145,7 +145,10 @@ def PositiveNotNullInt(value):
 
 
 def Timespan(value):
```
sileht: Do we still need that?

gnocchi/storage/__init__.py
```diff
@@ -254,10 +254,12 @@ def get_cross_metric_measures(metrics, from_timestamp=None,
         for metric in metrics:
             if aggregation not in metric.archive_policy.aggregation_methods:
                 raise AggregationDoesNotExist(metric, aggregation)
-            if (granularity is not None and granularity
-               not in set(d.granularity
-                          for d in metric.archive_policy.definition)):
-                raise GranularityDoesNotExist(metric, granularity)
+            if granularity is not None:
```
sileht: unless change ?

gnocchi/archive_policy.py
```diff
@@ -155,20 +158,25 @@ def __init__(self, granularity=None, points=None, timespan=None):
                 raise ValueError(
                     u"timespan ≠ granularity × points")
 
-        if granularity is not None and granularity <= 0:
-            raise ValueError("Granularity should be > 0")
+        if granularity is not None:
+            if not isinstance(granularity, numpy.timedelta64):
```
jd: Agreed. REST does the conversion but SQLAlchemy JSON type does not, it only unserialize the JSON and returns integer. So I can convert integer to Numpy type in SQLAlchemy but I'm not sure it's better than doing that here.

gnocchi/rest/__init__.py
jd: Yes in order to return a proper error from voluptuous and not have https://github.com/gnocchixyz/gnocchi/issues/195

gnocchi/storage/__init__.py
jd: IIRC it was not because set uses `__hash__` and not `__eq__` for comparison and the `__hash__` function for `SplitKey` was shit. But I fixed it so it might work correctly again?
Anyway my break-based-loop _can_ be faster than using a set that iterate over all archive policy so I ask for a laisser-passer on this one. :-P

gnocchi/rest/__init__.py
sileht: gotcha

jd:
@sileht is there anything you still want me to change?

sileht:
No it's ok

gnocchi/storage/_carbonara.py
```diff
@@ -117,9 +118,15 @@ def _store_metric_measures(metric, timestamp_key, aggregation,
                                granularity, data, offset=None, version=3):
         raise NotImplementedError
 
-    @staticmethod
-    def _list_split_keys_for_metric(metric, aggregation, granularity,
+    def _list_split_keys_for_metric(self, metric, aggregation, granularity,
                                     version=3):
+        return set(map(
+            functools.partial(carbonara.SplitKey, sampling=granularity),
+            utils.to_timestamps(self._list_split_keys(
```
chungg: this seems strange. _list_split_keys will give back epoch time in seconds, then to_timestamps will convert to nanoseconds... shouldn't we just pass _list_split_keys straight to SplitKey?

gnocchi/carbonara.py
```diff
@@ -407,6 +410,10 @@ def __init__(self, value, sampling):
             self.key = value.key
         elif isinstance(value, pandas.Timestamp):
             self.key = value.value / 10e8
+        elif isinstance(value, numpy.datetime64):
```
chungg: what path would give splitkey a datetime64? having a little trouble figuring this out.

gnocchi/storage/_carbonara.py
jd: `_list_split_keys` returns time in seconds with nanoseconds precision.

> shouldn't we just pass _list_split_keys straight to SplitKey?

SplitKey wants a timestamps (= e.g `numpy.dateteim64`) as first arg. I can use `numpy.datetime64(int(x * 10e8, 'ns')` but it's what `to_timestamps` will do so I thought it's be simpler to just use it.

chungg: the following is my understanding.

`+    def _list_split_keys(self, metric, aggregation, granularity, version=3):
          key = self._metric_key(metric)
          if not self._client.exists(key):
              raise storage.MetricDoesNotExist(metric)
         split_keys = set()
         hashes = self._client.hscan_iter(
             key, match=self._aggregated_field_for_split(aggregation, '*',
                                                         granularity, version))
         for f, __ in hashes:
             meta = f.decode("utf8").split(self.FIELD_SEP, 1)
             split_keys.add(meta[0])
         return split_keys`

this seems to be just returning a set of strings representing epoch time in seconds. if we pass it string to splitkey, it will correctly convert it to epoch time in seconds cast to float: https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/carbonara.py#L411

it seems to_timestamps unnecessarily converts to nanosecond float, https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/utils.py#L129... maybe i'm missing something.

jd: Yes, you miss that `gnocchi.carbonara.SplitKey` has been changed in this patch. :)

chungg: which patch? my comments stop at 2nd commit https://github.com/gnocchixyz/gnocchi/pull/197/commits/0641fd109ac4875d281bfe72d48c672757815ff5

chungg: i also noticed, the 2nd commit adds 

```
 +        elif isinstance(value, numpy.datetime64):
 +            self.key = (
 +                (value - UNIX_UNIVERSAL_START64) / numpy.timedelta64(1, 's')
 +            )
```

but it disappears eventually.

jd: https://github.com/gnocchixyz/gnocchi/pull/197/commits/a130bb4ef88047e72750087633f16d79251f1417#diff-d99450bd1e2fa5016cec0a79f33d77afL408

jd:
Pull-request updated, HEAD is now 97b144470ea290110102bba2a8b18b929c7b09dd

jd:
Pull-request updated, HEAD is now 0af9451b2a0ee29f718c1f7e1750ab0161144fc6

jd:
Pull-request updated, HEAD is now 2b95b6282ecce36f5208351055bf7a83c4fdc17c

gnocchi/carbonara.py
```diff
@@ -408,22 +406,17 @@ class SplitKey(object):
...
-            self.key = value.value / 10e8
-        elif isinstance(value, numpy.datetime64):
-            self.key = (
-                (value - UNIX_UNIVERSAL_START64) / numpy.timedelta64(1, 's')
-            )
         else:
-            self.key = float(value)
+            self.key = value
 
         self.sampling = sampling
```
chungg: is there a reason you are changing sampling to be a numpy.timedelta and not just float representing seconds? or i guess is there a reason why it only accepts timedelta as sampling? is it possible to split that into another patch to possible shrink this patch?

jd: We definitely want a numpy type. But I'll try to split that out.

jd:
I need to rework the whole branch anyway since the first patch has been heavily modified in https://github.com/gnocchixyz/gnocchi/pull/208

I'll let that branch be for the time being as a "backup" but it's definitely not mergeable now. :)

Issue #17, opened 2017-05-18
Make measures aggregation start & stop parameters mandatory


jd:
See https://bugs.launchpad.net/gnocchi/+bug/1676902

BigChief45:
Should these parameters still be made mandatory? From the documentation:

> Also, aggregation across metrics have different behavior depending on whether boundary values are set (‘start’ and ‘stop’) and if ‘needed_overlap’ is set.
> 
> **If boundaries are not set**, Gnocchi makes the aggregation only with points at timestamp present in all timeseries. When boundaries are set, Gnocchi expects that we have certain percent of timestamps common between timeseries, this percent is controlled by needed_overlap (defaulted with 100%). If this percent is not reached an error is returned.

jd:
@chungg is #200 good enough to close this one?

chungg:
*shrugs* we can enforce boundaries if you're ok for me to delete a test... it will also break api agreement but i don't think it matters much as the result you get is unpredictable if you don't have boundaries and you have holes in series.

i'd say good enough and we can fix if someone else raises issue again.

sileht:
I have never send the patch to break the API, because of Openstack policies, but now :) This behavior really confuse people.

jd:
I am in favor of raising an error rather than sending random data that make no sense. So go ahead.

chungg:
@sileht it is very confusing. that's why i added a note that says my explanation of expected behaviour might be wrong in some cases.lol

ok. i'll try sending something if someone else hasn't done it by next week.

PR #200, opened 2017-07-14
document open-ended aggregation behaviour

chungg:
not setting a boundary on aggregates results in some unpredicatable
behaviour. make it a bit more clear what is happening and potential
risks


chungg:
this relates to https://github.com/gnocchixyz/gnocchi/issues/17. i can add support to enforce boundary if overlap is specified to solve the unpredictable behaviour... but it also breaks current api... and this test https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/tests/test_rest.py#L1623

PR #201, opened 2017-07-17
Allow to transform measures (absolute, negative, resample)

sileht:
Allow to transform measures
use map instead of enumeration

### DIFF
--- a/gnocchi/rest/__init__.py
+++ b/gnocchi/rest/__init__.py
@@ -1588,7 +1588,8 @@ class AggregationResourceController(rest.RestController):
     @pecan.expose('json')
     def post(self, start=None, stop=None, aggregation='mean',
              reaggregation=None, granularity=None, needed_overlap=100.0,
-             groupby=None, fill=None, refresh=False, resample=None):
+             groupby=None, fill=None, refresh=False, resample=None,
+             transform=None):
         # First, set groupby in the right format: a sorted list of unique
         # strings.
         groupby = sorted(set(arg_to_list(groupby)))
@@ -1612,7 +1613,8 @@ class AggregationResourceController(rest.RestController):
                                    for r in resources)))
             return AggregationController.get_cross_metric_measures_from_objs(
                 metrics, start, stop, aggregation, reaggregation,
-                granularity, needed_overlap, fill, refresh, resample)
+                granularity, needed_overlap, fill, refresh, resample,
+                transform)
 
         def groupper(r):
             return tuple((attr, r[attr]) for attr in groupby)
@@ -1626,7 +1628,8 @@ class AggregationResourceController(rest.RestController):
                 "group": dict(key),
                 "measures": AggregationController.get_cross_metric_measures_from_objs(  # noqa
                     metrics, start, stop, aggregation, reaggregation,
-                    granularity, needed_overlap, fill, refresh, resample)
+                    granularity, needed_overlap, fill, refresh, resample,
+                    transform)
             })
 
         return results
@@ -1657,7 +1660,8 @@ class AggregationController(rest.RestController):
                                             reaggregation=None,
                                             granularity=None,
                                             needed_overlap=100.0, fill=None,
-                                            refresh=False, resample=None):
+                                            refresh=False, resample=None,
+                                            transform=None):
         try:
             needed_overlap = float(needed_overlap)
         except ValueError:
@@ -1699,6 +1703,9 @@ class AggregationController(rest.RestController):
                 abort(400, e)
 
         if resample:
+            if transform:
+                abort(400, 'transform and resample are exclusive')
+
             if not granularity:
                 abort(400, 'A granularity must be specified to resample')
             try:
@@ -1714,6 +1721,9 @@ class AggregationController(rest.RestController):
             except ValueError as e:
                 if fill != 'null':
                     abort(400, "fill must be a float or \'null\': %s" % e)
+
+        if transform is not None:
+            transform = TransformSchema(transform)
 
         try:
             if strtobool("refresh", refresh):
@@ -1732,10 +1742,11 @@ class AggregationController(rest.RestController):
                 # metric
                 return pecan.request.storage.get_measures(
                     metrics[0], start, stop, aggregation,
-                    granularity, resample)
+                    granularity, resample, transform)
             return pecan.request.storage.get_cross_metric_measures(
                 metrics, start, stop, aggregation,
-                reaggregation, resample, granularity, needed_overlap, fill)
+                reaggregation, resample, granularity, needed_overlap, fill,
+                transform)
         except storage.MetricUnaggregatable as e:
             abort(400, ("One of the metrics being aggregated doesn't have "
                         "matching granularity: %s") % str(e))
@@ -1750,7 +1761,7 @@ class AggregationController(rest.RestController):
     def get_metric(self, metric=None, start=None, stop=None,
                    aggregation='mean', reaggregation=None, granularity=None,
                    needed_overlap=100.0, fill=None,
-                   refresh=False, resample=None):
+                   refresh=False, resample=None, transform=None):
         if pecan.request.method == 'GET':
             try:
                 metric_ids = voluptuous.Schema(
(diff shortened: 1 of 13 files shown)


gnocchi/rest/__init__.py
```diff
@@ -375,6 +375,21 @@ def MeasuresListSchema(measures):
         times.tolist(), values))
 
 
+VALID_TRANSFORMATION_METHODS = ["absolute", "negative"]
+
+
+def TransformSchema(transform):
```
jd: You can and should use voluptuous here.

sileht: We parse a string and not a python structure. So pyparsing looks better for this use case.

Also the code is really simple as-is, adding pyparsing now make this change complex for no value.

sileht:
Pull-request updated, HEAD is now e5bdccc5dbfd706a3cc04a12fc01355bba0328bb

sileht:
Pull-request updated, HEAD is now f1810b3ddc4ed75ce373d8c9e36edbbd63791377

sileht:
Pull-request updated, HEAD is now 02b971f97a8932cc1f72eaa9101f6a4f69745fa7

sileht:
Pull-request updated, HEAD is now 5c0fdad755f3ece2fb40fc0284a6aaabf5949ae1

sileht:
What about a transformation like:

absolute:resample(60):negative

chungg:
how hard is it to support absolute:resample(60):negative? i imagine it's difficult since i think even now, resample and reaggregate are done in hardcoded order if you provide both.

we can add this later. i imagine we'd have to figure out how to handle 

> http://url?resample=120&transform=absolute:resample(60):negative

i guess if you put resample or reaggregate in both we just raise error?

sileht:
Pull-request updated, HEAD is now 232ff48730310d8c76e6fab3320791aedc2a6510

sileht:
Pull-request updated, HEAD is now 01928d36099a50f3732dec634da17c5b2b930d6e

sileht:
Pull-request updated, HEAD is now bd341d86008a9766582459c6ec866e9594e61e7c

sileht:
Pull-request updated, HEAD is now fc37fc52384fabc0b64c7b5a818df69bed1f8f6d

sileht:
Pull-request updated, HEAD is now 6cb850168401bfdbdce6895f130c6a548b6381f0

sileht:
Pull-request updated, HEAD is now 631743df6a0bbac97e03f0fa215b0149a8051e6b

sileht:
Pull-request updated, HEAD is now 60de71c72f224404feeec0d27ba4fb5f2722aa5b

sileht:
Pull-request updated, HEAD is now eb9ad711fa3ba7db3c78c29bcc6a3382a3d0953c

gnocchi/tests/test_transformation.py
```diff
@@ -0,0 +1,70 @@
...
+            "absolute:negative)",
+            "foobar:",
+            "absolute:",
+            ":absolute",
+            "absolute():negative():",
+            "()",
+            "(",
+            "foobar",
+            "foobar()",
+            "resample()",
```
chungg: what happens if we have resample(5):resample(10)?

releasenotes/notes/tranform-API-da38b196c1c72d44.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    It's now possible to made some transformation to the stored timeseries
+    during the retrieval of the measures. The allowed transformations are
+    absolute and negative.
```
chungg: it's not clear but is negative like the opposite of absolute? ie. -5 is still -5 if negative transform?

gnocchi/tests/test_transformation.py
sileht: It's valid and the series will be resample twice, It's not really useful.

releasenotes/notes/tranform-API-da38b196c1c72d44.yaml
sileht: No, it's  a "x -1", to have the reverse of absolute, you need to use absolute:negative

gnocchi/tests/test_transformation.py
sileht: I will add a test

releasenotes/notes/tranform-API-da38b196c1c72d44.yaml
sileht: It's documented here: https://github.com/gnocchixyz/gnocchi/pull/201/files#diff-3c5a1f8905561bf156b5e5fc97645cb2R120

but if the writting is bad (and I will be not surprise), tell me how I can enhance it.

chungg: so if we negate -5, does it become +5? that would make sense to me

gnocchi/tests/test_transformation.py
chungg: ok. just want to verify if we support or not.

releasenotes/notes/tranform-API-da38b196c1c72d44.yaml
sileht: yes

chungg: ah ok, it wasn't really clear in existing tests. works for me.

gnocchi/rest/__init__.py
jd: pyparsing? But I don't see any pyparsing here.

```python
TransformSchema = voluptuous.Schema([VALID_TRANSFORMATION_METHODS])

# later use
transform = TransformSchema(transform or "")
```
just an idea but it might look neater?

gnocchi/transformation.py
```diff
@@ -0,0 +1,60 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import collections
+import functools
+
+import pyparsing as pp
```
jd: Ah ok now I see pyparsing lol

gnocchi/transformation.py
```diff
@@ -0,0 +1,60 @@
...
+
+from gnocchi import utils
+
+
+Transformation = collections.namedtuple('Transformation', ["method", "args"])
```
jd: define this in `gnocchi.carbonara` because that's where it's really used in the end?

gnocchi/transformation.py
```diff
@@ -0,0 +1,60 @@
+# -*- encoding: utf-8 -*-
```
jd: This seems to be used by the rest API for Carbonara, so I'd suggest moving it to `gnocchi.rest.transformation`.

gnocchi/transformation.py
sileht: done

gnocchi/transformation.py
sileht: done

sileht:
Pull-request updated, HEAD is now 311ae7e19aa82e2c87def105cdff97225bdcdb39

sileht:
Pull-request updated, HEAD is now a0fc9b2ff583110cfc5a4dfd63634127a89b8262

sileht:
Pull-request updated, HEAD is now fd5ac9452d3e63c76c0cee7a46e39d155f4857fb

gnocchi/rest/__init__.py
```diff
@@ -431,13 +442,21 @@ def get_measures(self, start=None, stop=None, aggregation='mean',
             except Exception:
                 abort(400, "Invalid value for stop")
 
+        if transform is not None:
+            transform = TransformSchema(transform)
+
         if resample:
+            # TODO(sileht): This have to be deprecated at some point
```
jd: This has

PR #203, opened 2017-07-17
redis: remove default connection timeout

jd:
There's no good reason to timeout by default if nothing happens. This just
makes reconnection being handled every 30 seconds by default, which sounds like
a bad idea.


gnocchi/common/redis.py
```diff
@@ -104,8 +104,6 @@ def get_client(conf):
         else:
```
sileht: We can remove L74

jd:
Pull-request updated, HEAD is now 50b544a94a7851637cb36fef1c4525f2036be533

gnocchi/common/redis.py
jd: done

PR #204, opened 2017-07-17
Remove worker_sync_rate option

jd:
If you poll e.g every 60s, there's no need to get the member list and update
the partitioner every 30s. Actually the partitioner status should be updated
just before requesting the stack, that'll be perfect.

So let's do that!


jd:
No, I'll take that into account in my notifier patch, don't worry.

PR #208, opened 2017-07-18
storage: return SplitKey objects when listing split keys

jd:
This avoids comparing strings and keep it to timestamp comparison everywhere.

### DIFF
--- a/gnocchi/storage/_carbonara.py
+++ b/gnocchi/storage/_carbonara.py
@@ -16,6 +16,7 @@
 # under the License.
 import collections
 import datetime
+import functools
 import itertools
 import operator
 
@@ -117,9 +118,15 @@ class CarbonaraBasedStorage(storage.StorageDriver):
                                granularity, data, offset=None, version=3):
         raise NotImplementedError
 
-    @staticmethod
-    def _list_split_keys_for_metric(metric, aggregation, granularity,
+    def _list_split_keys_for_metric(self, metric, aggregation, granularity,
                                     version=3):
+        return set(map(
+            functools.partial(carbonara.SplitKey, sampling=granularity),
+            self._list_split_keys(
+                metric, aggregation, granularity, version)))
+
+    @staticmethod
+    def _list_split_keys(metric, aggregation, granularity, version=3):
         raise NotImplementedError
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


gnocchi/storage/_carbonara.py
```diff
@@ -117,9 +118,15 @@ def _store_metric_measures(metric, timestamp_key, aggregation,
                                granularity, data, offset=None, version=3):
         raise NotImplementedError
 
-    @staticmethod
-    def _list_split_keys_for_metric(metric, aggregation, granularity,
+    def _list_split_keys_for_metric(self, metric, aggregation, granularity,
                                     version=3):
+        return set(map(
+            functools.partial(carbonara.SplitKey, sampling=granularity),
+            utils.to_timestamps(self._list_split_keys(
```
chungg:
```
-            utils.to_timestamps(self._list_split_keys(
-                metric, aggregation, granularity, version))))
+            self._list_split_keys(
+                metric, aggregation, granularity, version)))
```

i applied that and all the tests still pass which makes me think it's either not needed or not related to the changes in this commit.

gnocchi/carbonara.py
```diff
@@ -407,6 +410,10 @@ def __init__(self, value, sampling):
             self.key = value.key
         elif isinstance(value, pandas.Timestamp):
             self.key = value.value / 10e8
+        elif isinstance(value, numpy.datetime64):
```
chungg: if i remove this, the only tests that fail are the ones changed as part of this commit. i get the feeling this isn't needed since it's remove in your later in your chain of commits.

gnocchi/storage/_carbonara.py
jd: No it's ground work for the next commits… So I can merge it in the next commits if you prefer but then it'll be really big :p

gnocchi/carbonara.py
jd: ah, I'll check that.

gnocchi/storage/_carbonara.py
chungg: i'd prefer this added when needed. seems safer.

that way, we can (maybe) remove the temporary isinstance(value, numpy.datetime64) check in SplitKey and also the test changes? and if we need them, we can added them in the required commit (if actually needed).

disclaimer, my knowledge of this branch goes up to only https://github.com/gnocchixyz/gnocchi/pull/197/commits/706a9a5ce71c0dcf578cc5af728e3bf17bc389b0

jd: We'll never remove the test change because we want to pass SplitKey objects with numpy types in it.

That's why the patch is big in the end. We change the typo in the tests.

jd: What I can do is to change the test in this patch so rather than having:

`+            carbonara.SplitKey(numpy.datetime64('2016-01-10T00:00:00'), 60),`

we'll have
`+            carbonara.SplitKey(1452384000.0, 60),`

and then the workaround is not needed.

But then i'll have to rechange to

`+            carbonara.SplitKey(numpy.datetime64('2016-01-10T00:00:00'), 60),`

in a later patch.

Is this worth my time? :)

chungg: your next commit already seems to use that format https://github.com/gnocchixyz/gnocchi/pull/197/commits/c85fdd13cbe98af20c747299450bb065371b6913

:p

jd: lol

yeah ok i'll do it anyway I need to fix a failure in Ceph. And since you're doing the review one by one, let's make things correctly!

it's a big patch after all!

jd:
Pull-request updated, HEAD is now 3f91e8d8152dc6ef8e82a11d87d6bd0e5c8f12c3

PR #209, opened 2017-07-18
Pull request for storage-merge-carbonara

jd:
storage: define SackLockTimeoutError as a StorageError
storage: make CorruptionError a StorageError
Merge gnocchi.storage._carbonara into gnocchi.storage

PR #215, opened 2017-07-19
carbonara: use numpy native types for computing

jd:
This replaces various weak data type (e.g. integers for sampling are replaced
by numpy.timedelta64) by native numpy data types.

Pandas data types are replaced by numpy data types where possible.

Change-Id: I923d5d4fedf571deb4dac74a72cdd539c51cc5a9

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -21,7 +21,6 @@ import functools
 import itertools
 import logging
 import math
-import numbers
 import random
 import re
 import struct
@@ -41,6 +40,10 @@ import six
 time.strptime("2016-02-19", "%Y-%m-%d")
 
 LOG = logging.getLogger(__name__)
+
+
+UNIX_UNIVERSAL_START64 = numpy.datetime64("1970", 'ns')
+ONE_SECOND = numpy.timedelta64(1, 's')
 
 
 class BeforeEpochError(Exception):
@@ -73,9 +76,13 @@ class InvalidData(ValueError):
         super(InvalidData, self).__init__("Unable to unpack, invalid data")
 
 
+def datetime64_to_epoch(dt):
+    return (dt - UNIX_UNIVERSAL_START64) / ONE_SECOND
+
+
 def round_timestamp(ts, freq):
-    return pandas.Timestamp(
-        (pandas.Timestamp(ts).value // freq) * freq)
+    return UNIX_UNIVERSAL_START64 + numpy.floor(
+        (ts - UNIX_UNIVERSAL_START64) / freq) * freq
 
 
 class GroupedTimeSeries(object):
@@ -84,18 +91,17 @@ class GroupedTimeSeries(object):
         # duplicate timestamps, it uses numpy.unique that sorted list, but
         # we always assume the orderd to be the same as the input.
         self.granularity = granularity
-        freq = granularity * 10e8
         self.start = start
         if start is None:
             self._ts = ts
             self._ts_for_derive = ts
         else:
             self._ts = ts[start:]
-            start_derive = start - pandas.Timedelta(freq, unit='ns')
+            start_derive = start - granularity
             self._ts_for_derive = ts[start_derive:]
 
-        self.indexes = (numpy.array(self._ts.index,
-                                    numpy.float) // freq) * freq
+        self.indexes = round_timestamp(
+            numpy.array(self._ts.index, dtype=numpy.datetime64), granularity)
         self.tstamps, self.counts = numpy.unique(self.indexes,
                                                  return_counts=True)
 
@@ -219,12 +225,6 @@ class TimeSerie(object):
     def __len__(self):
         return len(self.ts)
 
-    @staticmethod
-    def _to_offset(value):
-        if isinstance(value, numbers.Real):
-            return pandas.tseries.offsets.Nano(value * 10e8)
-        return pandas.tseries.frequencies.to_offset(value)
-
     @property
     def first(self):
         try:
@@ -274,7 +274,7 @@ class BoundTimeSerie(TimeSerie):
 
         """
         super(BoundTimeSerie, self).__init__(ts)
-        self.block_size = self._to_offset(block_size)
+        self.block_size = block_size
         self.back_window = back_window
         self._truncate()
 
@@ -387,7 +387,7 @@ class BoundTimeSerie(TimeSerie):
 
             t0 = time.time()
             for i in six.moves.range(serialize_times):
-                cls.unserialize(s, 1, 1)
+                cls.unserialize(s, ONE_SECOND, 1)
             t1 = time.time()
             print("  Unserialization speed: %.2f MB/s"
                   % (((points * 2 * 8)
@@ -396,9 +396,13 @@ class BoundTimeSerie(TimeSerie):
     def first_block_timestamp(self):
         """Return the timestamp of the first block."""
         rounded = round_timestamp(self.ts.index[-1],
-                                  self.block_size.delta.value)
-
-        return rounded - (self.block_size * self.back_window)
+                                  self.block_size)
+        # FIXME(jd) Return the result as a pandas.Timestamp object as Pandas is
+        # faster at indexing pandas.Timestamp than numpy.timedelta64 objects
+        # for whatever reason
+        return pandas.Timestamp(
+            rounded - (self.block_size * self.back_window)
+        )
 
     def _truncate(self):
         """Truncate the timeserie."""
@@ -424,10 +428,8 @@ class SplitKey(object):
     def __init__(self, value, sampling):
         if isinstance(value, SplitKey):
             self.key = value.key
-        elif isinstance(value, pandas.Timestamp):
-            self.key = value.value / 10e8
         else:
-            self.key = float(value)
+            self.key = value
 
         self.sampling = sampling
 
@@ -435,7 +437,8 @@ class SplitKey(object):
     def from_timestamp_and_sampling(cls, timestamp, sampling):
         return cls(
             round_timestamp(
-                timestamp, freq=sampling * cls.POINTS_PER_SPLIT * 10e8),
+                timestamp,
+                freq=sampling * cls.POINTS_PER_SPLIT),
             sampling)
 
     def __next__(self):
@@ -453,35 +456,48 @@ class SplitKey(object):
         return self
 
     def __hash__(self):
-        return hash(self.key)
+        return hash(
+            str(datetime64_to_epoch(self.key))
+            +
+            str(self.sampling / ONE_SECOND)
+        )
 
     def __lt__(self, other):
         if isinstance(other, SplitKey):
+            if self.sampling != other.sampling:
+                raise TypeError(
+                    "Cannot compare %s with different sampling" %
+                    self.__class__.__name__)
             return self.key < other.key
         if isinstance(other, pandas.Timestamp):
-            return self.key * 10e8 < other.value
-        return self.key < other
+            return pandas.Timestamp(self.key) < other
+        if isinstance(other, numpy.datetime64):
+            return self.key < other
+        raise TypeError("Cannot compare %r with %r" % (self, other))
 
     def __eq__(self, other):
         if isinstance(other, SplitKey):
+            if self.sampling != other.sampling:
+                raise TypeError(
+                    "Cannot compare %s with different sampling" %
+                    self.__class__.__name__)
             return self.key == other.key
         if isinstance(other, pandas.Timestamp):
-            return self.key * 10e8 == other.value
-        return self.key == other
+            return pandas.Timestamp(self.key) == other
+        if isinstance(other, numpy.datetime64):
+            return self.key == other
+        raise TypeError("Cannot compare %r with %r" % (self, other))
 
     def __str__(self):
         return str(float(self))
 
     def __float__(self):
-        return self.key
-
-    def as_datetime(self):
-        return pandas.Timestamp(self.key, unit='s')
+        return datetime64_to_epoch(self.key)
 
     def __repr__(self):
-        return "<%s: %s / %fs>" % (self.__class__.__name__,
-                                   repr(self.key),
-                                   self.sampling)
+        return "<%s: %s / %s>" % (self.__class__.__name__,
+                                  self.key,
+                                  self.sampling)
 
 
 class AggregatedTimeSerie(TimeSerie):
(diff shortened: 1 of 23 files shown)


gnocchi/carbonara.py
```diff
@@ -73,19 +76,23 @@ def __init__(self):
...
 
 
+def datetime64_to_epoch(dt):
+    return (dt - UNIX_UNIVERSAL_START64) / ONE_SECOND
+
+
 def round_timestamp(ts, freq):
-    return pandas.Timestamp(
-        (pandas.Timestamp(ts).value // freq) * freq)
+    return UNIX_UNIVERSAL_START64 + numpy.floor(
```
chungg: UNIX_UNIVERSAL_START64 doesn't seem to do anything... it resolves to 0

```
>>> ts = numpy.array(rng[:5], dtype=numpy.datetime64)
>>> ts
array(['2012-01-01T00:00:00.000000000', '2012-01-01T00:00:01.000000000',
       '2012-01-01T00:00:02.000000000', '2012-01-01T00:00:03.000000000',
       '2012-01-01T00:00:04.000000000'], dtype='datetime64[ns]')
>>> ts - np.datetime64("1970", 'ns')
array([1325376000000000000, 1325376001000000000, 1325376002000000000,
       1325376003000000000, 1325376004000000000], dtype='timedelta64[ns]')
>>> ts.astype(int)
array([1325376000000000000, 1325376001000000000, 1325376002000000000,
       1325376003000000000, 1325376004000000000])
>>> np.datetime64("1970", 'ns').astype(int)
0
```

gnocchi/utils.py
```diff
@@ -140,12 +140,21 @@ def to_timespan(value):
...
         raise ValueError("Timespan must be positive")
-    return datetime.timedelta(seconds=seconds)
+    return seconds
+
+
+_ONE_SECOND = numpy.timedelta64(1, 's')
+
+
+def timespan_total_seconds(td):
+    return td / _ONE_SECOND
```
chungg: just use astype('timedelta64[s]')

i'm assuming you want timedelta64[s]

gnocchi/carbonara.py
jd: This is used for type conversion. The result on right side of `+` is a `numpy.timedelta64`, so in order to convert it, we need to add it to what is the 0 time indeed.

gnocchi/utils.py
jd: >>> numpy.timedelta64(100, 'ns').astype('timedelta64[s]').astype(int)
0

jd:
We don't use of Carbonara and what uses Carbonara. I don't understand what you mean?

chungg:
ie. this. if you look at archive_policy.py, why does ArchivePolicy need to be know anything about numpy? there's a bunch of places ArchivePolicy is either given numpy OR a datetime.timedelta that is just casting numpy.timedelta

jd:
Because Carbonara excepts Numpy types, and all the storage layer is built around Carbonara. There's no need to have intermediary float/int type whereas the type are actually `datetime.timedelta` (which don't have nanoseconds precision – so let's just use `numpy.timedelta64`.

We can merge https://github.com/gnocchixyz/gnocchi/pull/209 first if you feel more comfortable with asserting we depend on Carbonara.

chungg:
i'll keep looking but this unrelated to #209. regardless of #209, everything outside of carbonara is either bytes or string representation of epoch. i would think if a carbonara timeseries wants numpy, it should convert/serialise as it needs on creation. it just looks super strange that archivepolicy has numpy when it's just a number.

jd:
> i'll keep looking but this unrelated to #209. regardless of #209, everything outside of carbonara is either bytes or string representation of epoch.

This is not true with this branch/patch and it's exactly why this patch is *awesome*.

> i would think if a carbonara timeseries wants numpy, it should convert/serialise as it needs on creation. it just looks super strange that archivepolicy has numpy when it's just a number.

It's not a *number*. It's a *timedelta with a nanoseconds precision*. Python does not have it. Numpy has it. So… I use Numpy :)

gnocchi/utils.py
chungg: i don't understand. the functions returns seconds. what do you expect 100ns to resolve to in seconds?

chungg:
> It's not a number. It's a timedelta with a nanoseconds precision. Python does not have it. Numpy has it. So… I use Numpy :)

you sure? i'm pretty sure it's just a string epoch in indexer and in object name/key.

gnocchi/utils.py
jd: It returns an integer of the total in seconds.
`numpy.timedelta64(100, 'ns')` should be converted to `0.00000001`.

jd:
> you sure? i'm pretty sure it's just a string epoch in indexer and in object name/key.

In indexer, native type of MySQL/PostgreSQL are used and then `datetime.timedelta` are returned. We don't have (nor need I'd say) nanoseconds precision so that's good enough. We anyway don't use epoch from the indexer in any storage place, so it's really two different things.

In object name, it's `SplitKey`s that are passed around. To construct the final object name in the driver, they are converted to a string with `str` when, as late as possible, an object name is computed. As soon as a key is read (see `list_split_key`), it's converted to a SplitKey (with a `numpy.datetime64` in it). This work has been done in the first patch of the serie.

So there's nowhere a string/float carried anymore. That's one of the point of this branch, to stop mixing string, float, and datetimes, and have a **correct** type: a timedelta/datetime with nanosecond precision, everywhere. :)

gnocchi/utils.py
chungg: right... but what scenario do you need 0.0000001? you're supporting a scenario which the storage can't support... unless your plan is to support nanosecond aggregates?

jd: We always supported nanoseconds aggregate. :)

sileht:
I also prefer the numpy everywhere, I have never liked the API-string->datetime.datetime->pandas.Timestamp->string/int-for-split-key->numpy.timestamp-for-serialization and the reverse.

pandas is cool when you have ton of mixed unsorted data. But for simple thing like us, ordered two columns table, timestamp/float, it's overkill and pandas does not always choice the good algorithm because it always assume data are mixed and unsorted. We already improve that by removing it from aggregation computing and serialization/unserialization.

This change just goes one step further.

chungg:
ok, im starting to see the number of places we need to do archivepolicy conversion.

that said, i've found another concern. unserialisation speed is consistently, noticeably slower in the benchmarks. testing the aggregated series, it's about ~30% slower but the bounded series benchmark is ~400% slower. i don't know what's happening here but it's visible in the gate and in my local environment.

jd:
> that said, i've found another concern. unserialisation speed is consistently, noticeably slower in the benchmarks. testing the aggregated series, it's about ~30% slower but the bounded series benchmark is ~400% slower. i don't know what's happening here but it's visible in the gate and in my local environment.

Good point. So be aware that most benchmarks are not reliable as they use `random()` and that do changes a lot the performance of the different computing and compression behind. So changes of 20-30% in various test run is expected.

The goal of the benchmark is more to notice drawdown of at 2x and more.

That being said, I found one benchmark that is considerably slow, and it's indeed the `BoundTimeSerie` serialization. After digging a bit, I did not find why it was slower, but I found how to make it as fast as previously:

I've modified this patch to add the following:
```python
        # FIXME(jd) Return the result as a pandas.Timestamp object as Pandas is
        # faster at indexing pandas.Timestamp than numpy.timedelta64 objects
        # for whatever reason
        return pandas.Timestamp(
            rounded - (self.block_size * self.back_window)
        )
```

So we clearly need to continue ditching Pandas…

jd:
Pull-request updated, HEAD is now 1931cf240aa5ce8d19293691e8804b1d28d691a3

chungg:
> Good point. So be aware that most benchmarks are not reliable as they use random() and that do changes a lot the performance of the different computing and compression behind. So changes of 20-30% in various test run is expected.

i don't think that's true. i usually see maybe 5-10% deviation on the same code. especially when looking at stuff like 'all zeros'. i would say 20-30% deviation is notable change in performance.

chungg:
the 4x slow down it better now though :)

jd:
> i don't think that's true. i usually see maybe 5-10% deviation on the same code.

I really saw weird jumps in perf, mainly because of lz4 I presume, when things are random, much more than 5-10%.

> especially when looking at stuff like 'all zeros'. i would say 20-30% deviation is notable change in performance.

For all 0 / all 1 (everything that is not random) that's true, yep (that's what I wrote).

gnocchi/tests/functional/gabbits/archive.yaml
```diff
@@ -561,8 +561,8 @@ tests:
       status: 201
       response_json_paths:
           $.definition[0].points: 1000
-          $.definition[0].granularity: "0:00:04"
-          $.definition[0].timespan: "1:06:40"
+          $.definition[0].granularity: "0:00:03.600000"
+          $.definition[0].timespan: "1:00:00"
```
chungg: i guess this was broken before?

gnocchi/tests/test_carbonara.py
```diff
@@ -19,13 +19,17 @@
...
 import iso8601
-import pandas
+import numpy
 import six
 
 from gnocchi import carbonara
 from gnocchi.tests import base
 
 
+def datetime64(*args):
```
chungg: is this change necessary?

chungg: for the non-assert stuff i mean... i guess it doesn't matter... it's a sh*tload of code regardless

gnocchi/tests/test_storage.py
```diff
@@ -354,34 +398,58 @@ def test_rewrite_measures(self):
         self.trigger_processing()
 
         self.assertEqual({
-            carbonara.SplitKey(1452384000.0, 60),
-            carbonara.SplitKey(1451736000.0, 60),
-            carbonara.SplitKey(1451520000.0, 60),
-            carbonara.SplitKey(1451952000.0, 60),
-        }, self.storage._list_split_keys_for_metric(self.metric, "mean", 60.0))
+            carbonara.SplitKey(numpy.datetime64('2016-01-10T00:00:00'),
```
chungg: maybe keep it as epoch so we can match it with the rest of the test.

gnocchi/tests/test_storage.py
```diff
@@ -404,33 +472,51 @@ def test_rewrite_measures_oldest_mutable_timestamp_eq_next_key(self):
         self.trigger_processing()
 
         self.assertEqual({
-            carbonara.SplitKey(1451520000.0, 60),
-            carbonara.SplitKey(1451736000.0, 60),
-            carbonara.SplitKey(1451952000.0, 60),
-        }, self.storage._list_split_keys_for_metric(self.metric, "mean", 60.0))
+            carbonara.SplitKey(numpy.datetime64('2015-12-31T00:00:00'),
```
chungg: same thing

gnocchi/tests/test_storage.py
```diff
@@ -445,33 +531,56 @@ def test_rewrite_measures_oldest_mutable_timestamp_eq_next_key(self):
         self.trigger_processing()
 
         self.assertEqual({
-            carbonara.SplitKey(1452384000.0, 60),
-            carbonara.SplitKey(1451736000.0, 60),
-            carbonara.SplitKey(1451520000.0, 60),
-            carbonara.SplitKey(1451952000.0, 60),
-        }, self.storage._list_split_keys_for_metric(self.metric, "mean", 60.0))
+            carbonara.SplitKey(numpy.datetime64('2016-01-10T00:00:00'),
```
chungg: same

gnocchi/json.py
```diff
@@ -33,6 +34,15 @@ def to_primitive(obj):
     if isinstance(obj, numpy.datetime64):
         # Do not include nanoseconds if null
         return str(obj).rpartition(".000000000")[0] + "+00:00"
+    if isinstance(obj, numpy.timedelta64):
+        return obj / numpy.timedelta64(1, 's')
+    if isinstance(obj, pandas.Timedelta):
+        # >>> pandas.Timedelta("1 minute").total_seconds()
+        # 60.00000000000001
+        # Never forget that bro.
+        return to_primitive(obj.to_timedelta64())
```
jd: Changed, thanks!

gnocchi/tests/functional/gabbits/archive.yaml
jd: Yes. I read that twice but we always supported nanoseconds aggregation so if you request 1 hours and 1k points this is what will happen. I guess it was not clear when we wrote the test that nanoseconds aggregation was expected to work and no rounding was done.

The underlying aggregation done was already 3.6s so that does not change. We just fix the returned result on this API call.

gnocchi/tests/test_carbonara.py
jd: Yup. It's just search-and-replace everywhere anyway.

gnocchi/tests/test_storage.py
jd: Changed those back to epoch.

gnocchi/tests/test_storage.py
jd: Fixed

gnocchi/tests/test_storage.py
jd: Fixed

jd:
@chungg I digged a bit in the benchmark stuff but did not find anything obvious. The code has barely been edited for serialization here but it seems slower. The slowest operation seems to be `         serial[locs] = values` in serialization, but it's all numpy, before this patch, and after this patch. So maybe a few others operations are slowed down, but it's barely possible to tell.

I tried profiling yesterday but it's mainly optimized (and therefore invisible) numpy calls so it's barely possible to get anything useful.

I suspect it's another Pandas side effect like `BoundTimeSeries`. If you're not comfortable merging this as is, I'm OK to wait until more Pandas is removed and performance are back to (and I hope better) than previously. Though it means it'll make that PR bigger. ;)

jd:
Pull-request updated, HEAD is now 754ce1f1dee5c4cdf867fca7c1e17f89ce61446c

jd:
Rebase because of rate: agg method add

gnocchi/carbonara.py
```diff
@@ -151,7 +157,7 @@ def quantile(self, q):
                                          numpy.percentile,
                                          q=q,
                                      ),
-                                     out_dtype='float64',
+                                     out_dtype='float6y4',
```
chungg: guessing this is stray character

jd:
Fixed typo spotted by gordc

jd:
Fixed tests after rate: patch

jd:
Pull-request updated, HEAD is now e7d080ee71220629ff1558711d9bb0971d74ad80

jd:
It seems I cant get anything working today

PR #220, opened 2017-07-20
Make compatible with tenacity<4.0.0

sileht:


sileht:
Pull-request updated, HEAD is now b2dca112bbb16bd9351deeee8dce2753edec3c1e

PR #225, opened 2017-07-21
indexer: fix ne operator

### LINKED ISSUE #224: resource query doesn't work in some circumstances
For example:

`gnocchi resource search project_id!=foobar`

will return a empty list if one project_id is None.

sileht:
When we use the ne operator, row with NULL values are not returned by
mysql or postgresql.

This change fixes the sql query to return them.

Closes #224

### DIFF
--- a/gnocchi/indexer/sqlalchemy.py
+++ b/gnocchi/indexer/sqlalchemy.py
@@ -1221,7 +1221,11 @@ class QueryTransformer(object):
                             raise indexer.QueryValueError(value, field_name)
                         break
 
-        return op(attr, value)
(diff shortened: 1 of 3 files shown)


sileht:
Pull-request updated, HEAD is now eb89cbe0fdaf5dfa9c542242cf9b5b955a5b3fec

sileht:
Pull-request updated, HEAD is now 343bebf20837eefaf85caa379e6da41509181069

sileht:
Pull-request updated, HEAD is now aa96b8b5eb3f461d766c74d917e043eaf23443fd

PR #228, opened 2017-07-24
incoming: unserialize using numpy rather than Pandas

jd:


chungg:
i'm torn. this improves aggregated timeseries serialisation maybe ~15%... but makes bounded timeseries unserialisation 6x slower... :(

gnocchi/carbonara.py
```diff
@@ -331,8 +331,8 @@ def unserialize(cls, data, block_size, back_window):
 
     def serialize(self):
         # NOTE(jd) Use a double delta encoding for timestamps
-        timestamps = numpy.insert(numpy.diff(self.ts.index),
-                                  0, self.first.value)
+        timestamps = numpy.insert(numpy.diff(self.ts.index), 0,
+                                  int(datetime64_to_epoch(self.first) * 10e8))
```
chungg: maybe use self.first.astype('timedelta64[ns]')? so we don't need to convert to timedelta64[s] and then multiply by 10e8

jd: `self.first` is already a `timedelta64[ns]`. It seems to work without conversion indeed, I've simplified.

jd:
@chungg yeah, I know it's slower, but we'll need this patch to move on.
But I'm ok to wait for the big-final-patch that will replace `pandas.Series` with Numpy, if you want, before merging this part.

I'm trying to go by small increments when possible. ;)

chungg:
i'm indifferent. we just need to be aware of slow down. we already have a significant slowdown from previous numpy patch.

let's make the edit in comment for now.

jd:
which edit?

gnocchi/carbonara.py
chungg: this one. where you said it works without conversion

jd: Oh I though I updated the PR sorry. Facepalm. Sending it now.

jd:
Pull-request updated, HEAD is now cc46876e0ddafa478c63c9b438ff1a8442747ea7

PR #235, opened 2017-07-25
carbonara: make sure all data are list and not just iterables

jd:


jd:
I'm trying to split my next big patch ;)

PR #236, opened 2017-07-25
simplify hashing

chungg:
we don't need math to build a hash

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -453,11 +453,8 @@ class SplitKey(object):
         return self
 
     def __hash__(self):
-        return hash(
-            str(datetime64_to_epoch(self.key))
-            +
-            str(self.sampling / ONE_SECOND)
-        )
+        return hash(str(self.key.astype('datetime64[ns]')) +
+                    str(self.sampling.astype('timedelta64[ns]')))
 
     def __lt__(self, other):
         if isinstance(other, SplitKey):


chungg:
i should probably add, this does next to nothing with regards to performance.

gnocchi/carbonara.py
```diff
@@ -453,11 +453,8 @@ def __iter__(self):
         return self
 
     def __hash__(self):
-        return hash(
-            str(datetime64_to_epoch(self.key))
-            +
-            str(self.sampling / ONE_SECOND)
-        )
+        return hash(str(self.key.astype('datetime64[ns]')) +
```
jd: So it seems that it's better to use `view` than `astype`: `view` does not return a copy, where `astype` does (unless you pass `copy=False`, but then, just use `view`).

chungg: i don't think this works as expected. from what i read, it's just tries to reevaluate based on bytes. it is faster though.

```
In [62]: dt.view()
Out[62]: numpy.datetime64('2017-01-01')

In [63]: dt.view('datetime64[ns]')
Out[63]: numpy.datetime64('1970-01-01T00:00:00.000017167')

In [64]: dt.astype('datetime64[ns]')
Out[64]: numpy.datetime64('2017-01-01T00:00:00.000000000')
```

jd: Oh sh**. Ok. Maybe `copy=False`?
No big deal though.

chungg: hmm. copy=False should work... but it gives me an error with numpy 1.12.1

```
In [85]: td.astype('timedelta64[ns]', copy=False)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-85-17b625fb7ebb> in <module>()
----> 1 td.astype('timedelta64[ns]', copy=False)

TypeError: astype() takes no keyword arguments
```

PR #238, opened 2017-07-25
tests: set logging level to debug when setting GNOCCHI_TEST_DEBUG

jd:
Other than that the tests are only running using WARNING.

### DIFF
--- a/gnocchi/tests/base.py
+++ b/gnocchi/tests/base.py
@@ -15,6 +15,7 @@
 # under the License.
 import functools
 import json
+import logging
 import os
 import subprocess
 import threading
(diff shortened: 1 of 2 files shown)


gnocchi/tests/base.py
```diff
@@ -268,9 +269,15 @@ class TestCase(BaseTestCase):
...
-        self.conf = service.prepare_service([],
-                                            default_config_files=[])
-        if not os.getenv("GNOCCHI_TEST_DEBUG"):
+
+        debug = "GNOCCHI_TEST_DEBUG" in os.environ
+
+        self.conf = service.prepare_service(
+            [],
+            default_config_files=[],
+            logging_level=logging.DEBUG if debug else logging.NOTSET)
```
chungg: what's logging.NOTSET? that resolves to 0 which is less than logging.DEBUG so it would print everything also?

jd: It's pretty well explained here https://docs.python.org/2/library/logging.html

What happens if that we set the log level to NOTSET, the logging level of the handler above (the root one in this case) is used.

There's actually not really any interest of setting NOTSET, since the output is removed anyway. I'll update the patch to simplify.

jd:
Pull-request updated, HEAD is now 042c9dce3e2a1e3ee2819ebd4d7da04b61eddb56

PR #239, opened 2017-07-25
enforce boundaries on overlap

chungg:
if no boundary is set, the returned series does not necessarily
honour the overlap value. change behaviour so we require a boundary
if overlap is not 100%.

Fixes: #17

### DIFF
--- a/gnocchi/rest/__init__.py
+++ b/gnocchi/rest/__init__.py
@@ -1642,6 +1642,9 @@ class AggregationController(rest.RestController):
             needed_overlap = float(needed_overlap)
         except ValueError:
             abort(400, 'needed_overlap must be a number')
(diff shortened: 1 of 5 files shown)


sileht:
This need also a release note, since it breaks api

gnocchi/rest/__init__.py
```diff
@@ -1649,6 +1649,9 @@ def get_cross_metric_measures_from_objs(metrics, start=None, stop=None,
                                             refresh=False, resample=None):
         try:
             needed_overlap = float(needed_overlap)
+            if needed_overlap != 100.0 and start is None and stop is None:
```
jd: Can you put that outside of the `try/except`? It should not be caught as part of this block if it'd ever raise a `ValueError` (unlikely now, but who knows in the future :))

PR #241, opened 2017-07-26
Modify a mistake in plan Gnocchi storage

zhang-shaoman:
In section plan Gnocchi's storage, there is a little mistake in the
example, so fix it.

### DIFF
--- a/doc/source/operating.rst
+++ b/doc/source/operating.rst
@@ -153,7 +153,7 @@ For example, if you want to keep a year of data with a one minute resolution::
 
 Then::
 
-    size in bytes = 525 600 bytes × 6 = 3 159 600 bytes = 3 085 KiB
+    size in bytes = 525 600 points × 8 bytes = 4 204 800 bytes = 4 106 KiB
 
 This is just for a single aggregated time series. If your archive policy uses
 the 6 default aggregation methods (mean, min, max, sum, std, count) with the


doc/source/operating.rst
```diff
@@ -153,7 +153,7 @@ For example, if you want to keep a year of data with a one minute resolution::
 
 Then::
 
-    size in bytes = 525 600 bytes × 6 = 3 159 600 bytes = 3 085 KiB
+    size in bytes = 525 600 bytes × 8 = 4 204 800 bytes = 4 106 KiB
```
jd: why? there's only 6 archive policies enabled by default

zhang-shaoman: to jd,
    because one point uses 8 bytes, so 'size in bytes' just compute the total storage for one archive policy.
    if use 6 archive policy, then the total storage shoule be 'size in bytes' * 6. This is mentioned below in the doc.

jd: Ok, so the code is wrong but the fix too. I thought it was about the number of archive policies, but it's indeed a number of bytes, but the bytes are on the wrong side. It should be:

```size in bytes = 525 600 points × 8 bytes = 4 204 800 bytes = 4 106 KiB```

zhang-shaoman: Yes, this is more accurate, I will modify it again :)

PR #245, opened 2017-07-26
travis: print python versions

sileht:


chungg:
was this for debug only? oh well

PR #247, opened 2017-07-26
carbonara: speed up BoundTimeSerie.unserialize by leveraging frombuffer

jd:
numpy.frombuffer can have offset/count kwargs, so no need to copy the
uncompressed string around.

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -321,14 +321,14 @@ class BoundTimeSerie(TimeSerie):
         nb_points = (
             len(uncompressed) // cls._SERIALIZATION_TIMESTAMP_VALUE_LEN
         )
-        timestamps_raw = uncompressed[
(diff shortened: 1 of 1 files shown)


gnocchi/carbonara.py
```diff
@@ -321,14 +321,14 @@ def unserialize(cls, data, block_size, back_window):
         nb_points = (
             len(uncompressed) // cls._SERIALIZATION_TIMESTAMP_VALUE_LEN
         )
-        timestamps_raw = uncompressed[
-            :nb_points*cls._SERIALIZATION_TIMESTAMP_LEN]
-        timestamps = numpy.frombuffer(timestamps_raw, dtype='<Q')
+        timestamps = numpy.frombuffer(uncompressed, dtype='<Q',
+                                      count=nb_points)
```
sileht: that's so beautiful :)

PR #248, opened 2017-07-26
carbonara: correctly handle all frombuffer errors

jd:
Only a few `numpy.frombuffer` calls are covered by `try/except ValueError`
blocks, whereas they should all be covered and raise InvalidData in case of
error.


jd:
Pull-request updated, HEAD is now 4442cae69bff4d497ac51177d363cda6113c1d82

PR #253, opened 2017-07-27
Pull request for use-numpy-in-carbonara

jd:
carbonara: remove weird key check
Remove most of Pandas usage from Carbonara

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -598,8 +598,6 @@ class AggregatedTimeSerie(TimeSerie):
         start = 0
         for key, count in six.moves.zip(keys, counts):
             end = start + count
-            if key == -0.0:
-                key = abs(key)
             yield (SplitKey(key, self.sampling),
                    AggregatedTimeSerie(self.sampling, self.aggregation_method,
                                        self[start:end]))
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -16,7 +16,6 @@
 # under the License.
 """Time series data manipulation, better with pancetta."""
 
-import datetime
 import functools
 import itertools
 import logging
@@ -85,6 +84,42 @@ def round_timestamp(ts, freq):
         (ts - UNIX_UNIVERSAL_START64) / freq) * freq
 
 
+TIMESERIES_ARRAY_DTYPE = [('timestamps', 'datetime64[ns]'),
+                          ('values', 'float64')]
+
+
+def make_timeseries(timestamps, values):
+    """Return a Numpy array representing a timeseries.
+
+    This array specifies correctly the data types, which is important for
+    Numpy to operate fastly.
+    """
+    l = len(timestamps)
+    if l != len(values):
+        raise ValueError("Timestamps and values must have the same length")
+    arr = numpy.zeros(l, dtype=TIMESERIES_ARRAY_DTYPE)
+    arr['timestamps'] = timestamps
+    arr['values'] = values
+    return arr
+
+
+def combine_timeseries(ts1, ts2):
+    """Combine a timeseries into this one.
+
+    The timeseries does not need to be sorted.
+
+    If a timestamp is present in both `ts1` and `ts2`, then value from `ts1`
+    is used.
+
+    :param ts: The timeseries to combine.
+    :return: A new timeseries.
+    """
+    _, index = numpy.unique(
+        numpy.append(ts1['timestamps'], ts2['timestamps']),
+        return_index=True)
+    return numpy.append(ts1, ts2)[index]
+
+
 class GroupedTimeSeries(object):
     def __init__(self, ts, granularity, start=None):
         # NOTE(sileht): The whole class assumes ts is ordered and don't have
@@ -96,11 +131,13 @@ class GroupedTimeSeries(object):
             self._ts = ts
             self._ts_for_derive = ts
         else:
-            self._ts = ts[start:]
+            self._ts = ts[numpy.searchsorted(ts['timestamps'], start):]
             start_derive = start - granularity
-            self._ts_for_derive = ts[start_derive:]
+            self._ts_for_derive = ts[
+                numpy.searchsorted(ts['timestamps'], start_derive):
+            ]
 
-        self.indexes = round_timestamp(self._ts.index.values, granularity)
+        self.indexes = round_timestamp(self._ts['timestamps'], granularity)
         self.tstamps, self.counts = numpy.unique(self.indexes,
                                                  return_counts=True)
 
@@ -129,25 +166,20 @@ class GroupedTimeSeries(object):
                                      out_dtype='float64',
                                      default=None)
 
-    def _count(self):
-        return (self.counts, self.tstamps)
-
     def count(self):
-        return pandas.Series(*self._count())
+        return make_timeseries(self.tstamps, self.counts)
 
     def last(self):
-        counts, timestamps = self._count()
-        cumcounts = numpy.cumsum(counts) - 1
-        values = self._ts.values[cumcounts]
+        cumcounts = numpy.cumsum(self.counts) - 1
+        values = self._ts['values'][cumcounts]
 
-        return pandas.Series(values, timestamps)
+        return make_timeseries(self.tstamps, values)
 
     def first(self):
-        counts, timestamps = self._count()
-        counts = numpy.insert(counts[:-1], 0, 0)
+        counts = numpy.insert(self.counts[:-1], 0, 0)
         cumcounts = numpy.cumsum(counts)
-        values = self._ts.values[cumcounts]
-        return pandas.Series(values, timestamps)
+        values = self._ts['values'][cumcounts]
+        return make_timeseries(self.tstamps, values)
 
     def quantile(self, q):
         return self._scipy_aggregate(ndimage.labeled_comprehension,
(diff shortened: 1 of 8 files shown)


jd:
Pull-request updated, HEAD is now 9f2df6331bb7925ba578bfbb64f6ff515ed7519b

jd:
Pull-request updated, HEAD is now 0af4ea879ac9cd532f957944199665ed60faf1de

gnocchi/carbonara.py
```diff
@@ -85,6 +84,46 @@ def round_timestamp(ts, freq):
...
+    If a timestamp is present in both `ts1` and `ts2`, then value from `ts1`
+    is used.
+
+    :param ts: The timeseries to combine.
+    :return: A new timeseries.
+    """
+    uniq_tstamps, index = numpy.unique(
+        numpy.append(ts1['timestamps'], ts2['timestamps']),
+        return_index=True)
+    uniq_values = numpy.append(ts1['values'], ts2['values'])[index]
```
chungg: return numpy.append(ts1, ts2)[index]

gnocchi/carbonara.py
```diff
@@ -85,6 +84,46 @@ def round_timestamp(ts, freq):
...
+def make_timeseries(timestamps, values):
+    """Return a Numpy array representing a timeseries.
+
+    This array specifies correctly the data types, which is important for
+    Numpy to operate fastly.
+    """
+    l = len(timestamps)
+    if l != len(values):
+        raise ValueError("Timestamps and values must have the same length")
+    arr = make_timeseries_array(numpy.zeros(l))
```
jd: Yes, this is to make it have the right shape. You can't assign otherwise as the shape would be  (0,) and the values/timestamps would be (number of points,).

I did not find a way to create directly a `numpy.array` based on 2 array. I might be able to use records instead, I'll look into that.

gnocchi/carbonara.py
jd: Fair enough, should be faster I imagine.

gnocchi/carbonara.py
```diff
@@ -189,59 +225,85 @@ class TimeSerie(object):
...
+        return (isinstance(other, TimeSerie) and
+                numpy.all(self.ts == other.ts))
 
     def __getitem__(self, key):
+        if isinstance(key, numpy.datetime64):
+            idx = numpy.searchsorted(self.timestamps, key)
+            if self.timestamps[idx] == key:
+                return self[idx]
+            raise KeyError(key)
+        if isinstance(key, slice):
```
chungg: fancy

gnocchi/carbonara.py
jd: Ok, records kinda sucks too. If you find a way to create an array in any other way, I'll take it!

gnocchi/carbonara.py
```diff
@@ -737,28 +783,20 @@ def fetch(self, from_timestamp=None, to_timestamp=None):
...
         points = SplitKey.POINTS_PER_SPLIT
-        sampling = 5
+        sampling = numpy.timedelta64(5, 's')
         resample = numpy.timedelta64(35, 's')
 
-        now = datetime.datetime(2015, 4, 3, 23, 11)
+        now = numpy.datetime64("2015-04-03 23:11")
         timestamps = numpy.sort(numpy.array(
-            [now + datetime.timedelta(seconds=i*sampling)
+            [now + numpy.timedelta64(i*sampling, 's')
```
chungg: sampling is already a timedelta64

jd:
Pull-request updated, HEAD is now 0fb12aaf714c19ecd1ce8b4f3c2598d5a12fd4f0

gnocchi/carbonara.py
jd: Fixed.

jd:
Pull-request updated, HEAD is now da721caf3db2aec4d9a4e0a4c6df89a51eff6e72

gnocchi/storage/_carbonara.py
```diff
@@ -391,8 +388,6 @@ def _compute_and_store_timeseries(self, metric, measures):
             LOG.debug("Skipping %s (already processed)", metric)
             return
 
-        measures = sorted(measures, key=operator.itemgetter(0))
```
chungg: last question before i look at tests, where is this being sorted?

gnocchi/carbonara.py
```diff
@@ -85,6 +84,42 @@ def round_timestamp(ts, freq):
         (ts - UNIX_UNIVERSAL_START64) / freq) * freq
 
 
+TIMESERIES_ARRAY_DTYPE = [('timestamps', 'datetime64[ns]'),
+                          ('values', 'float64')]
+
+
+def make_timeseries(timestamps, values):
```
sileht: Shouldn't base timeseries creation and manipulation hold into a class for clarity ?

I known you prefer functional code :)

But that may help to read the code.

gnocchi/carbonara.py
```diff
@@ -85,6 +84,46 @@ def round_timestamp(ts, freq):
...
+    l = len(timestamps)
+    if l != len(values):
+        raise ValueError("Timestamps and values must have the same length")
+    arr = make_timeseries_array(numpy.zeros(l))
+    arr['timestamps'] = timestamps
+    arr['values'] = values
+    return arr
+
+
+def combine_timeseries(ts1, ts2):
```
sileht: I remember that I have not removed combinefirst because it was a lot slower than this two numpy lines

sileht: Pandas have an optimization on the indexes that make this operation faster.

jd:
Pull-request updated, HEAD is now 3a8132ab69919f5d4d9f688cba86fac3fbccf9b0

gnocchi/storage/_carbonara.py
jd: It's not. I thought I could optimize that but I can't. That's what it bugged. Thanks!

gnocchi/carbonara.py
jd: I thought about it but `make_timeseries` is used by `GroupedTimeSerie` and `TimeSerie` and they both have no ancestor in common.
Since it's a rather short function that just wraps Numpy for convenience, I decided to let it here, alone. :)

If we ever make `GroupedTimeSeries` inherits from `TimeSerie` (which may or may not make sense) we could do it. Thought it'd still be a `staticmethod` anyway. :)

gnocchi/carbonara.py
jd: Maybe, but overall it seems the whole code is still faster so I'm not really bothered?

I'll let @chungg yell at me if it's way slower than Pandas. :)

chungg: fine! i'll do some local benching... but i ultimately don't think it matters. we'd have to conver to pd.Series to leverage combine_first and pd.Series init is slow as is.

chungg: i think we should be good:

```
In [48]: [blah,blah2]
Out[48]: 
[array([('2017-07-27T13:33:21.000000000',  1.),
        ('2017-07-27T13:34:27.000000000',  2.)],
       dtype=[('timestamps', '<M8[ns]'), ('values', '<f8')]),
 array([('2017-07-27T13:33:27.000000000',  3.)],
       dtype=[('timestamps', '<M8[ns]'), ('values', '<f8')])]

In [49]: %timeit __, index = np.unique(np.append(blah['timestamps'], blah2['timestamps']), return_index=True); np.append(blah, blah2)[
    ...: index]
The slowest run took 8.65 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 22.9 µs per loop
```

vs

```
In [56]: ts = pd.Series(blah['values'], blah['timestamps'])

In [57]: ts2 = pd.Series(blah2['values'], blah2['timestamps'])

In [58]: %timeit ts.combine_first(ts2)
1000 loops, best of 3: 619 µs per loop
```

too lazy to make something with real size

gnocchi/tests/test_carbonara.py
```diff
@@ -84,19 +84,19 @@ def test_duplicate_timestamps(self):
...
                        (datetime64(2014, 1, 1, 13, 0, 11), 4)])
         self.assertEqual(4, len(ts))
-        self.assertEqual(10.0, ts[0])
-        self.assertEqual(23.0, ts[1])
-        self.assertEqual(3.0, ts[2])
-        self.assertEqual(4.0, ts[3])
+        self.assertEqual(10.0, ts[0][1])
+        self.assertEqual(23.0, ts[1][1])
+        self.assertEqual(3.0, ts[2][1])
+        self.assertEqual(9.0, ts[3][1])
```
chungg: uh... this is a bit concerning. this is a blocker i think if we previously guarantee'd or had some guarantee that read in measures based on POST time. redis definitely has this as we form a 'queue'. i don't know if we considered in the timestamps in other drivers.

gnocchi/tests/test_carbonara.py
```diff
@@ -256,10 +256,10 @@ def test_74_percentile_serialized(self):
              (datetime64(2014, 1, 1, 12, 0, 4), 5),
              (datetime64(2014, 1, 1, 12, 0, 9), 6)])
         ts = self._resample(ts, numpy.timedelta64(60, 's'), '74pct')
-        ts.merge(saved_ts)
+        saved_ts.merge(ts)
```
chungg: i don't like how we swapped positions here but i'm not completely worried about this.

jd:
Pull-request updated, HEAD is now 1a4118f7a094ffdeac2e6375971fb028c485acc5

gnocchi/tests/test_carbonara.py
jd: It's actually more logic to me now :)

gnocchi/tests/test_carbonara.py
chungg: ok yeah, i'm ok. i don't think there's any way ceph is honouring ordering as its object name is: measures_<metric_id>_<uuid>_<timestamp>. i guess it is what it is. if you send in identical timestamps. if it's not processed, you'll have no idea which one you'll get.

gnocchi/carbonara.py
sileht: fair enough

gnocchi/tests/test_carbonara.py
chungg: weird, github hides what i typed for object name. measures_metricid_uuid_timestamp. done.

PR #254, opened 2017-07-27
incoming: return directly a Numpy array of measures

jd:
This read the incoming measures as a Numpy array ready to be sent to Carbonara,
rather than a list that needs a cast.

### DIFF
--- a/gnocchi/incoming/_carbonara.py
+++ b/gnocchi/incoming/_carbonara.py
@@ -75,6 +75,15 @@ class CarbonaraBasedStorage(incoming.StorageDriver):
     _SERIALIZE_DTYPE = [('timestamps', '<datetime64[ns]'),
                         ('values', '<d')]
 
+    def _make_measures_array(self):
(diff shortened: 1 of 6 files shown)


jd:
Pull-request updated, HEAD is now 3b3e9f19d1a5a1981b0ff01c070bb20ccdf40bcd

jd:
Pull-request updated, HEAD is now 4824aff970cce20caf88383521089b3a7a7ef6e6

gnocchi/incoming/_carbonara.py
```diff
@@ -75,6 +75,15 @@ def get_sack_lock(coord, sack):
     _SERIALIZE_DTYPE = [('timestamps', '<datetime64[ns]'),
                         ('values', '<d')]
 
+    def _make_measures_array(self):
```
sileht: I'm feeling we should have a class that represent measures. Instead of all these almost static methods in the middle of the storage logic to manipulate the mesures array.

PR #255, opened 2017-07-27
ceph: fix ReportGenerationError raise

jd:
The class is in gnocchi.incoming, not gnocchi.incoming._carbonara


jd:
Pull-request updated, HEAD is now 5a3ebd7e8e0b277b1ba4703df7b131290aba297e

PR #256, opened 2017-07-27
Merge gnocchi.incoming._carbonara into gnocchi.incoming

jd:
There's no driver that is not based on Carbonara, and we don't envision this
happening any time now. There is a lot of assumption everywhere that all
drivers are based on Carbonara anyway, so let's remove that abstraction layer.


jd:
Pull-request updated, HEAD is now 5d01444f78eaffbb0453f5f3582edcf8bc1b8099

jd:
Pull-request updated, HEAD is now f27f2963a54a035189c04d673da209dd2a28a2c2

jd:
Pull-request updated, HEAD is now f1ed59e2e4d948da45f36a5d72d8b98a12cc1dc2

jd:
@gordc lol right. fixed.

Issue #60, opened 2017-06-01
regional incoming storage targets


chungg:
copy of thead discussion here: http://lists.openstack.org/pipermail/openstack-dev/2017-May/117741.html

here's a scenario: i'd like aggregates stored centrally like it does 
currently with ceph/swift/s3 drivers but i want to collect data from 
many different regions spanning globe. they can all hit the same 
incoming storage but:
- that will be a hell of a lot of load
- single incoming storage locality might not be optimal for all regions 
causing the write performance to take longer than needed for a 'cache' 
storage
- sending HTTP POST with JSON payload probably more bandwidth than 
binary serialised format gnocchi uses internally.

i'm thinking it'd be good to support ability to have each region store 
data 'locally' to minimise latency and then have regional metricd agents 
aggregate into a central target. this is technically possible right now 
by just declaring regional (write-only?) APIs with same storage target 
and indexer targets but a different incoming target per region. the 
problem i think is how to handle coordination_url. it cannot be the same 
coordination_url since that would cause sack locks to overlap. if 
they're different, then i think there's an issue with having a 
centralised API (in addition to regional APIs). specifically, the 
centralised API cannot 'refresh'.

1. look at supporting this fully
2. document small, medium, large deployment strategies

chungg:
https://medium.com/@gord.chung/ubiquitous-gnocchi-7e18813acf7f

PR #264, opened 2017-07-28
add hint on deployment strategy

chungg:
we can do special deployment if it's a multi-site/region storage

fixes: #60


jd:
It seems to me the coordinator needs to be reached by the site for some features like `refresh`. Are we sure we want to push that in doc knowing it has drawbacks?

chungg:
yeah, you're right.

we could technically support refresh. but we need to support multiple incoming targets. this means we would need a new option for metricd to grab from multiple incoming i think.

chungg:
so this might actually work. except we need a coordinator and metricd workers per region.

i don't know how we want to satisfy a single coordinated metricd across regions. it' need to be able to connect multiple targets somehow.

chungg:
i'm going to maybe change this to a blog for now

PR #266, opened 2017-07-31
ceph: order unprocessed measures before processing

akrzos:
This change adjusts the way in which Gnocchi with Ceph storage obtains
unprocessed measures for processing. Previously unprocessed measures can
enter a condition where specific metrics are favored which causes some
metrics with unprocessed measures to continuously grow. This continuous
growth of the backlog stored in a Ceph omap object in the form of keys
and a small 16-byte object will eventually cause an OSD to fail. By
sorting the order of unprocessed measures by date timestamp, we gain
three things:

1. Efficiency in that once "realtime" capacity of Gnocchi is exceeded, more
   measures are processed per processing of a metric
2. No more starved metrics, which grow unbounded in backlog until a failure
3. Scalability of Gnocchi before failures is extended above "realtime"
   processing capacity

### DIFF
--- a/gnocchi/storage/ceph.py
+++ b/gnocchi/storage/ceph.py
@@ -202,10 +202,12 @@ class CephStorage(_carbonara.CarbonaraBasedStorage):
         return len(list(self._list_object_names_to_process(object_prefix)))
 
     def list_metric_with_measures_to_process(self, size, part, full=False):
-        names = self._list_object_names_to_process()
         if full:
-            objs_it = names
+            objs_it = self._list_object_names_to_process()
         else:
+            names = sorted(
+                self._list_object_names_to_process(),
+                key=lambda x: x.split('_', 3)[-1])
             objs_it = itertools.islice(names, size * part, size * (part + 1))
         return set([name.split("_")[1] for name in objs_it])
 


chungg:
this was the main issue with ceph driver... the main reason i didn't do this is because i dont know the effect pulling in 1million+ keys. previously, this just took a few hundred/thousand maybe?

akrzos:
> this was the main issue with ceph driver... the main reason i didn't do this is because i dont know the  effect pulling in 1million+ keys. previously, this just took a few hundred/thousand maybe?

So if the names of the unprocessed measures are actually structured with the date timestamp first in the key, the keys will actually be returned sorted already.  So we could have Ceph do the work but then as far as backwards compatibility and migration to the newly structured names would be more difficult or require a more complex migration path.  
I wrote a tiny test script [here](https://gist.github.com/akrzos/d955d67fb3b4d42e6fcd59d62548a236) with output [here](https://gist.github.com/akrzos/b23c2ced422118a2c33412744f2c9091)

So if we have python just do the sorting, we have to pay whatever additional memory is required but we don't have to worry about "upgrading" into the new name structure.  If we have Ceph sort for us, we have to have some sort of migration steps to convert unprocessed measures into the new naming structure I would assume.

I am looking at memory usage post applying this patch and I can say that memory usage when you hit the "failure" condition (backlog growing indefinitely) is far worse than just sorting in python so this is a reasonable trade-off with the data I currently have.

chungg:
i don't think we can change naming. iirc, the naming structure is to allow us to grab all measures of a metric. ceph only allows us to grab keys based on a prefix that can't be wildcarded so that's why metric_id is one of the leading values in object_name and not the timestamp.

i don't mind this solution per se because you're correct, this does address the main issue of measures going unprocessed. i know changing https://github.com/gnocchixyz/gnocchi/blob/stable/3.1/gnocchi/cli.py#L161 also fixes it somewhat but not as well as yours does.

akrzos:
> i don't think we can change naming. iirc, the naming structure is to allow us to grab all measures of a metric. ceph only allows us to grab keys based on a prefix that can't be wildcarded so that's why metric_id is one of the leading values in object_name and not the timestamp.

Ah that make sense now.  Just saved me some time from trying a futile effort of seeing if I could change the names of the unprocessed keys/small objects to have Ceph order it.

> i know changing https://github.com/gnocchixyz/gnocchi/blob/stable/3.1/gnocchi/cli.py#L161 also fixes it somewhat but not as well as yours does.

So we need some more concrete numbers on what an effective setting for `TASKS_PER_WORKER` should be so exposing this as a configuration item would be best IMO right now.  From my current testing, raising `TASKS_PER_WORKER` still does not actually prevent the infinite backlog growth problem but rather just raises the bar of when it can occur.  Once "realtime" capacity is less than the number of new measures, some metrics get starved and you will end up with a failed OSD. 

While attempting to characterize the behavior or `TASKS_PER_WORKER` raised to 256 at larger scales, it seems to "slow down" processing at the tail end of a backlog (backlog really small) - if that doesn't make sense I can always share a graph that better illustrates the witnessed behavior.

Thus I believe to improve scalability, we need a two-fold solution of preventing starved metrics from growing their backlog (Sorted processing of backlog) and the ability to adjust `TASKS_PER_WORKER`.  I can open that p/r too.

chungg:
> So we need some more concrete numbers on what an effective setting for TASKS_PER_WORKER should be so exposing this as a configuration item would be best IMO right now. From my current testing, raising TASKS_PER_WORKER still does not actually prevent the infinite backlog growth problem but rather just raises the bar of when it can occur. Once "realtime" capacity is less than the number of new measures, some metrics get starved and you will end up with a failed OSD.

agreed, it only delays inevitable to change TASKS_PER_WORKER

the reason i'm somewhat hesitant regarding sort solution is because if we're already doing the work of pulling in all the keys, why don't we just schedule them. i imagine we could just pull in all keys and hash out keys. this way we're not just doing a bunch of work and throwing away the majority of it. this would be a less complete solution to what we did in v4.

akrzos:
> the reason i'm somewhat hesitant regarding sort solution is because if we're already doing the work of pulling in all the keys, why don't we just schedule them. i imagine we could just pull in all keys and hash out keys. this way we're not just doing a bunch of work and throwing away the majority of it. this would be a less complete solution to what we did in v4.

IIUC, the point in time which I have sorted the metrics in python which is at the daemon level (MetricsScheduler), thus there is no mechanism in place right now to only sort what you need for your block(at the coordinator level and across daemons/nodes with gnocchi metricd).  If there is another method/way I am open to exploring it, it just seems to me that this is a fairly low complexity cost here to avoid a very real issue with Gnocchi 3.0/3.1 causing an OSD to fail.  I think the trade off of throwing away part of a sorted list is acceptable compared to starving some metrics and eventually failing an osd and/or causing a group wsgi services to fail due to hosting in a singular httpd instance.

Though if you can help point me in a direction of throwing away less of the sorted keys I am all open.

chungg:
tangent: i should start off and say, i just noticed stable/3.0 does not have a memory tweak we have in stable/3.1 (https://github.com/gnocchixyz/gnocchi/blob/stable/3.1/gnocchi/storage/incoming/ceph.py#L143-L144) this resolves memory usage but not scheduling (aka growing backlog)

you're right. i don't think there's a way to sort only what you'll use since what you'll use is probably from the end of the iterable. because of that, i'm thinking it's better to just completely reconsider what it is 'metricd needs'.

specifically, as the only real solution is to retrieve the full list of objects, if we pull in the full list, we can just hash the result and based on what 'part' we're suppose to handle, we take only those objects. so in pseudo code:

```
def list_metric_with_measures_to_process(self, size, part, full=False):
    names = self._list_object_names_to_process()
    names_hashed = names_with_only_metric_id mod total_parts
    return set(whatever_names_in_part)
```

as i typed that i realised we'd need to change the method definiton to pass in total_parts... :(

gnocchi/storage/ceph.py
```diff
@@ -202,7 +202,9 @@ def _pending_measures_to_process_count(self, metric_id):
         return len(list(self._list_object_names_to_process(object_prefix)))
 
     def list_metric_with_measures_to_process(self, size, part, full=False):
-        names = self._list_object_names_to_process()
+        names = sorted(
```
jd: There's no need to sort if `full= True`, right?

gnocchi/storage/ceph.py
```diff
@@ -202,7 +202,9 @@ def _pending_measures_to_process_count(self, metric_id):
...
+            self._list_object_names_to_process(),
+            key=lambda x: '{}-{}'.format(x.split('_')[3], x.split('_')[4]))
```
jd: I'm not sure it's worth replacing the `_` by `-` between `[3]` and `[4]` so you can just do:

`lambda x: x.split("_", 3)[-1]`

jd:
Closing for #275 instead.

PR #267, opened 2017-07-31
rest: add a like test

sileht:


gnocchi/tests/functional/gabbits/search.yaml
```diff
@@ -64,6 +64,14 @@ tests:
       response_json_paths:
         $.`len`: 0
 
+    - name: search like id
+      POST: /v1/search/resource/generic
+      data:
+        eq:
+          id: 'faef212f%'
```
jd: not sure why it works? the operator is `eq`?

sileht:
Pull-request updated, HEAD is now 45fd16c42d567a46517873df98441b2a64a713ba

gnocchi/tests/functional/gabbits/search.yaml
sileht: The operator should be like.

sileht:
Pull-request updated, HEAD is now b0e3d5b113c2606cb9a9ddbb9221384278df692b

sileht:
Pull-request updated, HEAD is now 4c39e4bec0b85a39c3fa709102bedc716e2cf287

sileht:
I haven't dig, but yeah the test fail with created_by_project_id...

PR #269, opened 2017-08-03
don't blindly truncate on init

chungg:
we don't need to truncate for most scenarios.

- from_grouped_serie doesn't need to be truncated because it's
based on bound_series which definitely doesn't need to be truncated
- from_data is used to grab from storage but there is no point
truncating here as don't check max_size when doing so. (we may want to)
- from_timeseries needs to truncate as that's what used to return
results

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -547,7 +547,8 @@ class AggregatedTimeSerie(TimeSerie):
     COMPRESSED_SERIAL_LEN = struct.calcsize("<Hd")
     COMPRESSED_TIMESPAMP_LEN = struct.calcsize("<H")
 
-    def __init__(self, sampling, aggregation_method, ts=None, max_size=None):
+    def __init__(self, sampling, aggregation_method, ts=None, max_size=None,
+                 truncate=True):
         """A time serie that is downsampled.
 
         Used to represent the downsampled timeserie for a single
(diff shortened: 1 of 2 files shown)


gnocchi/carbonara.py
```diff
@@ -752,12 +758,11 @@ def serialize(self, start, compressed=True):
         offset = int((first - start.key) / offset_div) * self.PADDED_SERIAL_LEN
         return offset, payload
 
-    def _truncate(self, quick=False):
+    def _truncate(self):
         """Truncate the timeserie."""
         if self.max_size is not None:
             # Remove empty points if any that could be added by aggregation
-            self.ts = (self.ts[-self.max_size:] if quick
-                       else self.ts.dropna()[-self.max_size:])
```
jd: fun fact: this code is never executed. That's cool because it's wrong: `self.ts.dropna()` does not work anymore since `self.ts` is not a `pandas.Series` anymore.

That means currently `quick` is always `True`. Which is correct since `_truncate` is only called by `__init__` with `quick=True`.

chungg:
we actually do need to truncate because when we store measures we don't necessarily remove measures beyond policy. for example, in ceph we only ever write new measures. the older than policy measures are always there until the object gets purged. so when we read, we need to throw away older than policy measures possibly.

jd:
Fair enough. I'm not sure this is clearly covered in the tests though?

PR #272, opened 2017-08-04
Fix pep8 with file()

zaitcev:
Apparently our CI does not check for tox -e pep8.


zaitcev:
Clearly we do run pep8 checks in CI. Maybe my in-system pep8 overrides the one that tox would download? I have python-pep8-1.6.2-8.fc26 on the OS (Fedora).

jd:
One thing we do is depends on https://pypi.python.org/pypi/hacking which pulls old pep8 and other tools. That might be it?

PR #274, opened 2017-08-11
metricd: expose tasks_per_worker as a config option

jd:


jd:
Pull-request updated, HEAD is now 3d6c30dc5ec5c0a7b3cb266db7ea5c255e2cd794

jd:
Pull-request updated, HEAD is now 41c1906cdfa525ae3a8ae754a32ebc1543300b2a

gnocchi/storage/__init__.py
```diff
@@ -28,9 +28,14 @@
                default='file',
                help='Storage driver to use'),
     cfg.IntOpt('metric_processing_delay',
-               default=60,
+               default=30,
                help="How many seconds to wait between "
                "scheduling new metrics to process"),
+    cfg.IntOpt('tasks_per_worker',
+               default=256,
```
chungg: 256 seems high to be honest. this means worker1 will get 256 metrics in non-ceph drivers and if there are only 200 metrics in total, worker2+ will get nothing.

gnocchi/storage/__init__.py
```diff
@@ -28,9 +28,14 @@
...
+               min=1,
+               help="How many tasks to assign each metricd worker when "
+               "scheduling measures processing jobs"),
```
chungg: i think this is only really useful on ceph tbh. should probably add that to help?

chungg:
should we be applying this to stable/3.1 first?

gnocchi/storage/__init__.py
jd: You can fight with @akrzos on that but he tested a lot of values from 16 to 1024 and that's what worked best.

The problem is that anything under 256 metrics is barely enough to keeping a metricd busy for 30s, so when you're backlog is huge, metricd are idle 50% of the time. :(

gnocchi/storage/__init__.py
jd: Not really, it's useful for any driver that wants to tweak the throughput.

chungg: that's kinda my point. the throughput is some combination of: processing_delay, number of unqiue metrics, metrics posting interval, and number of workers.

that's the reason for my concern with setting default to 256, it only makes sense for specific scenario. we need to clarify what that scenario is because TASKS_PER_WORKER = 16 worked  fine for my scenario (1000metrics, every minute, 72 workers).
e for my scenario (1000metrics,

jd: 72 workers is *huge* for 1k metrics which is *small*.

This is exactly why we need to change to 256. You are fixing the CPU not being used at 100% per worker over a your interval by running A LOT of workers which consumes a ton of memory. It's way more memory friendly to run less workers actually doing a lot of their jobs.

Ultimately I don't mind leaving 16 as a default. We'll carry a different default in TripleO, that's all :/

chungg: sorry, i meant 1K resources and 20K metrics (still small)... but this is with ceph. the impact this has on the non-ceph drivers is different.

point is, i  think we need to define what the default is configured for, not that 256 (or any number) is wrong.

jd: Well the current default of 60s for `metric_processing_delay` and 16 `tasks_per_worker` *cannot* keep a worker busy for more than 3 seconds unless you're using an old Pentium 133 MHz.
So even we switch to a default of 5s for `metric_processing_delay`, which will heavily hit the the storage by listing over and over again the measures, or we keep the CPU busy for a decent amount of time (at least `metric_processing_delay` seconds) which means grabbing a whole bunch of measures to churn on, i.e. 256.

This is not perfect, and this algorithm cannot be perfect and it needs tuning based on your workload, anyway. But I can't see how the current default are ok on any decent computer. :(

chungg: again

> point is, i think we need to define what the default is configured for, not that 256 (or any number) is wrong.

so if the default is designed for 30s delay, x metrics, y metricd, etc... which should note it. or else the user has no idea what they should change for their system.

jd: I don't think you understand the problem completely.
The default (workers + delay + tasks_per_worker) are designed for processors < 200 MHz or a workload of `16 * workers` metrics. That's ridiculous.

We can document that, for sure. But it seems counter-productive to have defaults that are that shitty… so what about we fix them in the meantime? That's what this patch's trying to do. :)

chungg: i know the problem. it's why i always tell people to change TASKS_PER_WORKER :p 

what i'm suggesting is we document better as only people who know the scheduling works in v3.x can actually figure out that this value should be adjusted. something to inform user that TASKS_PER_WORKER * workers should equal the number of metrics (or measures for ceph case) accumulated over processing_delay.

jd:
Pull-request updated, HEAD is now ae852c22471371675d0ab289501d09342e465436

jd:
I've added documentation.

doc/source/configuration.rst
```diff
@@ -70,6 +70,25 @@ edit the `api-paste.ini` file to add the Keystone authentication middleware::
...
+workers using the coordination backend.
+
+Each polling is ran after `metricd_processing_delay` seconds and grabs
+`tasks_per_worker` metrics to process. Those two values are configurable.
+
+The shorter `metricd_processing_delay` is, the less lag you will have in your
+computed metrics, but the more the workers will hit the storage backend often.
+
+The value of `tasks_per_worker` is the number of metrics that will be grabbed
+for processing on each polling. That means it should be big enought to keep a
```
chungg: s/enought/enough/

doc/source/configuration.rst
```diff
@@ -70,6 +70,25 @@ edit the `api-paste.ini` file to add the Keystone authentication middleware::
...
 .. _`Grafana`: http://grafana.org/
 
+Configuring metricd
+-------------------
+
+The `gnocchi-metricd` daemon regularly polls the storage backend for new
+measures to process and processes them, spreading the workload across all its
+workers using the coordination backend.
+
+Each polling is ran after `metricd_processing_delay` seconds and grabs
```
chungg: s/ran/run/

doc/source/configuration.rst
```diff
@@ -70,6 +70,25 @@ edit the `api-paste.ini` file to add the Keystone authentication middleware::
...
+`tasks_per_worker` metrics to process. Those two values are configurable.
```
chungg: s/Those/These/

jd:
Pull-request updated, HEAD is now 86cd2e6867164f77610371f7a5e9a953eb947d6b

PR #275, opened 2017-08-11
storage: fix metric processing scheduling inefficiency

jd:
metricd scheduler learns about which metrics to process by calling
`list_metric_with_measures_to_process(size, part)`.

The problem currently is that list_metric_with_measures_to_process() uses
(size, part) to poke holes into a list of _measures_, which are very likely to
be ordered by the name of the measure file (true at least for file and ceph
drivers). That means that if the backlog is something like:

[ metric1_measure1, metric1_measure2, … metric2_measure1, metric2_measure2, …]

and if metric1 has e.g. 100k measures to be processed, then all schedulers will
poke holes starting from the beginning of the backlog:

 scheduler1 backlog:
  [ metric1_measure1, metric1_measure2, … metric1_measure256 ]

 scheduler2 backlog:
  [ metric1_measure257, metric1_measure258, … metric1_measure512 ]

which is then boiled down to:

  scheduler1 metric to process: metric1
  scheduler2 metric to process: metric1

this is completely inefficient as they might be tons of metrics to process but
all scheduler are focusing on the same first metric.

This patch tries to fix that by using the full measure listing, boiling down
the list of metric that needs to be processed, sorting it, and *then* poking
holes in that backlog (so the backlog is not about measures, but about metrics
this time). This is way more efficient as it would solve the case described
above.

The only downside is that the Swift driver, which was the only one not doing a
full measure listing on each call, has now to do a full listing too. But it
should be a trade-off worth it.

### DIFF
--- a/gnocchi/storage/ceph.py
+++ b/gnocchi/storage/ceph.py
@@ -201,13 +201,13 @@ class CephStorage(_carbonara.CarbonaraBasedStorage):
         object_prefix = self.MEASURE_PREFIX + "_" + str(metric_id)
         return len(list(self._list_object_names_to_process(object_prefix)))
 
-    def list_metric_with_measures_to_process(self, size, part, full=False):
-        names = self._list_object_names_to_process()
-        if full:
-            objs_it = names
-        else:
-            objs_it = itertools.islice(names, size * part, size * (part + 1))
-        return set([name.split("_")[1] for name in objs_it])
+    def _list_metric_with_measures_to_process(self):
+        # Sort measures objects per metric id
+        names = sorted(o.split("_")[1]
+                       for o in self._list_object_names_to_process())
+        # Group per metric id and store len() of the number of measures
+        return ((metric, len(list(measures)))
+                for metric, measures in itertools.groupby(names))
 
     def _delete_unprocessed_measures_for_metric_id(self, metric_id):
         object_prefix = self.MEASURE_PREFIX + "_" + str(metric_id)
--- a/gnocchi/storage/_carbonara.py
+++ b/gnocchi/storage/_carbonara.py
@@ -358,8 +358,20 @@ class CarbonaraBasedStorage(storage.StorageDriver):
     def _delete_metric(metric):
         raise NotImplementedError
 
+    def list_metric_with_measures_to_process(self, size, part, full=False):
+        metrics = map(operator.itemgetter(0),
+                      # Sort by the number of measures, bigger first (reverse)
+                      sorted(
+                          self._list_metric_with_measures_to_process(),
+                          key=operator.itemgetter(1),
+                          reverse=True))
+        if full:
+            return set(metrics)
+        return set(list(metrics)[size * part:size * (part + 1)])
+
     @staticmethod
-    def list_metric_with_measures_to_process(size, part, full=False):
+    def _list_metric_with_measures_to_process():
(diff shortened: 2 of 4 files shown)


jd:
Pull-request updated, HEAD is now 49e4b223d22f4c446b7ba6326253d7127d932ea9

gnocchi/storage/ceph.py
```diff
@@ -201,13 +200,9 @@ def _pending_measures_to_process_count(self, metric_id):
...
-    def list_metric_with_measures_to_process(self, size, part, full=False):
-        names = self._list_object_names_to_process()
-        if full:
-            objs_it = names
-        else:
-            objs_it = itertools.islice(names, size * part, size * (part + 1))
-        return set([name.split("_")[1] for name in objs_it])
+    def _list_metric_with_measures_to_process(self):
+        objs_it = self._list_object_names_to_process()
+        return sorted(set(name.split("_")[1] for name in objs_it))
```
akrzos: The sorting needs to occur based on datetime stamp or it just sorts based on metric uuid which will favor any uuids that "sort to the top by default"  this is especially harmful if those metric uuids exist for a very long time. (similar to my test with persisting instances and thus persisting metric uuids.)  

What about:
```python
return [name.split("_")[1] for name in sorted(objs_it, key=lambda x: x.split("_", 3)[-1])]
```
A further benefit of sorting by the incoming date timestamp is that the behavior becomes deterministic over the randomness of uuids.

jd: The best sorting would be done on the number of measures that a metric have actually. That'd batch the measures treatment so that'd speed things up a bit.
I'll try to change the patch to add that.

Even without any sorting this should fix the biggest issue of the current algorithm, fingers crossed.

jd:
Pull-request updated, HEAD is now 65734e11147a99847c9fd524c9045c842c398290

jd:
Pull-request updated, HEAD is now 7dbc5766ba6ec981379a5db6a92f79685ed33a37

jd:
Pull-request updated, HEAD is now 8fb8cb56f60007a68bb27784c06015b9520ab832

jd:
Pull-request updated, HEAD is now 0d24433b532ddd2295875749239447a275ac6db3

jd:
Pull-request updated, HEAD is now 6c1fae2ae46f39136ae696a327f4d1fa4234758d

jd:
@akrzos I can write a similar patch for Swift but that'll require listing all the backlog via the REST API, I don't know how heavy it'll be. I'll give it a try.

jd:
Pull-request updated, HEAD is now 5cae4915349181546c05ae3bb0e0cd2c48281703

jd:
Pull-request updated, HEAD is now f9adb5249fd64edd08d785ca40fc41230b3ffefc

jd:
Pull-request updated, HEAD is now f3befab3427c236939c96f5ae724d2b593cfd92a

jd:
Pull-request updated, HEAD is now a77a0b230d07f431a825051290f7628b655e721b

jd:
Pull-request updated, HEAD is now 1930dc9e8dd9147fec51322a9a8422c2e9d40db1

jd:
Pull-request updated, HEAD is now c3d10a557783ba09b8a458251509d2dea8eba2b3

PR #280, opened 2017-08-17
Add support for sack notification on incoming driver

jd:
incoming: add iter_on_sacks_to_process support for Redis
Add notification mechanism in incoming driver to be used by metricd worker
cli: add a wakeup() method to stop process waiting if needed
cli: rename _get_tasks to _get_sacks_to_process

### DIFF
--- a/gnocchi/cli/metricd.py
+++ b/gnocchi/cli/metricd.py
@@ -143,7 +143,7 @@ class MetricProcessor(MetricProcessBase):
                       'partitioning. Retrying: %s', e)
             raise tenacity.TryAgain(e)
 
-    def _get_tasks(self):
+    def _get_sacks_to_process(self):
         try:
             self.coord.run_watchers()
             if (not self._tasks or
@@ -161,7 +161,7 @@ class MetricProcessor(MetricProcessBase):
     def _run_job(self):
         m_count = 0
         s_count = 0
-        for s in self._get_tasks():
+        for s in self._get_sacks_to_process():
             # TODO(gordc): support delay release lock so we don't
             # process a sack right after another process
             lock = self.incoming.get_sack_lock(self.coord, s)
--- a/gnocchi/cli/metricd.py
+++ b/gnocchi/cli/metricd.py
@@ -24,6 +24,7 @@ import six
 import tenacity
 import tooz
 
+from gnocchi import exceptions
 from gnocchi import incoming
 from gnocchi import indexer
 from gnocchi import service
@@ -121,6 +122,10 @@ class MetricProcessor(MetricProcessBase):
             worker_id, conf, conf.metricd.metric_processing_delay)
         self._tasks = []
         self.group_state = None
+        self.sacks_with_measures_to_process = set()
+        # This stores the last time the processor did a scan on all the sack it
+        # is responsible for
+        self._last_full_sack_scan = utils.StopWatch().start()
 
     @tenacity.retry(
         wait=_wait_exponential,
@@ -149,6 +154,29 @@ class MetricProcessor(MetricProcessBase):
                       'partitioning. Retrying: %s', e)
             raise tenacity.TryAgain(e)
 
+        filler = threading.Thread(target=self._fill_sacks_to_process)
+        filler.daemon = True
+        filler.start()
+
+    @retry_on_exception.wraps
+    def _fill_sacks_to_process(self):
+        try:
+            for sack in self.incoming.iter_on_sacks_to_process():
+                if sack in self._get_sacks_to_process():
+                    LOG.debug(
+                        "Got notification for sack %s, waking up processing",
+                        sack)
+                    self.sacks_with_measures_to_process.add(sack)
+                    self.wakeup()
+        except exceptions.NotImplementedError:
+            LOG.info("Incoming driver does not support notification")
+        except Exception as e:
+            LOG.error(
+                "Error while listening for new measures notification, "
+                "retrying",
+                exc_info=True)
+            raise tenacity.TryAgain(e)
+
     def _get_sacks_to_process(self):
         try:
             self.coord.run_watchers()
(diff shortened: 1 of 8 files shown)


jd:
Pull-request updated, HEAD is now 1440443c2374c02216e950d067876f1781412d0e

jd:
Pull-request updated, HEAD is now 45ce8485f11f4f8eeed0e25f8a2eaa4a76f5b585

jd:
Pull-request updated, HEAD is now 927426d9c690f3ef5cbb8782ad39190b3e90ed14

gnocchi/incoming/redis.py
```diff
@@ -108,3 +108,15 @@ def process_measure_for_metric(self, metric):
...
         self._client.ltrim(key, item_len + 1, -1)
+
+    def iter_on_sacks_to_process(self):
+        self._client.config_set("notify-keyspace-events", "El")
+        p = self._client.pubsub()
+        db = self._client.connection_pool.connection_kwargs['db']
+        channel = b"__keyevent@" + str(db).encode() + b"__:rpush"
+        p.subscribe(channel)
+        for message in p.listen():
+            if message['type'] == 'message' and message['channel'] == channel:
```
chungg: we're only subscribed to a single channel so we should only get that channel no?

gnocchi/incoming/redis.py
```diff
@@ -108,3 +108,15 @@ def process_measure_for_metric(self, metric):
 
         # ltrim is inclusive, bump 1 to remove up to and including nth item
         self._client.ltrim(key, item_len + 1, -1)
+
+    def iter_on_sacks_to_process(self):
+        self._client.config_set("notify-keyspace-events", "El")
+        p = self._client.pubsub()
+        db = self._client.connection_pool.connection_kwargs['db']
+        channel = b"__keyevent@" + str(db).encode() + b"__:rpush"
```
chungg: the above should be part of init? we shouldn't need to do it constantly or does the p.listen() never exit?

gnocchi/tests/test_incoming.py
```diff
@@ -0,0 +1,65 @@
...
+import uuid
+
+import numpy
+
+from gnocchi import incoming
+from gnocchi import storage
+from gnocchi.tests import base as tests_base
+
+
+class TestIncomingDriver(tests_base.TestCase):
```
chungg: TestIncomingNotifierDriver? something less generic... same with filename.

gnocchi/cli.py
```diff
@@ -231,6 +236,30 @@ def _configure(self):
...
+            for sack in self.incoming.iter_on_sacks_to_process():
+                if sack in self._get_sacks_to_process():
+                    LOG.debug(
+                        "Got notification for sack %s, waking up processing",
+                        sack)
+                    self.sacks_with_measures_to_process.add(sack)
+                    self.wakeup()
+        except exceptions.NotImplementedError:
+            LOG.info("Incoming driver does not support notification")
+            return
```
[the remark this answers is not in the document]
jd: Fixed.

gnocchi/incoming/redis.py
jd: If any exception occurs, you need to reconnect and resubscribe. The reconnection is handled by the library but then it needs to recreate a publisher.

If everything goes fine, this is executed only once. Otherwise, it re-sub.

gnocchi/incoming/redis.py
jd: Yes though you never know if e.g. Redis starts publishing other stuff by default or whatever. Better be safe then sorry :)

gnocchi/tests/test_incoming.py
jd: The incoming tests we have right now are in mixed with storage in `test_storage` for historical reason. This is the first test that is incoming specific. That's why it's called `test_incoming`.

My point: is this file and class should already exist and I'd be just adding `test_iter_on_sacks_to_process` to that. But it does not (for historical reason) so I'm just fixing it here. :)

jd:
Pull-request updated, HEAD is now 2b010a1db40424208ea0930ea570947fb5bdb4d7

jd:
Pull-request updated, HEAD is now acff856c822bdf8acf153b47a9ebbee6442b4d5b

jd:
@gnocchixyz/devs please weight in :)

chungg:
i'm still curious why there's all the encoding changes

gnocchi/cli/metricd.py
```diff
@@ -178,7 +221,13 @@ def _run_job(self):
                           exc_info=True)
             finally:
                 lock.release()
+            # Discard the sack from the sacks since we just did it.
+            self.sacks_with_measures_to_process.discard(s)
```
chungg: should this be cleared before we process it? or else it could remove a newly added entry between us processing and getting to this line.

gnocchi/cli/metricd.py
```diff
@@ -63,11 +68,13 @@ def run(self):
         while not self._shutdown.is_set():
             with utils.StopWatch() as timer:
                 self._run_job()
-            self._shutdown.wait(max(0, self.interval_delay - timer.elapsed()))
+            self._wake_up.wait(max(0, self.interval_delay - timer.elapsed()))
+            self._wake_up.clear()
```
chungg: same thing.

jd:
The encoding changes are made because Redis sends bytes on the wire and wants bytes for the channel – which makes sense, Redis does not care about encoding.

So I've switched the SEPARATOR to bytes, which makes it usable as bytes on the wire or for the channel, but not for the rest of the pyredis stuff, which uses string and not bytes unfortunately.

gnocchi/cli/metricd.py
jd: same thing as?

gnocchi/cli/metricd.py
jd: Yes, you're right, but, if we more it before, and an exception occurs in the processing, then, it'll be removed and it won't be processed again. So here I've made the trade-off of saying, well, you can be processing measures files 1, 2, 3, and while you're doing it, there's 4 coming, but you missed it, well, it'll be processed next interval _anyway_.

gnocchi/cli/metricd.py
chungg: as the "clear before we process it" comment.

chungg:
if we change keys to bytes... we a need a migration or we lose anything not encoding in bytes?

jd:
> if we change keys to bytes... we a need a migration or we lose anything not encoding in bytes?

oh no, there's no change in that regard. the keys already are bytes internally in Redis, that's just that we now manipulate bytes on our side, encode into strings for pyredis, which decodes them into bytes lol

https://github.com/andymccurdy/redis-py/blob/master/redis/connection.py#L97

gnocchi/cli/metricd.py
jd: not sure I follow here… `_wake_up` is a `threading.Event` object :)

gnocchi/cli/metricd.py
chungg: can we add it back if there's an exception? or we worried that it's bad to blindly add something back?

chungg:
> oh no, there's no change in that regard. the keys already are bytes internally in Redis, that's just that we now manipulate bytes on our side, encode into strings for pyredis, which decodes them into bytes lol

ah, cool cool

gnocchi/cli/metricd.py
jd: I think we can add it back if we make sure it was there in the first place, of course. Modified the patch that way.

jd:
Pull-request updated, HEAD is now f7c20252df91add81a16b3bcab034d713972a544

gnocchi/cli/metricd.py
chungg: ok nevermind. i was thinking wrong.

Issue #61, opened 2017-06-01
Reduce memory footprint for gnocchi-{janitor,reporter} processes


chungg:
this might unavoidable but when running memory_profiler

having the line "from gnocchi import service" increases memory ~5x from ~25MB to ~115MB. this seems weird as hell considering it's just initialising conf file.

jd:
This probably imports carbonara which import pandas which imports everything (numpy etc).

sileht:
I confirm that: 

```
etox -epy27-mysql-file mprof run python p.py
mprof: Sampling memory every 0.1s
Filename: p.py

Line #    Mem usage    Increment   Line Contents
================================================
     6     12.7 MiB      0.0 MiB   @profile
     7                             def load():
     8     13.0 MiB      0.2 MiB       import lz4.block
     9     24.4 MiB     11.5 MiB       import numpy
    10     24.4 MiB      0.0 MiB       import numpy.lib.recfunctions
    11     83.2 MiB     58.7 MiB       import pandas
    12     95.3 MiB     12.1 MiB       from scipy import ndimage
    13     95.3 MiB      0.0 MiB       import six
```

jd:
Right, so nothing we're not aware of or we really want to fix I guess. I'm closing @chungg but feel free to reopen if you really want to fix it.

chungg:
hmm. i'd like to minimise memory... i think it's possible to make janitor and reporter workers to be less but we'll have way more processing workers which i'm not sure we can minimise if pandas is the main reason. i'll leave it closed unless you folks have ideas.

jd:
I'll reopen because I think it's possible and it's a good idea. I did not think about the impact on reporter/janitor.

We might need to split service.py and write some tests to make we don't break that again later. It's easy to be greedy on import.

chungg:
just looking at this more. the switch to numpy has been awesome. the remaining pandas usage i see are for 2 things, either: for processing timespan input or for computing cross metric computations. both of these are used solely for API so it seems we could avoid pulling in pandas for metricd if we correctly separate the pandas code to API.

jd:
Yes, and both those usage should be removed. I'll probably need @sileht help to remove the cross metric computing (since he wrote it), but the time parsing should be not too hard to extract from pandas.

chungg:
i'm wondering if we should add a new API service that does analysis or something. there's stuff like moving/rolling average/max/min/etc... which is useful but may need pandas (or maybe not)

[...]

PR #284, opened 2017-08-21
Use pytimeparse to parse time span

jd:
pytimeparse was used before switching entirely the parsing to Pandas, which was
already in use. Since we're trying to get rid of Pandas, pytimeparse looks like
the best alternative out there to parse the time span.

Improves #61

### DIFF
--- a/gnocchi/utils.py
+++ b/gnocchi/utils.py
@@ -26,7 +26,7 @@ import daiquiri
 import iso8601
 import monotonic
 import numpy
-import pandas as pd
+import pytimeparse
 import six
 from stevedore import driver
 from tooz import coordination
@@ -103,7 +103,7 @@ def to_timestamps(values):
                     # e.g. "-10 seconds" or "5 minutes"
                     times = numpy.fromiter(
                         numpy.add(numpy.datetime64(utcnow()),
-                                  pd.to_timedelta(values, box=False)),
+                                  [to_timespan(v, True) for v in values]),
                         dtype='datetime64[ns]')
                 else:
                     times = numpy.array(values, dtype='datetime64[ns]')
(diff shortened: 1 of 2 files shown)


gnocchi/utils.py
```diff
@@ -133,19 +133,17 @@ def timestamp_to_datetime(v):
         v.astype(float) / 10e8).replace(tzinfo=iso8601.iso8601.UTC)
 
 
-def to_timespan(value):
+def to_timespan(value, allow_le_zero=False):
```
chungg: what's allow_le_zero for?

jd: allow lesser or equal than zero. It's used on L106.

chungg: right, i'm just wondering where we actually allow that currently? becauses the previous code does not support this condition

jd: Oh it does support it, see line 106 again. When converting "-5 seconds" to a *datetime64*, what's computed is `now() - 5 seconds`. So we need to allow for those negative timespan.

The current code used `pandas.to_timedelta` in `to_timestamps` where this new one use the new `to_timespan` with `allow_le_zero` set to `True`.

Welcome back ;)

chungg: oh word.

chungg:
@sileht you're right. it is slower. about 20-60% slower depending on size. :(  not sure how to get around that.

```
>>> values = ["-%s seconds" % i for i in xrange(10)]
>>> valuesbig = ["-%s seconds" % i for i in xrange(100)]
>>> valuesbigger = ["-%s seconds" % i for i in xrange(1000)]

>>> timeit.timeit("[to_timespan(v, True) for v in values]", "import pytimeparse; from __main__ import to_timespan, values, valuesbig, valuesbigger", number=1000)
0.21312499046325684
>>> timeit.timeit("pd.to_timedelta(values, box=False)", "import pandas as pd; from __main__ import values, valuesbig, valuesbigger", number=1000)
0.17818307876586914

>>> timeit.timeit("[to_timespan(v, True) for v in valuesbig]", "import pytimeparse; from __main__ import to_timespan, values, valuesbig, valuesbigger", number=1000)
1.2020509243011475
>>> timeit.timeit("pd.to_timedelta(valuesbig, box=False)", "import pandas as pd; from __main__ import values, valuesbig, valuesbigger", number=1000)
0.8025670051574707

>>> timeit.timeit("[to_timespan(v, True) for v in valuesbigger]", "import pytimeparse; from __main__ import to_timespan, values, valuesbig, valuesbigger", number=1000)
11.29253888130188
>>> timeit.timeit("pd.to_timedelta(valuesbigger, box=False)", "import pandas as pd; from __main__ import values, valuesbig, valuesbigger", number=1000)
7.07202410697937
```

jd:
You can send pull requests to pytimeparse, I already did a while back, if you know how to improve it.
It uses regexp whereas Pandas might use something written in Cython.

chungg:
yeah, cython. https://github.com/pandas-dev/pandas/blob/473a7f3c186f6b0bfd9d3ce413fb627cf7a8f111/pandas/_libs/tslib.pyx#L3369

it'd be cool if we could split pandas so it wasn't all or nothing.

chungg:
whelp my attempt at a single regex comparison in pytimeparse made it slower. oh well

PR #288, opened 2017-08-21
indexer: fix ne operator

### LINKED ISSUE #224: resource query doesn't work in some circumstances

jd:
When we use the ne operator, row with NULL values are not returned by
mysql or postgresql.

This change fixes the sql query to return them.

Closes #224

(cherry picked from commit 051b11abb41b239a1873e4009d0a6625795f1f76)


jd:
Pull-request updated, HEAD is now f92f3264c288d752c165cfd3db3e9087141e479e

PR #290, opened 2017-08-21
rest: Fix invalid query validation

### LINKED ISSUE #240: resource query sometimes raise 500 instead of 400 on invalid queries
I found some queries that returns 500 instead of 400:

```
ne:
  project_id: ["foobar"]
```

or

`not: [{"foobar": "barfoo"}]`

jd:
Some operators accept a list or a dict at rest layer, when it should
accept only an attribute value, like bool, str, None.

For Postgresql, that result to an sql error, turned in HTTP 500.
For Mysql, the request succeed with some magic casting.

Closes #240

(cherry picked from commit cd093e4641a202c720ec8b54e335cc91a07f97ca)

### DIFF
--- a/gnocchi/rest/__init__.py
+++ b/gnocchi/rest/__init__.py
@@ -1140,6 +1140,12 @@ def ResourceSearchSchema(v):
     return _ResourceSearchSchema()(v)
 
 
+# NOTE(sileht): indexer will cast this type to the real attribute
+# type, here we just want to be sure this is not a dict or a list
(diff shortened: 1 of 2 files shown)


jd:
Pull-request updated, HEAD is now 07f07e68f7b3802319c0b2e73b620689a1906ad1

jd:
Pull-request updated, HEAD is now 49e61e42d5828ef458c06e6668ca58ef147190c0

PR #297, opened 2017-08-26
Add #!/usr/bin/python to gnocchi-api

zaitcev:
Otherwise
Aug 26 01:04:04 rhev-a24c-01 systemd: openstack-gnocchi-api.service:
 Failed at step EXEC spawning /usr/bin/gnocchi-api: Exec format error

The unit file is like

[Service]
ExecStart=/usr/bin/gnocchi-api -- --log-file /var/log/gnocchi/api.log

### DIFF
--- a/gnocchi/rest/gnocchi-api
+++ b/gnocchi/rest/gnocchi-api
@@ -1,3 +1,5 @@
+#!/usr/bin/python
+
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at


zaitcev:
The pbr does that with the generated scripts, such as /usr/bin/gnocchi-metricd. If you are fully compatible with py3, we can just make it /usr/bin/python without qualifiers.

jd:
Yeah it works fully on Python 3. I'd say go with `/usr/bin/python` even if I imagine it's not the best case, as it might be an issue with virtualenv. Not sure.

PR #298, opened 2017-08-28
Pull request for stable-3.1-task-per-worker

jd:
Change default processing delay and tasks per worker
metricd: expose tasks_per_worker as a config option


jd:
Pull-request updated, HEAD is now a81c8e3181cc4ca1e23a2dd5aa4554df611f7f90

PR #300, opened 2017-08-29
Add Gordon Chung to .mailmap

jd:
This dude uses two different names in the git log, confusing git-log and other
tools as there was two different authors.


chungg:
lol, i'm committing from 3 different machines.

jd:
You mean, you both are? :p

PR #304, opened 2017-08-29
Added support for max_pool_connections on s3 driver

leandroreox:
Signed-off-by: Leandro Reox <leitan@MacBook-Pro-de-Leandro.local>


gnocchi/storage/s3.py
```diff
@@ -50,6 +50,12 @@
                  help="Maximum time to wait checking data consistency when "
                  "writing to S3. Set to 0 to disable data consistency "
                  "validation."),
+    cfg.FloatOpt('s3_max_pool_connections',
```
chungg: should probably be IntOpt

gnocchi/storage/s3.py
```diff
@@ -50,6 +50,12 @@
...
+                 min=10,
+                 default=50,
+                 help="The maximum number of connections to keep in a "
+                 "connection pool. If this value is not set, the default "
+                 "value of 10 is used."),
```
chungg: i'm not sure the last sentence is correct. can you unset the value if there is a default?

leandroreox: no, ill change that, i extracted it from the boto doc

gnocchi/storage/s3.py
leandroreox: yes, changing that

chungg:
could you squash all the fixes into a single commit? something like this: https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/

and then when you push, use git push -f origin master

leandroreox:
Merged al changes into a single squashed commit

gnocchi/storage/s3.py
```diff
@@ -50,6 +50,11 @@
                  help="Maximum time to wait checking data consistency when "
                  "writing to S3. Set to 0 to disable data consistency "
                  "validation."),
+    cfg.IntOpt('s3_max_pool_connections',
+                min=10,
```
jd: I this really 10? Seems to me that 1 should work.

leandroreox: @jd  10 is de default value that comes on the adapter.py of the python requests library, i really think lowering it is not a good call: https://github.com/requests/requests/blob/master/requests/adapters.py#L47

jd: It's not saying it's a good idea. It's saying it's possible. That's what this min/max values need to represent.

The _default_ value is what should be a good idea. :)

leandroreox:
Changed the minimun value since what @jd said makes a lot of sense

jd:
You still need to fix the tests it seems.

leandroreox:
@jd squashed :)

requirements.txt
```diff
@@ -23,3 +23,4 @@ Paste
 PasteDeploy
 monotonic
 daiquiri
+pyparsing>=2.2.0
```
jd: uh?

jd: You probably need to rebase on master

jd:
@leandroreox I suggest using https://github.com/jd/git-pull-request :)

leandroreox:
@jd merged from upstream

jd:
@leandroreox any update/fix?

leandroreox:
@jd what im seeing is that the test are failing with "boto3 unavailable" thats seems to be triggered when boto3 cannot be imported, boto3 is there since day one, i didnt touched that part, am i missing something here ? or this tests should not be passing like ever ?

leandroreox:
@jd fixed

chungg:
you should squash your commits :)

gnocchi/common/s3.py
```diff
@@ -41,7 +42,9 @@ def get_connection(conf):
         endpoint_url=conf.s3_endpoint_url,
         region_name=conf.s3_region_name,
         aws_access_key_id=conf.s3_access_key_id,
-        aws_secret_access_key=conf.s3_secret_access_key)
+        aws_secret_access_key=conf.s3_secret_access_key,
+        config = boto_config.Config(max_pool_connections=conf.
+                 s3_max_pool_connections))
```
chungg:
```
        config = boto_config.Config(
            max_pool_connections=conf.s3_max_pool_connections))
```

leandroreox:
Yes @chungg  i have done it kind that way, but commit and squash was pending, this should pass al the tests

jd:
@leandroreox looks good but you need to squash those commits

jd:
@leandroreox do you need help to rebase/squash?

leandroreox:
@jd yes, now i have merged with master i have a lot of commits in between mine, im looking how to rebase without screwing the other commits

chungg:
oh, you pushed your change to your master :/ ... i don't know what the correct way to fix that is. usually it's best (i think) to create:

1) sync your local master with upstream master
2) create a new branch
3) push that branch

maybe just reset your fork and push a new PR? the others know git better than me

jd:
@leandroreox You want to do `git rebase -i master` on your branch and put everything except the "main" commit on `squash` so they're all squashed. Then write a proper commit message and should should only have one commit. Resend your branch (use `git pull-request` if you're lazy) and you should be done.

leandroreox:
@jd this also applies if my branch is master ?

jd:
@leandroreox yeah that might not work, so you need to precise the master branch from gnocchi, which is very likely `origin/master`, so do `git rebase --interactive origin/master`

leandroreox:
Sorry about the whole rebase stuff, still learning how not to get fuc*** by git, is some action from my side still neede ? i saw that @jd commited the PR. Thanks

jd:
@leandroreox don't worry, I took care, you're good :)

leandroreox:
@jd haha thanks, promise to do the next better

PR #308, opened 2017-08-31
don't pass empty job list

chungg:
we should check if work is actually required rather than passing an
empty list. it will end up calling indexer for no reason.


chungg:
i should probably mention, i could also just add check here: https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/storage/_carbonara.py#L372

chungg:
word, i didn't look far enough.

PR #321, opened 2017-09-06
Move cross aggregation computing to gnocchi.rest

### LINKED ISSUE #61: Reduce memory footprint for gnocchi-{janitor,reporter} processes

jd:
This moves the cross aggregation of timeseries to a new dedicated
module gnocchi.rest.cross_metric. It turns out that this is one of the
latest module depending on Pandas (for probably good reasons).

However it is not directly related to Carbonara, so there's no need to
move it there. It could be put into gnocchi.storage, but it actually
have nothing to do with storage: it's only recomputing stuff once they
are retrieved.

Moving this to the REST area makes sure that it is only loaded by the
REST API server, and therefore prevent other processes such as metricd
to load Pandas (which is huge in memory).

Closes #61


jd:
Pull-request updated, HEAD is now 2afd8a4641396d9d05d88efa36f11b2a6dd4887b

chungg:
i like this (i was too lazy to do this).

to play devil's advocate. is this where we want the code or do we want another location? i ask because as you mentioned, the aggregated (and custom aggregations ie rolling/moving average) are the only 
ones that require pandas and i see them secondary aggregates for analysis. do we want to move them to an "analysis"/"BI" api so that gnocchi API does just basic CRUD functions and a second endpoint which extends/leverages the data gnocchi stores?

jd:
@chungg yup good point. so we can't change the current API _anyway_ so it's a good first step. If we want to do a more "complicated" API at some point, maybe create a new endpoint later?

I don't have the feeling that the current "analysis" part of the API is so big. At the point we start calling it "BI", indeed, we probably need something different ;)

PR #322, opened 2017-09-06
swift: rename and use swift_preauthurl

jd:
This option has been introduced on the first commit of Gnocchi… and never ever
used! Haha.


sileht:
It's actually a bug, it should be passed to swiftlclient.Connection

We already pass preauthtoken, preauthtoken and preauthurl works together.

In another side, in context of a server application. preauth* may be useless. Perhaps we should also drop preauthtoken instead.

jd:
Pull-request updated, HEAD is now e3d4d7213c5d6b0ff36cbacfc8a52cbbfe1d5797

PR #325, opened 2017-09-06
Move REST API code in its own submodule

jd:
The whole code is in gnocchi.rest, making it imported as soon as anything in
gnocchi.rest is imported. This might be needed when just importing e.g.
gnocchi.rest.app to get access to the options.


BigChief45:
Nice. I would like to add, how about also separating different endpoint related code into different files? For example, something like `rest/metrics.py`, `rest/measures.py`, etc. etc. Maybe separated according to the API's controllers?

jd:
@BigChief45 it's more likely to add a lot of churn for little gain

TBH I'm not particularly happy with Pecan, it's a real pain to find anything in the API.

jd:
Pull-request updated, HEAD is now 2d1c13f357580d9b071fe2fd5fc57e6a56b77c14

PR #328, opened 2017-09-07
rest: don't allow empty in

sileht:
Currently we can write {"id": []}. This query doesn't make sense it will
always return an empty list and it's obviously a input error.

This change returns 400 is this case.

### DIFF
--- a/gnocchi/indexer/sqlalchemy.py
+++ b/gnocchi/indexer/sqlalchemy.py
@@ -1104,6 +1104,7 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
 
 
 class QueryTransformer(object):
+
     unary_operators = {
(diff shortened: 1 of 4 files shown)


sileht:
Pull-request updated, HEAD is now 20d60275e46e1f151877f639b4511ddaa74b9e07

sileht:
Pull-request updated, HEAD is now fe9119789b2968c4d18aa58faa6bdc81c226948f

gnocchi/indexer/sqlalchemy.py
```diff
@@ -1104,6 +1104,7 @@ def _build_sort_keys(sorts):
 
 
 class QueryTransformer(object):
+
```
jd: O_O

PR #329, opened 2017-09-08
rest: fix ACL enforcement for non-admin on metric list

jd:
The current metric listing on /v1/metric does not enforce correctly the
attribute filtering on the creator field when the user is non-admin.

That means the attr_filter is empty and a non-admin can actually list all
metrics for all users.

This fixes that by setting the filter correctly.


jd:
Pull-request updated, HEAD is now 10a30b595514b7a2a08ddf28e9546515cf17fc35

PR #335, opened 2017-09-11
rest: implements pagination links for /metric

sileht:
This change implements pagination link for /metric according the
RFC5988.

It only implements the 'next' link, since other need extra mysql query
that we don't really want for now.

Related #3


sileht:
Pull-request updated, HEAD is now 9f7382a8162697f3cd36e43b1035742ce85c9e1e

sileht:
Pull-request updated, HEAD is now 8444633cce7d7b465bd84afcb8484ac9a9b81119

sileht:
Pull-request updated, HEAD is now b829d785570b225169c5dc0b92742f0fce330893

sileht:
Pull-request updated, HEAD is now c5fd7ec603dfee85b3d940cbcda2c9354791644a

sileht:
Pull-request updated, HEAD is now 608bc34d8e935c0ee54dde991c14935feda3ad07

sileht:
Pull-request updated, HEAD is now d1d97f20c325a722d351c4f95ec59a22c1e95bef

sileht:
Pull-request updated, HEAD is now c0506b21bc412d22bb22b09db8ea80a75a3eefbb

sileht:
Pull-request updated, HEAD is now 17bf737414fa027e76be633f361e3917e61eb920

sileht:
Pull-request updated, HEAD is now 3b6407ffaa40a58742a489b3f947c0bd8f9ef9ff

gnocchi/tests/functional/gabbits/resource.yaml
```diff
@@ -216,6 +216,15 @@ tests:
         $[1].revision_end: null
         $[1].metrics.'disk.iops': $RESPONSE["metrics.'disk.iops'"]
 
+    - name: get generic history with links
+      desc: Ensure we can get the history
+      xfail: true
```
jd: Why?

sileht: I have added the oslo.db bug ref for each xfail

sileht:
Pull-request updated, HEAD is now 2372926cf248237eac1dfe3163b21c39ecd9ccb4

sileht:
Pull-request updated, HEAD is now 3a94f38fedfc432f3135b8d20a852fe7834b7e74

sileht:
Pull-request updated, HEAD is now e9f843731495d276ddc3dbbf950da7cc7ac05e29

sileht:
Pull-request updated, HEAD is now 20c30c8f2f50ac71891b2d68d9acea4058ca4de8

chungg:
can we add docs? its not clear what 'marker=id@1' and 'marker=id@2'... is it saying '@1' is the first x number of of entries after id and '@2' is the second x number of entries after id (where x is the limit)?

jd:
This does not need to be documented. It's internal soup. The `Links` header are present and will be presented in the doc rendering since they are returned by the REST API. It's transparent for users.

chungg:
err... so does '@1' and '@2' mean what i typed? because as a developer, i'm not sure. :p

jd:
IIUC it's `resource_id@revision`

chungg:
what does revision mean in this case? that's the value i'm unsure what it means and the rfc 5988 doc doesn't mention anything either.

jd:
> what does revision mean in this case? that's the value i'm unsure what it means and the rfc 5988 doc doesn't mention anything either.

I think you miss the point. The RFC says that you can follow the links that are passed into headers. How they are built is an application problem and are **opaque** to the user and can change at anytime.

PR #338, opened 2017-09-12
add note regarding basic auth encoding

### LINKED ISSUE #316: need help on providing authorization header in request
when i try to create metric through v1/metric,  i am getting 401 unauthorized error.  As document said , i need to provide admin username in authorization header for basic authentication . Could you please some one tell me the format for basic authentication request

chungg:
basic auth requires Base64 encoding

Closes #316

### DIFF
--- a/doc/source/rest.j2
+++ b/doc/source/rest.j2
@@ -5,10 +5,12 @@
 Authentication
 ==============
 
-By default, the authentication is configured to the "basic" mode. You need to
+By default, the authentication is configured to the `"basic" mode`_. You need to
 provide an `Authorization` header in your HTTP requests with a valid username
 (the password is not used). The "admin" username is granted all privileges,
 whereas any other username is recognize as having standard permissions.
+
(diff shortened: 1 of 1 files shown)


doc/source/rest.j2
```diff
@@ -10,6 +10,13 @@ provide an `Authorization` header in your HTTP requests with a valid username
 (the password is not used). The "admin" username is granted all privileges,
 whereas any other username is recognize as having standard permissions.
 
+.. note::
+
+   The username should be encoded in Base64 when using `basic` mode as per
```
jd: If you want to paraphrase the standard, I don't think this is enough. It's `username:password` that must be encoded, except the password is not used (yet), so it'd be just `admin:` that needs to be encoded. :)

I'd prefer we make just it clear that the `basic` authentication mechanism of Gnocchi is actually a `Basic` authentication is the sense of HTTP.

chungg:
sure

Issue #186, opened 2017-07-05
support rolling computations


chungg:
rolling average and other rolling computations are commonly used by data 'scientist' and 'quants'... we should probably provide them with this functionality.

we support moving average but it's some custom python code which is arguably not correct (and now deprecated). we should just use pandas.

in case someone wants to work on this before i start.

PR #339, opened 2017-09-12
supporting rolling aggregates

chungg:
this implements rolling aggregates which works similar to how
pandas rolling works but more trivial. this differs from deprecated
moving_average by correctly matching points to trailing timestamp
rather than leading timestamp.

it does not return the NaN values that result in shift in values.

#186

### DIFF
--- a/doc/source/rest.j2
+++ b/doc/source/rest.j2
@@ -139,7 +139,8 @@ Or we can apply absolute to the values and then negate them for example.
 
 {{ scenarios['get-measures-transform']['doc'] }}
 
-Supported transformations are `absolute`, `negative` and `resample(sampling-in-second)`.
+Supported transformations are `absolute`, `negative`,
+`rolling(aggregation-method, window-size)` and `resample(sampling-in-second)`.
 
 .. note::
 
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -28,6 +28,7 @@ import time
 import lz4.block
 import numpy
 import numpy.lib.recfunctions
+from numpy.lib.stride_tricks import as_strided
 from scipy import ndimage
 import six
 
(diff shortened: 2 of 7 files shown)


gnocchi/aggregates/rolling.py
```diff
@@ -0,0 +1,49 @@
...
+    # http://arogozhnikov.github.io/2015/09/30/NumpyTipsAndTricks2.html
+    stride = values.strides[0]
+    return as_strided(values, shape=[len(values) - window + 1, window],
+                      strides=[stride, stride])
+
+
+def build_array(storage, metric, start, stop, granularity):
+    serie = storage.get_measures(metric, from_timestamp=start,
+                                 to_timestamp=stop, granularity=granularity)
+    ts = np.array(serie, dtype=[('timestamps', '<datetime64[ns]'),
```
jd: I'm feeling we could change `fetch` in Carbonara to directly return that format.

gnocchi/aggregates/rolling.py
```diff
@@ -0,0 +1,49 @@
...
+                                ('granularity', '<timedelta64[ns]'),
+                                ('values', '<d')])
+    return ts
+
+
+class RollingMean(aggregates.CustomAggregator):
+    def compute(self, storage_obj, metric, start, stop, granularity, window):
```
jd: seems `@staticmethod`

chungg:
not sure what you mean by old api?

this is a re-aggregation so it is slightly misleading as i've currently implemented it. i could switch this over to aggregation endpoint but that's generally used for multiple metrics...

jd:
By old API I meant the fact this it is passed as an `aggregation` parameter. If that fits well with that computing, I'm good with it. Just questioning it since I don't think anyone ever used the moving average one.

chungg:
i don't think they did either -- it doesn't give correct values typical of sliding/moving/rolling windows computations.

i'll try moving this to reaggregation since that is what it is.

PR #353, opened 2017-09-14
rest: fix ACL enforcement for non-admin on metric list

jd:
The current metric listing on /v1/metric does not enforce correctly the
attribute filtering on the creator field when the user is non-admin.

That means the attr_filter is empty and a non-admin can actually list all
metrics for all users.

This fixes that by setting the filter correctly.

(cherry picked from commit 2be95704a18229e4d05e57dd3e96de489a16e61d)


jd:
Pull-request updated, HEAD is now 4cc5745c87f8c94c95c4d39aa4221008ba14f98d

PR #355, opened 2017-09-14
travis: only build tags, not branches

jd:


sileht:
That doesn't work and that have broken pastamaker :)

Travis does not build this PR

jd:
Damn.

PR #358, opened 2017-09-15
cli: retry less aggressively if metricd can't connect to storage

jd:
Right now if the indexer or storage is not reachable/functioning, the process
crashes and cotyledon restarts it without any delay, which causes hammering the
e.g. storage.

This leverages the retry.utils decorator to retry in a less aggressive manner
to reconnect to storage.


jd:
Not needed actually.

PR #360, opened 2017-09-15
catch transformation errors

chungg:
- add a transformerror exception in case something goes wrong
during transform
- remove the valueerror for unknown transform method. the api
already validates this.


gnocchi/carbonara.py
```diff
@@ -569,9 +577,6 @@ def transform(self, transform):
                 ts = ts.resample(sampling)
                 timestamps = ts["timestamps"]
                 values = ts["values"]
-
-            else:
-                raise ValueError("Transformation '%s' doesn't exists" % trans)
```
jd: How is it that you can remove that? I miss something.

chungg: it never gets to this code which is why the tests never saw the error uncaught. when the api validates the transformation, it already gives a list of known methods: https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/rest/transformation.py#L54

jd: ok-ish. I still like to think this is an independent layer that can be used elsewhere so I like having that here.

chungg:
added test for carbonara. can't test api as i mentioned, transform validation already raises issue earlier. will have api tests in the when i add rolling transform (or in sileht's math transform)

PR #361, opened 2017-09-16
doc: Ensure doc/build is empty

sileht:


sileht:
fast merging to fix the online doc

Issue #364, opened 2017-09-19
Measure and Metric are not storage related


chungg:
we capture Measure and Metric in storage/__init__.py but neither one is related to storage.

Measure is used as model for incoming data from either REST or statsd

Metric is used by indexer

we should move code to right place.

PR #365, opened 2017-09-19
move Measure model to incoming and Metric to indexer

chungg:
Measure is only leveraged by incoming driver.


chungg:
we could also create a models module... i'm indifferent.

chungg:
related to #364

jd:
having models modules or exception modules give me headache. it's like you'd have a variable and a function module lol.

Issue #340, opened 2017-09-13
Documentation improvement


Asu4ni:
I'm feeling that our doc is lacking tons of details and there are so much stuf that one can't figure out until browsing through the codes. I'll give some examples down below, and I'd help adding the content now and then. 

the list of things to be improved:
- [ ] Glossary: needs more items such as timespan, archive policy rule, resource type, etc. and anchors should be placed for linking.
- [ ] Details of each entities: for entities like resource, metric, measure, etc., the relation between them (e.g. resource contains metrics), the attributes of them (e.g. metric has creator) should be explicitly explained, listed and illustrated in a page.
- [ ] Aggregation method: a complete list showing all ones and the explanations of them.
- [ ] REST API: the usage of each api should be explained in details such as what attribute names are invalid.

If I come up more or someone suggests some, I'll update the list

Asu4ni:
@jd one question. How can I preview the result of my alterations?

jd:
@Asu4ni Run `tox -e docs-gnocchi.xyz` and check `doc/build/html` for output.

Asu4ni:
I cloned the repo and ran the tox command. The page was correctly displayed.
Afterwards, I modified the rst file and ran the command. The html files were updated (checked the modified time) but the contents were the same. Even if I commited my change, it still didn't work.
I'm digging into the doc of sphinx now, but if someone catches what I've missed, please tell me. Thanks!

Asu4ni:
The error message of the above problem which might be the cause.
`Exception: Versioning for this project requires either an sdist tarball, or access to an upstream git repository. It's also possible that there is a mismatch between the package name in setup.cfg and the argument given to pbr.version.VersionInfo. Project name gnocchi was given, but was not able to be found.`

jd:
@Asu4ni ah don't bother, it's likely sphinx versioncontrib give you problems. Just run `tox -e docs` in your case, that'll be enough.

Asu4ni:
I think we need to distinguish between 'incoming measures' and 'processed measures'.

I'm considering this because our [glossary](http://gnocchi.xyz/glossary.html#) in the document have the item 'measure'. But it does not explicitly show which one it represents(incoming or processed).

I'd like to gain some thoughts from others' opinion. Mine is that we should use two words for these two concepts. Maybe 'measure' for incoming and 'aggregates' for processed.

Or I simply mistook the concepts. XD

jd:
[...]

PR #367, opened 2017-09-20
Glossary: add 'aggregate' and modify the doc

Asu4ni:
related #340

  1. add item 'aggregate': differentiate aggregates from measures
  2. add item 'timespan'
  3. modify measures to aggregates if needed
  4. glossary minor fixes


doc/source/glossary.rst
```diff
@@ -3,31 +3,42 @@ Glossary
...
 
    Measure
-     A datapoint tuple composed of timestamp and a value.
+     An incoming datapoint tuple sent to Gnocchi by the api. It is composed
+     of a timestamp and a value.
 
    Archive policy
-     A measure storage policy attached to a metric. It determines how long
-     measures will be kept in a metric and how they will be aggregated.
+     A aggregate storage policy attached to a metric. It determines how long
```
jd: An aggregate?

Asu4ni:
I only changed the document parts. I don't understand why other parts failed.

Asu4ni:
@chungg 
> can you leave the picture as is
I could, but why is it the case? Is there any problems with my image?

Asu4ni:
@chungg, I remade the image not only because the text should be changed but also because the original one is less organized. Is it fine to use the image I made?

jd:
I like the new image, and it's in SVG, so that's better. Which software did you use to build it?
I think it misses the link between coordinator and API (which barely use it, but still a bit). Maybe it was missing in the old schema too, but we should add it for completeness I think.

chungg:
i don't have a problem with picture per se. it's just we need to know what tool was used or it'll be impossible to actually edit the picture later if you're not around :)

my only question is why the resolution is so small (or maybe it's just how github formats svg)... i have to change my browser to zoom in 300% to view it.

if i were to give feedback. i'd say drop the 'gnocchi' part of api and metricd. i'm pretty sure everyone knows we're talking about gnocchi

jd:
> i don't have a problem with picture per se. it's just we need to know what tool was used or it'll be impossible to actually edit the picture later if you're not around :)

Yeah, my question. But it's SVG so it's vectorial and so actually editable with any software that knows SVG – so that's better than my crappy PNG screenshoted from Google Docs :p 

> my only question is why the resolution is so small (or maybe it's just how github formats svg)... i have to change my browser to zoom in 300% to view it.

Yes it's vectorial so you can zoom infinitely (DON'T TRY THOUGH) if you want, that's a github/browser issue.

> if i were to give feedback. i'd say drop the 'gnocchi' part of api and metricd. i'm pretty sure everyone knows we're talking about gnocchi

+1

Asu4ni:
### Tool
It's Google Drawing. [Link](https://docs.google.com/drawings/d/1aHV86TPNFt7FlCLEjsTvV9FWoFYxXCaQOzfg7NdXVwM/edit?usp=sharing) I'm not sure how we will manage the share setting. Letting anyone with the link to edit it will be a mess.

### link between coordinator & api
Should it be a bi-direction arrow or a normal arrow (which side to point?) ? I don't know the relation between them. Please inform me of that.

jd:
> I'm not sure how we will manage the share setting. Letting anyone with the link to edit it will be a mess.

Just read only share is OK. Add the link as a comment in the source of the rst so we know where to go to create a copy and edit.

> Should it be a bi-direction arrow or a normal arrow (which side to point?) ? I don't know the relation between them. Please inform me of that.

Same as metricd, bi directional I'd say.

Asu4ni:
This link is for viewing: https://docs.google.com/drawings/d/1aHV86TPNFt7FlCLEjsTvV9FWoFYxXCaQOzfg7NdXVwM/edit?usp=sharing

And this link is for copying:
https://docs.google.com/drawings/d/1aHV86TPNFt7FlCLEjsTvV9FWoFYxXCaQOzfg7NdXVwM/copy

Which should I use? or both?
And what kind of format to put it in? In the 'alt' of the image?

jd:
go with the view link

you can put comments in RST:
 http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#comments

jd:
@Asu4ni it'd be good you squash all of those commits in one I think

doc/source/glossary.rst
```diff
@@ -3,31 +3,42 @@ Glossary
...
-     Function used to aggregate multiple measures in one. For example, the
-     `min` aggregation method will aggregate the values of different measures
-     to the minimum value of all the measures in time range.
+     Function used to aggregate multiple measures into an aggregate. For
+     example, the `min` aggregation method will aggregate the values of
+     different measures to the minimum value of all the measures in the time
+     range.
+
+   Aggregate
+     An datapoint tuple generated from several measures according to the
```
chungg: s/An/A/

doc/source/nagios.rst
```diff
@@ -10,7 +10,7 @@ service. It can be installed with pip::
 `Source`_ and `Documentation`_ are also available.
 
 Gnocchi-nagios collects perfdata files generated by `Nagios`_ or `Icinga`_;
-transforms them into Gnocchi resources, metrics and measures format; and
+transforms them into Gnocchi resources, metrics and aggregates format; and
```
chungg: i don't think nagios is passing in an aggregate.

PR #372, opened 2017-09-21
Simplify error handling in IncomingDriver.get_storage_sacks

hguemar:
In some race conditions, the get_storage_sacks method gets called too
early. And exceptions were caught in get_storage_sacks and preventing
callers to generate meaningful logs.

Discovered during gnocchi sprint at pycon.fr 2017


chungg:
looks good to me. could you squash your commits into one :)

PR #377, opened 2017-09-22
[doc] add glossary anchor

Asu4ni:
related #340 
add glossary anchor
little fix: move images into folder and change display size


Asu4ni:
not finish yet

Asu4ni:
done

Asu4ni:
I'll squash all commits to one after others have examined them.

doc/source/glossary.rst
```diff
@@ -7,38 +7,42 @@ Glossary
 
    Resource
      An entity representing anything in your infrastructure that you will
-     associate metric(s) with. It is identified by a unique ID and can
+     associate |metric|\ (s) with. It is identified by a unique ID and can
```
chungg: should there be a random space between \ and (s)

Asu4ni: that is to make | separated with (, but there should not be space there. Putting a \ before it makes it a functional space.

PR #378, opened 2017-09-22
Fix fake metricd used in gabbi

sileht:
storage.process_background_tasks() was originally locks each metrics it
processes. But since we move to the sack system the locks are handled by
the caller.

So, the fake-metricd thread of gabbi does not lock metric is process
anymore. This can lead to race when an API call use refresh=true and
metricd also try to process the same metric.

This change fixes that by also using refresh_metric in the fake metricd
used in gabbi tests.

Closes-bug: #244


sileht:
Pull-request updated, HEAD is now adc6f2b701f8e46df1ad45614c5dddb34fac2fd1

PR #381, opened 2017-09-22
statsd: don't fetch resource on each flush()

### LINKED ISSUE #375: statsd: don't get resource everytime we flush
we shouldn't need to get resource and it's metrics every time we flush. just save it.

https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/statsd.py#L91-L93

jd:
Fixes #375

### DIFF
--- a/gnocchi/statsd.py
+++ b/gnocchi/statsd.py
@@ -50,6 +50,13 @@ class Stats(object):
         self.gauges = {}
         self.counters = {}
         self.times = {}
+        self.metrics = {
+            metric.name: metric
+            for metric
+            in self.indexer.get_resource('generic',
+                                         self.conf.statsd.resource_id,
+                                         with_metrics=True).metrics
+        }
 
     def reset(self):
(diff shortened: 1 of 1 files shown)


hguemar:
Change do fix the issue. 👍

jd:
Pull-request updated, HEAD is now e80867ed8ebc49e68e74c141ec788ef907eab3a7

gnocchi/statsd.py
```diff
@@ -110,6 +113,7 @@ def flush(self):
                         archive_policy_name=ap_name,
                         name=metric_name,
                         resource_id=self.conf.statsd.resource_id)
```
chungg: can do this in another patch but we should try to catch NamedMetricAlreadyExists if someone creates metric via api

jd:
Pull-request updated, HEAD is now 9756cb2f0e5af79d87899132c7e7d4dd9cd20f8e

PR #382, opened 2017-09-22
Allow to rename archive policy rules

### LINKED ISSUE #374: Allow to rename archive policy rules
Currently, the archive policy rules cannot be renamed. It would be cool to allow them to be renamed.

0livd:
It is now possible to change the name of an archive policy rule.

Closes #374

### DIFF
--- a/doc/source/rest.j2
+++ b/doc/source/rest.j2
@@ -302,6 +302,10 @@ It is possible to delete an |archive policy| rule:
 
 {{ scenarios['delete-archive-policy-rule']['doc'] }}
 
+It is possible to rename an archive policy rule:
+
+{{ scenarios['rename-archive-policy-rule']['doc'] }}
+
 Resources
 =========
 
--- a/gnocchi/tests/functional/gabbits/archive-rule.yaml
+++ b/gnocchi/tests/functional/gabbits/archive-rule.yaml
@@ -158,6 +158,47 @@ tests:
         authorization: "basic YWRtaW46"
       status: 400
 
+# rename an archive policy rule
+
+    - name: rename archive policy rule with missing name
+      PATCH: /v1/archive_policy_rule/test_rule3
+      request_headers:
+        # User admin
+        authorization: "basic YWRtaW46"
+      status: 400
+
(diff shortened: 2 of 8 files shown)


doc/source/rest.j2
```diff
@@ -300,6 +300,10 @@ It is possible to delete an archive policy rule:
 
 {{ scenarios['delete-archive-policy-rule']['doc'] }}
 
+It is possible to remane an archive policy rule:
```
jd: rename

0livd: Corrected :)

gnocchi/tests/functional/gabbits/archive-rule.yaml
```diff
@@ -158,6 +158,32 @@ tests:
...
+
+    - name: rename archive policy rule with existing name
+      PATCH: /v1/archive_policy_rule/test_rule2
+      request_headers:
+        # User admin
+        authorization: "basic YWRtaW46"
+      data:
+          name: test_rule3_renamed
+      status: 400
+
```
chungg: could you maybe add a check to verify test_rule3 (original) doesn't exist after rename

0livd: Done, added a test with missing name parameter too

jd:
It'd be great to squash all of those in a single commit, but ok.

gnocchi/tests/functional/gabbits/archive-rule.yaml
```diff
@@ -158,6 +158,45 @@ tests:
...
```
jd: would be great to check for the reason of that status 400.

0livd:
Ah sorry I just rebased the whole thing but it dismissed your reviews...I have to check out how to use pastamaker...

gnocchi/tests/functional/gabbits/archive-rule.yaml
0livd: I added a check for the response string

jd:
Yeah it's normal that reviews disappear don't worry. :)

PR #383, opened 2017-09-22
Apply archive policy rules by name order

### LINKED ISSUE #199: ambiguous archive policy matching if duplicate patterns
we allow archive policy rule pattern to be duplicated. this creates ambiguity as we have no idea which rule it will pick... currently it matches with whatever order sql returns back (i assume create time).

we should not let duplicate archive rules with same pattern  OR we need to allow it to be scope to project (and still not let it have same pattern)

jd:
Fixes #199

### DIFF
--- a/gnocchi/indexer/sqlalchemy.py
+++ b/gnocchi/indexer/sqlalchemy.py
@@ -619,7 +619,9 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
     def list_archive_policy_rules(self):
         with self.facade.independent_reader() as session:
             return session.query(ArchivePolicyRule).order_by(
-                ArchivePolicyRule.metric_pattern.desc()).all()
+                ArchivePolicyRule.metric_pattern.desc(),
+                ArchivePolicyRule.name.asc()
+            ).all()
 
     def get_archive_policy_rule(self, name):
         with self.facade.independent_reader() as session:


chungg:
so my original issue was this scenario:

```
ap_ruleA = {'name': 'blahA', 'pattern': 'disk.*', ...}
ap_ruleB = {'name': 'blahB', 'pattern': 'disk.*', ...}
```
in this scenario, if i have metric disk.read.requests, i will get blahA. which i guess makes sense?

but say we have

```
ap_ruleA = {'name': 'blahA', 'pattern': 'disk.*', ...}
ap_ruleB = {'name': 'blahB', 'pattern': 'disk.*', ...}
ap_ruleC = {'name': 'blahC', 'pattern': 'disk.read.*', ...}
```
i would want metric disk.read.requests to match with blahC since it's more precise but it will match with blahA?

jd:
> in this scenario, if i have metric disk.read.requests, i will get blahA. which i guess makes sense?

you don't know actually. it's just "random" and depends on the database.

> i would want metric disk.read.requests to match with blahC since it's more precise but it will match with blahA?
It will match blahA because the name is before blahC. This is deterministic.

chungg:
> It will match blahA because the name is before blahC. This is deterministic.

right, but it was deterministic before as well -- it matched with the most precise match. it was only random if there were two rules with the exact same pattern. so if we made pattern unique, it would still work.

it seems with this change, the user will need to name their rules appropriately which was something already done in pattern.

jd:
Pull-request updated, HEAD is now 4a58f6086857dfb32a664cb85bfb66ee17764c96

jd:
@chungg ack, I've changed the patch so I think it fixes the problem with a simpler solution

gnocchi/indexer/sqlalchemy.py
```diff
@@ -619,7 +619,9 @@ def create_archive_policy(self, archive_policy):
     def list_archive_policy_rules(self):
         with self.facade.independent_reader() as session:
             return session.query(ArchivePolicyRule).order_by(
-                ArchivePolicyRule.metric_pattern.desc()).all()
+                ArchivePolicyRule.metric_pattern.asc(),
```
chungg: should we keep this descending? i think it's best if it matches the most specific pattern before the more generic pattern.

ie. match disk.io.* before disk.* or *

jd: Of course, me stupid.

jd:
Pull-request updated, HEAD is now 50a34923c0411df6fdc367cfe0ead65aa8b309a8

gnocchi/indexer/sqlalchemy.py
```diff
@@ -619,7 +619,9 @@ def create_archive_policy(self, archive_policy):
     def list_archive_policy_rules(self):
         with self.facade.independent_reader() as session:
             return session.query(ArchivePolicyRule).order_by(
-                ArchivePolicyRule.metric_pattern.desc()).all()
+                ArchivePolicyRule.metric_pattern.desc(),
```
Asu4ni: I'm not sure of the idea.
This does not guarantee that the longest pattern will appear in the front. For example:
```
pattern1: abc.*
pattern2: abc.(*)
```
metric `abc.(xyz)` should match `pattern2`, but if `metric_pattern.desc()` is used, `pattern1` is matched.

jd: Haha, that's exactly why I'm "WTF this is a bad idea" since the beginning but I was not able to find a proper example. Thanks @Asu4ni.

However, this is completely unrelated to this patch as this PR does not change that.

But the original PR was trying to fix that problem by ordering by pattern name. Maybe it's better.

When do we remove this archive policy rule entirely, @chungg ? :)

Asu4ni: Ceilometer now allows to specify the policy in each sink, so I guess removing it is fine.
BTW, in case one don't see, I comment above that should we add this point(name ordering) to the document before this merged?

jd: Yeah we can't really **remove** it, it'd break the API, but that's worth noting anyway. We probably need to document it, right. I'll wait for @gordc feedback. :)

chungg: i'm indifferent. i guess we could just sort by length after we query but *shrugs* does it matter. i'm ok with as is and documenting/deprecating.

as you folks mention, the rules will always be sort of ambiguous because of regex.

PR #388, opened 2017-09-27
Doc - add note regarding metric modification

### LINKED ISSUE #386: How to modify a archival policy rule?
Hey,

I had a question. Suppose I was using the default archival policy rule, which basically matches everything and maps it to archival policy "low". Now i deleted this rule and created another one which maps everything to "medium".

How can I update the exciting metrics to follow this rule in future?

Asu4ni:
close #386 
  In api usage, add note that it is not possible to modify
  metric after creation.

### DIFF
--- a/doc/source/rest.j2
+++ b/doc/source/rest.j2
@@ -39,6 +39,12 @@ To create a |metric|, the following API request should be used:
 
 {{ scenarios['create-metric']['doc'] }}
 
+.. note::
+
+  Once the |metric| is created, the |archive policy| attribute is fixed and
+  unchangeable. The definition of the |archive policy| can be changed through
+  the :ref:`archive_policy endpoint<archive-policy-patch>` though.
+
 Once created, you can retrieve the |metric| information:
 
 {{ scenarios['get-metric']['doc'] }}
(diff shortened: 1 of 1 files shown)


doc/source/rest.j2
```diff
@@ -39,6 +39,11 @@ To create a |metric|, the following API request should be used:
 
 {{ scenarios['create-metric']['doc'] }}
 
+.. note::
+
+  Once the |metric| is created, it cannot be PATCHed to modify the attributes.
+  Especially note that |archive policy| is fixed after metric creations.
```
jd: fixed? how?

doc/source/rest.j2
```diff
@@ -39,6 +39,11 @@ To create a |metric|, the following API request should be used:
 
 {{ scenarios['create-metric']['doc'] }}
 
+.. note::
+
+  Once the |metric| is created, it cannot be PATCHed to modify the attributes.
```
jd: You can modify the resource and name, but not by patching but by going through the resource API.

I think just saying that you cannot change the archive policy is what we want here.

Asu4ni: Thanks, there are so many details I have to pick up like this. I really think the document should be upgraded to improve the learning curve. Most information is there but not organized well.

Asu4ni: I just digged into the document and was not able to find where it mentioned metric's name and resource modification.

Asu4ni:
I just digged into the document and was not able to find where it mentioned metric's name and resource modification.

chungg:
but you can change the details of the archive policy...

Asu4ni:
Sure, this just mentions in the doc that the archive policy of a metric used cannot be changed to another ap. I'll update the text, maybe it is not clear enough.

PR #389, opened 2017-09-27
Allow to rename archive policies

0livd:
Archive policies can be renamed by issuing
a PATCH request containing the new name.
The definition of an ap can also be patched
along with the name within the same request.

### DIFF
--- a/doc/source/rest.j2
+++ b/doc/source/rest.j2
@@ -249,6 +249,15 @@ not retroactively calculated as backfill to accommodate the new |timespan|:
    |Granularities| cannot be changed to a different rate. Also, |granularities|
    cannot be added or dropped from a policy.
 
+Existing |archive policies| can be renamed:
+
+{{ scenarios['rename-archive-policy']['doc'] }}
+
+.. note::
+
+    The `name` and `definition` parameters can both be specified to update and
+    rename an archive policy within the same PATCH request.
+
 It is possible to delete an |archive policy| if it is not used by any |metric|:
 
 {{ scenarios['delete-archive-policy']['doc'] }}
--- a/gnocchi/indexer/alembic/versions/815b184585a5_use_cascade_update_for_metric_table_s_.py
+++ b/gnocchi/indexer/alembic/versions/815b184585a5_use_cascade_update_for_metric_table_s_.py
@@ -1,0 +1,43 @@
+# Copyright 2017 The Gnocchi Developers
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+#
+
+"""Use cascade update for metric table's ap foreign key
+
+Revision ID: 815b184585a5
+Revises: 1e1a63d3d186
+Create Date: 2017-09-25 17:38:23.954423
+
+"""
+
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '815b184585a5'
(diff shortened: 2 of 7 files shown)


doc/source/rest.j2
```diff
@@ -249,6 +249,15 @@ not retroactively calculated as backfill to accommodate the new |timespan|:
...
    cannot be added or dropped from a policy.
 
+Existing |archive policies| can be renamed:
+
+{{ scenarios['rename-archive-policy']['doc'] }}
+
+.. note::
+
+    The `name` and `definition` parameters can both be specified to update and
+    rename an archive policy within the same PATCH request.
```
Asu4ni: Would you mind adding the replace rule - `|archive policy|` ?

gnocchi/indexer/alembic/versions/815b184585a5_use_cascade_update_for_metric_table_s_.py
```diff
@@ -0,0 +1,43 @@
...
+def upgrade():
+    op.drop_constraint("fk_metric_ap_name_ap_name",
+                       "metric",
+                       type_="foreignkey")
+    op.create_foreign_key("fk_metric_ap_name_ap_name",
+                          "metric", "archive_policy",
+                          ["archive_policy_name"],
+                          ["name"],
+                          ondelete='RESTRICT',
+                          onupdate="CASCADE")
```
jd: Does this any perf impact? If you rename a AP that is used by 1M metrics, that might be a bit slow?
Just wondering. :)

gnocchi/indexer/sqlalchemy.py
```diff
@@ -568,25 +568,33 @@ def get_archive_policy(self, name):
...
-            if not ap:
-                raise indexer.NoSuchArchivePolicy(name)
-            current = sorted(ap.definition,
-                             key=operator.attrgetter('granularity'))
-            new = sorted(ap_items, key=operator.attrgetter('granularity'))
-            if len(current) != len(new):
+    def update_archive_policy(self, name, ap_items, new_name=None):
+        ap = self.get_archive_policy(name)
+        if not ap:
+            raise indexer.NoSuchArchivePolicy(name)
```
jd: I don't know if it has a great impact, but changing this creates 2 session rather than one.

gnocchi/indexer/sqlalchemy.py
```diff
@@ -568,25 +568,33 @@ def get_archive_policy(self, name):
...
-            return ap
+                    name, '%s granularity interval was changed'
+                    % utils.timespan_total_seconds(c.granularity))
+        # NOTE(gordc): ORM doesn't update JSON column unless new
+        ap.definition = ap_items
+        if new_name:
+            ap.name = new_name
+        try:
+            with self.facade.independent_writer() as session:
+                session.add(ap)
```
jd: Hm I'd expect this to do an insert rather than an update?

gnocchi/indexer/alembic/versions/815b184585a5_use_cascade_update_for_metric_table_s_.py
0livd: Another solution, which would offer better performances and avoid a mutable primary key, is to add an ``id`` column to the AP table and make it the primary key. The migration would be way trickier though...I can propose something if you prefer it to be implemented that way.

gnocchi/indexer/sqlalchemy.py
0livd: Yeah that's weird, the object comes from a different session and the primary key might be changed so I'd except an insert too but according to the tests it seems to be doing an update in any case... I haven't digged deeper but I think I'll just revert to the old structure of the ``update_archive_policy`` method to avoid this weird behavior and the creation of 2 sessions.

gnocchi/indexer/alembic/versions/815b184585a5_use_cascade_update_for_metric_table_s_.py
jd: Yes, I think it's the good solution to go. Agree it's gonna be trickier, but it's the good way to go. Go for it! :)

gnocchi/indexer/sqlalchemy.py
jd: Ok, cool!

PR #392, opened 2017-09-28
numpy instead of pandas cross-metric

chungg:
use numpy instead of pandas to compute cross metric
- find all the unique timestamps
- create a AxB pre-filled grid where
	- A is number of series, and B is length
- if fill, fill as appropriate
- if not fill
	- if needed_overlap > 0, compute overlap
	- reset start/end boundaries if not provided to first/last
	  times which overlap all series.
- aggregate as needed.

### DIFF
--- a/gnocchi/rest/cross_metric.py
+++ b/gnocchi/rest/cross_metric.py
@@ -17,16 +17,23 @@
 """Timeseries cross-aggregation."""
 import collections
 
-
 import daiquiri
-import iso8601
-import pandas
+import numpy
 import six
 
+from gnocchi import carbonara
 from gnocchi import storage as gnocchi_storage
 
 
 LOG = daiquiri.getLogger(__name__)
+
+
+AGG_MAP = {'mean': numpy.nanmean,
+           'median': numpy.nanmedian,
+           'std': numpy.nanstd,
+           'min': numpy.nanmin,
+           'max': numpy.nanmax,
+           'sum': numpy.nansum}
 
 
 class UnAggregableTimeseries(Exception):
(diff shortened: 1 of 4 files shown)


chungg:
if i rip the following test:
https://github.com/chungg/gnocchi/blob/265164b9cb980ac406d86ca2241d8d76d35a3619/gnocchi/tests/test_cross_metric.py#L346-L371

i get 100x improvement
```
In [155]: %timeit output = cross_metric.aggregated([tsc1['return'], tsc2['return']], aggregation='mean', fill=0)
The slowest run took 4.08 times longer than the fastest. This could mean that an intermediate result is being cached.
100 loops, best of 3: 15.2 ms per loop

In [156]: %timeit output = cross_metric.aggregated_fill([tsc1['return'], tsc2['return']], aggregation='mean', fill=0)
The slowest run took 4.75 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 107 µs per loop
```

gnocchi/rest/cross_metric.py
```diff
@@ -17,18 +17,27 @@
...
 import six
 
+from gnocchi import carbonara
 from gnocchi import storage as gnocchi_storage
 
 
 LOG = daiquiri.getLogger(__name__)
 
 
+agg_map = {'mean': numpy.nanmean,
```
jd: AGG_MAP would be better to quickly see it's a kinda module-wide constant

chungg: will do

sileht:
We may default "fill" to 0, now ?

chungg:
this is ~100x faster similar to fill path:

```
In [29]: %timeit output = (cross_metric.aggregated([tsc1['return'], tsc2['return']], aggregation='mean', needed_percent_of_overlap=25.0))
10000 loops, best of 3: 168 µs per loop

In [30]: %timeit output = (cross_metric.aggregated_old([tsc1['return'], tsc2['return']], aggregation='mean', needed_percent_of_overlap=25.0))
100 loops, best of 3: 16.9 ms per loop

```
i will add a disclaimer. i know two scenarios which makes this code behave differently than current code but both paths are untested:

- seriesA: t1, t2, t3, t6; seriesB: t0, t1, t2, t3, t4, needed_overlap=0.0
 - proposed code returns: t0, t1, t2, t3, t4, t6
 - current code returns: t1, t2, t3
- seriesA: t1, t2, t4, t6; seriesB:  t0, t1, t2, t3, t4; needed_overlap=25.0
 - proposed code returns: t1, t2, t3, t4
 - current code returns: nothing

chungg:
> seriesA: t1, t2, t3, t6; seriesB: t0, t1, t2, t3, t4, needed_overlap=0.0
> proposed code returns: t0, t1, t2, t3, t4, t6
> current code returns: t1, t2, t3

i can fix this scenario. the other one... not so much, and i don't think i want to fix it because it's not predictable.

PR #399, opened 2017-10-02
gnocchi-api: add support for VIRTUAL_ENV

jd:
uwsgi does not honor VIRTUAL_ENV by default, which leads to problem when
testing gnocchi-api in a virtual env.


sileht:
I think the issue is when uswgi is not install in the virtualenv. I don't like -H it can lead to other issue.

This make venv python start non venv python that load some modules (not all) from the virtualenv.

jd:
> I think the issue is when uswgi is not install in the virtualenv. I don't like -H it can lead to other issue.

It is in my venv for example, and it's loading stuff from /usr/local

> This make venv python start non venv python that load some modules (not all) from the virtualenv.

Now I'm confused.

sileht:
I remember you got the same issue in pifpaf, that was due to the uwsgi inside the virtualenv that was not compiled correctly or without python support, something like that.

jd:
Yeah I remember reverting that. I've sent a PR on pifpaf that seems to work ok, the same we merged in June. I'm gonna give it a try tomorrow and releasing to see if it breaks anything.

jd:
@sileht I've pushed a pifpaf release yesterday with that same change that fixes my problem, and it seems the world did not explode, so I think it's good. Maybe now it is, back then it broke something I don't remember what obviously.

jd:
This make sure that when `gnocchi-api` is executed, its Python environment takes into account any virtual env being used. This is the case for e.g. metricd, but not for gnocchi-api because uwsgi does not respect that by default.

So if you are in a Python venv, gnocchi-metricd will work but gnocchi-api will not, or if it does work, then it will load the *system wide* Gnocchi that is installed, no the one in the virtual env.

I already pushed the fix in pifpaf a few weeks ago and it works fine. :)

PR #401, opened 2017-10-03
metricd: only re-add sack to be processed on failure

### LINKED ISSUE #397: notifier goes crazy after emptying incoming
so, the incoming notifier seems to work really well in terms of improving real-time job scheduling. it clears the incoming storage really fast (not official testing, i just think it's moving well)

but one thing i noticed was after it clears incoming storage, it seems to keep attempting to process storage for a while afterwards until eventually it returns to whatever processing_delay was set. for example, my environment is set to processing_delay = 60 but for at least 60s after reporter tells me incoming is empty, the processing workers are still trying to process sacks non-stop.

there seems to be a backlog of process events which it isn't clearing.

jd:
The condition for re-adding the sack on failure has been put under finally by
mistake, rather than under the except block.

Fixes #397

### DIFF
--- a/gnocchi/cli/metricd.py
+++ b/gnocchi/cli/metricd.py
@@ -229,11 +229,11 @@ class MetricProcessor(MetricProcessBase):
             except Exception:
                 LOG.error("Unexpected error processing assigned job",
                           exc_info=True)
-            finally:
-                lock.release()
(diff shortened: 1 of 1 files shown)


chungg:
good catch. i'll test it out.

jd:
@chungg ack, I'll keep investigating :)

PR #403, opened 2017-10-04
rest: expose the gnocchi build

### LINKED ISSUE #402: Expose Gnocchi version in API
The current API only provide an useless information about version. We dd new stuff into the API but we keep the version 1.0 and updated is set to "2015-03-19" because we don't break anything.

On another side, client application want to known if a new thing in the API is usable.

For example, I have no way to discover : 
* if the batch GET metric is avialable.
* if transform= exists
* if resample= exists
...

I see two solutions:
1. Put the Gnocchi server version somewhere in the API
  - it's lazy but we can't forget to update it
  - it's easy to do
  
2. Add a new version on each API changes in API controller even if all changes are hosted on /v1/ endpoint. (Like Openstack microversion)
  - Microversion is cool when you have multiple applications that implements the same API but I don't think Gnocchi will be this kind of API
  - We have to take care of that version on API review
  - I don't see the point for us to have mapping like
    * Gnocchi 4.0 == API 1.0
    * Gnocchi 4.1 == API 1.1
    * Gnocchi 4.2 == API 1.1
    * Gnocchi 4.3 == API 1.2
    * Gnocchi 5.0 == API 1.3 AND API 2.0
  - We build an application not the spec of an API.

Go for 1/ ?

sileht:
Closes #402

### DIFF
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -20,6 +20,7 @@ import uuid
 import warnings
 
 import jsonpatch
+import pbr.version
 import pecan
 from pecan import rest
 import pyparsing
@@ -1972,6 +1973,7 @@ class VersionsController(object):
     @pecan.expose('json')
     def index():
         return {
+            "build": pbr.version.VersionInfo('gnocchi').version_string(),
             "versions": [
                 {
                     "status": "CURRENT",
--- a/gnocchi/tests/test_rest.py
+++ b/gnocchi/tests/test_rest.py
@@ -26,6 +26,7 @@ import uuid
 import iso8601
 from keystonemiddleware import fixture as ksm_fixture
 import mock
+import pbr.version
 import six
 from stevedore import extension
 import testscenarios
@@ -212,6 +213,13 @@ class RootTest(RestTest):
             self.assertEqual(
                 ['test_aggregation'],
                 result['dynamic_aggregation_methods'])
+
+    def test_version(self):
+        with self.app.use_admin_user():
+            r = self.app.get("/")
+        self.assertEqual(
+            json.loads(r.text)['build'],
+            pbr.version.VersionInfo('gnocchi').version_string())
 
     def test_status(self):
         with self.app.use_admin_user():


sileht:
Pull-request updated, HEAD is now 1b72ea4443bcd1e2747089d85913558a6525a647

gnocchi/tests/functional/gabbits/base.yaml
```diff
@@ -15,6 +15,7 @@ tests:
   response_headers:
       content-type: /^application\/json/
   response_json_paths:
+      $.build[0]: "5"
```
chungg: does this mean we have to change this every single time we tag?

jd: Yes, at least each time we change the major version pbr thinks we're going to release.

@sileht what bothers me is that pbr thinks we'll release 5 (which I can tblame it for) but when we'll tag 4.1.0 this test will fail for that particular tag.

chungg: :( seems like a pain in the ass... but i have no suggestion so meh.

i thought pbr appends devXYZ to the end of master builds where XYZ is changing all the time?

sileht:
Pull-request updated, HEAD is now 46dbf0f4760e797e13cfc2b31f811122fdc6c77c

sileht:
Pull-request updated, HEAD is now 46dbf0f4760e797e13cfc2b31f811122fdc6c77c

sileht:
Pull-request updated, HEAD is now 1b72ea4443bcd1e2747089d85913558a6525a647

sileht:
I will rewrite a test in test_rest that mock pbr

sileht:
Pull-request updated, HEAD is now f7a42b683e5b81db15ec7bbfa929118c21ae33c7

sileht:
Pull-request updated, HEAD is now 0cefd1332e62207a5b70bc0b9e952c2bab2b7fb4

sileht:
Pull-request updated, HEAD is now 0c1f8d549e6bef3e5f1d6bf59b2cbc6ef539e036

sileht:
Pull-request updated, HEAD is now 906d000c1de0b0a3f27ccf0f45f8b0c4d77a9582

PR #404, opened 2017-10-04
storage: remove process_background_tasks

jd:
The function does nothing anymore, just wraps process_new_measures. Remove it.


jd:
No, I think it was just there because of the non-Carbonara drivers we wanted to support, and then we merge `_carbonara` in `__init__` we just forgot to remove it. :)

PR #405, opened 2017-10-04
redis: do not emit notifications until sack is processed

jd:
The current behavior is to notify no matter what as soon as a new measure is
push into a metric, and therefore, into a sack.

This patches modifies this behaviour by having the measure storage putting a
flag into a special key named after the sack it is pushing into. Metricd
processes watches for this key to be set and get a notification _only the first
time it is set_. Once metricd finishes processing that sack, it delete the key
so it can get notified again if any new measures has arrived.

This patch also simplifies the notification processing in metricd by just
setting back on the notification system when processing is finished. Previous
version was trying to re-enable notification prior to processing but it just
complicates the workflow for little gain; there's no lock around notification,
and would just add overhead.

### DIFF
--- a/gnocchi/incoming/redis.py
+++ b/gnocchi/incoming/redis.py
@@ -41,16 +41,24 @@ class RedisStorage(incoming.IncomingDriver):
         # NOTE(gordc): redis doesn't maintain keys with empty values
         pass
 
+    def _build_measure_path_with_sack(self, metric_id, sack_name):
+        return redis.SEP.join([sack_name.encode(), str(metric_id).encode()])
+
     def _build_measure_path(self, metric_id):
-        return redis.SEP.join([
-            self.get_sack_name(self.sack_for_metric(metric_id)).encode(),
-            str(metric_id).encode()])
+        return self._build_measure_path_with_sack(
+            metric_id, self.get_sack_name(self.sack_for_metric(metric_id)))
 
     def add_measures_batch(self, metrics_and_measures):
+        notified_sacks = set()
         pipe = self._client.pipeline(transaction=False)
         for metric, measures in six.iteritems(metrics_and_measures):
-            path = self._build_measure_path(metric.id)
+            sack_name = self.get_sack_name(self.sack_for_metric(metric.id))
+            path = self._build_measure_path_with_sack(metric.id, sack_name)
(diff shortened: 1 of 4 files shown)


gnocchi/incoming/redis.py
```diff
@@ -110,13 +118,19 @@ def process_measure_for_metric(self, metric):
...
 
     def iter_on_sacks_to_process(self):
-        self._client.config_set("notify-keyspace-events", "El")
+        self._client.config_set("notify-keyspace-events", "K$")
         p = self._client.pubsub()
         db = self._client.connection_pool.connection_kwargs['db']
-        channel = b"__keyevent@" + str(db).encode() + b"__:rpush"
-        p.subscribe(channel)
+        keyspace = b"__keyspace@" + str(db).encode() + b"__:"
+        pattern = keyspace + self.SACK_PREFIX.encode() + b"*"
```
chungg: won't this still match with the existing keys? if we have metricA-uuid that gets pushed to sack 5, it will push to incoming128-5:metricA-uuid and it will set incoming128-5?

won't incoming* match with both?

jd: No because the `notify-keyspace-events` is set to `K$` and the `$` means it will only send string operations such as `setnx` that we use here. The rest of Redis operations we use are e.g. `rpush` and that's not matched by this filter.

chungg: i see. i'll test then.

gnocchi/tests/test_incoming.py
```diff
@@ -58,6 +58,7 @@ def _iter_on_sacks_to_process():
                 break
             # NOTE(jd) Retry to send measures. It cannot be done only once as
             # there might be a race condition between the threads
+            self.incoming.finish_sack_processing(sack_to_find)
```
chungg: just curious, why we do this?

jd: It's possible another test uses that same sack, and therefore that the flag in Redis is already set for this. Since there's no metricd in those tests, nobody ever call `finish_sack_processing` so set the `setnx` call never send any notification.
By forcing this to be called and re-adding measure, we're sure that at some point this test will get a notification.

PR #408, opened 2017-10-04
metricd: cache sacks to process

### LINKED ISSUE #406: minimise run_watchers call
don't run_watchers everytime we check for our assigned sacks. this may be called hundreds/thousands of times a second and it doesn't make sense that we're hammering coordination agent.

jd:
With the new notification system calling _get_sack_to_process as often as each
notification is received, the tooz `run_watchers` function is called too often,
hitting the backend a lot, often for no change.

This patch restores back the previous behavior of updating the list of sack to
process every `metric_processing_delay` maximum. In order to do that, it simply
leverages cachetools and cache the output for a defined TTL equals to
`metric_processing_delay`. That avoids trying to be too smart about when/how to
call run_watchers while giving the wanted result.

Closes #406

### DIFF
--- a/gnocchi/cli/metricd.py
+++ b/gnocchi/cli/metricd.py
@@ -16,6 +16,7 @@
 import threading
 import time
 
+import cachetools.func
 import cotyledon
 from cotyledon import oslo_config_glue
 import daiquiri
@@ -126,6 +127,11 @@ class MetricProcessor(MetricProcessBase):
         # This stores the last time the processor did a scan on all the sack it
         # is responsible for
         self._last_full_sack_scan = utils.StopWatch().start()
+        # Only update the list of sacks to process every
+        # metric_processing_delay
+        self._get_sacks_to_process = cachetools.func.ttl_cache(
+            ttl=conf.metricd.metric_processing_delay
+        )(self._get_sacks_to_process)
 
     @tenacity.retry(
         wait=_wait_exponential,
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,3 +26,4 @@ daiquiri
 pyparsing>=2.2.0
 lz4>=0.9.0
 tooz>=1.38
+cachetools


gnocchi/cli/metricd.py
```diff
@@ -126,6 +127,12 @@ def __init__(self, worker_id, conf):
         # This stores the last time the processor did a scan on all the sack it
         # is responsible for
         self._last_full_sack_scan = utils.StopWatch().start()
+        # Only update the list of sacks to process every
+        # metric_processing_delay
+        self._get_sacks_to_process = cachetools.func.tll_cache(
```
chungg: this seems fine? although i should point out there already is a cache inside _get_sacks_to_process (self.group_state) i don't know if you want to keep it? unify it? *shrugs*

jd:
Fix typo tll -> ttl

gnocchi/cli/metricd.py
jd: Yeah but they are different cache, so I think both make sense. This TTL one is for run_watchers and hitting the coordination backend whereas the other one avoids recomputing the hash of every sack to see if assignation changed. So I think we're double-good. :)

jd:
Pull-request updated, HEAD is now d7a312c8a77c29405ba6a238775096576254078c

gnocchi/cli/metricd.py
```diff
@@ -126,6 +127,11 @@ def __init__(self, worker_id, conf):
         # This stores the last time the processor did a scan on all the sack it
         # is responsible for
         self._last_full_sack_scan = utils.StopWatch().start()
+        # Only update the list of sacks to process every
+        # metric_processing_delay
+        self._get_sacks_to_process = cachetools.func.ttl_cache(
+            ttl=conf.metricd.metric_processing_delay
```
chungg: we use to use worker_sync_rate to control this? we want to reuse it? i don't know if we want to force non-notifier drivers to have to sync every time?

https://github.com/gnocchixyz/gnocchi/blob/stable/4.0/gnocchi/cli.py#L212

jd: We removed `worker_sync_rate` to simplify the configuration and I think it's good that way. Keeping things simple to configure. You wont your worker to be in sync each time you process, so every `metric_processing_delay` seconds. I don't see the point to add yet another knob.

I don't understand your point of non-notifier though. It should not change anything since `_get_sack_to_process` will be called every `metric_processing_delay` so the cache will be invalidated each time – so no cache in the end and called every `metric_processing_delay` as  before.

chungg: right, but previously, if i put a low metric_processing_delay to simulate 'real-time', i could still leave the run_watchers to just redistribute tasks at a less aggressive interval.

but i see i approved the original commit so meh.

PR #417, opened 2017-10-06
incoming: fix logging on SackDetectionError

jd:
Accessing NUM_SACKS and having SackDetectionError is the mechanism that
upgrade() uses to detect if it needs to do its job. But since logging is down
too low-level, the error message is print on the first call of gnocchi-upgrade,
which does not make any sense. Log only where it makes sense.


jd:
Pull-request updated, HEAD is now 544b6e842c8cf3a61c8a7d2da78d7abbddc9b89d

PR #418, opened 2017-10-06
storage: remove aggregation_workers_number option

jd:
This has been set to 1 by default for a long time now as threads overhead are
usually bigger than the gain they offer. In medium-large scale deployment,
there's no interest to change that to something else than 1.

It also is being used by cross_metric aggregation where it is actually not
related to it. That's weird.

Let's remove it and simplify the code base a bit.


chungg:
original attempt: https://review.openstack.org/#/c/458238/

jd:
Ok, I'm good with keeping it, I'll fix it differently :)

PR #420, opened 2017-10-10
Set a correct default value for aggregation_workers_number

jd:
The current value of 1 is not the optimal setting by default, spreading on the
number of available CPU is likely to have better results to avoid being I/O
bound.


jd:
@sileht this is what I remembered too but he just denied that in https://github.com/gnocchixyz/gnocchi/pull/418 lol

PR #422, opened 2017-10-11
Pull request for jsonify-exceptions

jd:
api: convert exceptions JSON data struct rather than just a string
api: remove abort unused extra arguments

### DIFF
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -52,13 +52,13 @@ def arg_to_list(value):
     return []
 
 
-def abort(status_code, detail='', headers=None, comment=None, **kw):
+def abort(status_code, detail=''):
     """Like pecan.abort, but make sure detail is a string."""
     if status_code == 404 and not detail:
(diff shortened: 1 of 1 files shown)


gnocchi/rest/api.py
```diff
@@ -57,7 +57,7 @@ def abort(status_code, detail=''):
     if status_code == 404 and not detail:
         raise RuntimeError("http code 404 must have 'detail' set")
     if isinstance(detail, Exception):
-        detail = six.text_type(detail)
+        detail = detail.jsonify()
```
chungg: is this making the assumption that all Exception have a jsonify method defined?

jd: Yes. That's the Ultimate Goal.

If they have not, then the code will fail and explodes and make sure the gentle developer will provide a proper exception with a right `jsonify` method.

Since so far the code has been written by bad developers, none of them have this jsonify method so this patch converts all the old exception to `six.text_type`.

The Ultimate Goal as you understand is to stop doing that! and return computer parsable errors.

Issue #421, opened 2017-10-11
Add 'aggregates' API endpoint


Asu4ni:
Since in the document 'aggregates' have been distinguished from 'measures', the api should be slightly modified as well.
And maybe at the same time, due to the issue #307, the DELETE methed for 'aggregates' could be implemented as well.

Here is the todo list:
- [ ] deprecate `GET /v1/metric/<id>/measures`
- [ ] deprecate `GET /v1/batch/metrics/measures`
- [ ] add `GET /v1/metric/<id>/aggregates`
- [ ] add `GET /v1/batch/metrics/aggregates`
- [ ] modify document to mention the changes
- [ ] add release note of the additions and the deprecations

<br />

Here is the *maybe* one:
- [ ] change the key value in content returned by 'groupby metric search' from `measures` to `aggregates`
- [ ] add `DELETE /v1/metric/<id>/aggregates` to delete the processed aggregates. The range to delete can be limited by `start` and `end` parameter

<br />

If there are something missed or to be modified, please comment to let me know.

PR #427, opened 2017-10-12
aggregates: rename cross_metric to aggregates

sileht:
This change reorganizes aggregates code.

* gnocchi.aggregates becomes gnocchi.deprecated_aggregates
* gnocchi.rest.cross_metric becomes gnocchi.rest.aggregates.processor

Related #419


gnocchi/rest/api.py
```diff
@@ -31,13 +31,13 @@
...
 
-from gnocchi import aggregates
 from gnocchi import archive_policy
+from gnocchi import deprecated_aggregates
 from gnocchi import incoming
 from gnocchi import indexer
 from gnocchi import json
 from gnocchi import resource_type
-from gnocchi.rest import cross_metric
+from gnocchi.rest.aggregates import processor
```
chungg: tbh, processor is a weird name but i also don't have a better word so meh.

Asu4ni:
I am now worried that (#367  #421) since 'aggregates' now has replaced the name 'measures' for processed data, using 'aggregates for crossing_metric may cause confusion.

PR #428, opened 2017-10-13
aggregates: make aggregated() extensible

sileht:
This changes make aggregated() method more generic.

Related #419

### DIFF
--- a/gnocchi/rest/aggregates/operations.py
+++ b/gnocchi/rest/aggregates/operations.py
@@ -1,0 +1,77 @@
+# -*- encoding: utf-8 -*-
+#
+# Copyright © 2016-2017 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import numpy
+
+from gnocchi import carbonara
+
+
+AGG_MAP = {
+    'mean': numpy.nanmean,
+    'median': numpy.nanmedian,
+    'std': numpy.nanstd,
+    'min': numpy.nanmin,
+    'max': numpy.nanmax,
+    'sum': numpy.nansum,
+    'var': numpy.nanvar,
+    'count': lambda values, axis: numpy.count_nonzero(
+        ~numpy.isnan(values), axis=axis),
+}
+
+
+def handle_aggregate(agg, granularity, timestamps, values, is_aggregated):
+    values = numpy.array([AGG_MAP[agg](values, axis=1)]).T
+    if values.shape[1] != 1:
+        raise RuntimeError("Unexpected resulting aggregated array shape: %s" %
+                           values)
+    return (granularity, timestamps, values, True)
+
+
+def handle_aggregation_operator(nodes, granularity, timestamps, initial_values,
+                                is_aggregated, references):
+    op = aggregation_operators[nodes[0]]
+    agg = nodes[1]
+    subnodes = nodes[-1]
+    args = nodes[2:-1]
+    if agg not in AGG_MAP:
+        raise carbonara.UnknownAggregationMethod(agg)
+    granularity, timestamps, values, is_aggregated = evaluate(
+        subnodes, granularity, timestamps, initial_values,
+        is_aggregated, references)
+    return op(agg, granularity, timestamps, values, is_aggregated, *args)
+
+
+aggregation_operators = {
+    u"aggregate": handle_aggregate,
+}
+
+
+def evaluate(nodes, granularity, timestamps, initial_values, is_aggregated,
+             references):
+    if nodes[0] in aggregation_operators:
+        return handle_aggregation_operator(nodes, granularity, timestamps,
+                                           initial_values, is_aggregated,
+                                           references)
+    elif nodes[0] == "metric":
+        if isinstance(nodes[1], list):
+            predicat = lambda r: r in nodes[1:]
+        else:
+            predicat = lambda r: r == nodes[1:]
+        indexes = [i for i, r in enumerate(references) if predicat(r)]
+        return (granularity, timestamps, initial_values.T[indexes].T,
+                is_aggregated)
+    else:
+        raise RuntimeError("Operation node tree is malformed: %s" % nodes)
--- a/gnocchi/rest/aggregates/processor.py
+++ b/gnocchi/rest/aggregates/processor.py
@@ -22,18 +22,11 @@ import numpy
 import six
 
 from gnocchi import carbonara
+from gnocchi.rest.aggregates import operations as agg_operations
 from gnocchi import storage as gnocchi_storage
 
 
 LOG = daiquiri.getLogger(__name__)
-
-
-AGG_MAP = {'mean': numpy.nanmean,
-           'median': numpy.nanmedian,
-           'std': numpy.nanstd,
-           'min': numpy.nanmin,
-           'max': numpy.nanmax,
-           'sum': numpy.nansum}
 
 
 class UnAggregableTimeseries(Exception):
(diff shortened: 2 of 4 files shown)


gnocchi/rest/aggregates/processor.py
```diff
@@ -46,33 +39,39 @@ def __init__(self, reason):
...
         self.reason = reason
+        metrics = ("%s/%s" % (m.id, a) for (m, a) in metrics_and_aggregations)
         super(MetricUnaggregatable, self).__init__(
-            "Metrics %s can't be aggregated: %s"
-            % (", ".join((str(m.id) for m in metrics)), reason))
+            "Metrics %s can't be aggregated: %s" % (
+                ", ".join(metrics), reason))
+
+
+def _get_measures_timeserie(storage, metric, aggregation, ref_identifier,
```
jd: i guess that'll be needed later? :)

sileht:
Pull-request updated, HEAD is now 43e31bc2c856c0741a784ee98d6bf98ead52a3bc

gnocchi/rest/aggregates/operations.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+aggregation_operators = {
+    u"aggregate": handle_aggregate,
+}
+
+
+def evaluate(nodes, granularity, timestamps, initial_values, is_aggregated,
+             references):
+    if isinstance(nodes, (int, float)):
```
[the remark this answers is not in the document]
sileht: This parts should be in future patch, it's a leftover. It will be fixed.

gnocchi/rest/aggregates/processor.py
sileht: I have put explanation in the commit msg !

sileht:
> You sure you don't want to add a few tests? :)

Yes, I have explained why in the commit msg.

gnocchi/rest/aggregates/processor.py
```diff
@@ -167,16 +169,34 @@ def aggregated(timeseries, aggregation, from_timestamp=None,
...
-        if aggregation in AGG_MAP:
-            values = AGG_MAP[aggregation](values, axis=1)
-        elif aggregation == 'count':
-            values = numpy.count_nonzero(~numpy.isnan(values), axis=1)
+        try:
+            granularity, times, values, is_aggregated = (
+                agg_operations.evaluate(operations, key, times, values,
+                                        False, references[key]))
+        except agg_operations.InvalidOperations as e:
+            raise UnAggregableTimeseries("Invalid operations: %s" % e.reason)
```
jd: Food for thought: do we really need this `UnAggregableTimeseries` here or could we just use ` except InvalidOperations` directly in `api.py`?

gnocchi/rest/aggregates/processor.py
```diff
@@ -167,16 +169,34 @@ def aggregated(timeseries, aggregation, from_timestamp=None,
...
+
+        # NOTE(sileht): Check if an operations have aggregated the result.
```
jd: "operation"
no s

gnocchi/rest/aggregates/processor.py
```diff
@@ -167,16 +169,34 @@ def aggregated(timeseries, aggregation, from_timestamp=None,
...
+        # (aggregate, or binary operation between metrics, ...)
+        # Otherwise we tag each resulting serie with its original reference.
+        if len(values.shape) != 2:
+            raise UnAggregableTimeseries("Unexpected resulting array shape")
+        elif is_aggregated:
+            if values.shape[1] != 1:
+                raise UnAggregableTimeseries("Unexpected resulting "
```
jd: IIUC this kind of exception cannot be raised by a user providing wrong data. So if it happens, it's a bug. I don't think we should use `UnAggregableTImeseries` because that would be caught by the `api.py` code and returned as a 400. We probably want a 500 here and a traceback. So we need a different exception type, something like a `RuntimeError`.

gnocchi/rest/aggregates/operations.py
```diff
@@ -0,0 +1,81 @@
...
+
+aggregation_operators = {
+    u"aggregate": handle_aggregate,
+}
+
+
+def evaluate(nodes, granularity, timestamps, initial_values, is_aggregated,
+             references):
+    if nodes[0] in aggregation_operators:
+        return handle_aggregation_operator(nodes, granularity, timestamps,
```
chungg: is the weird cyclic reference handle_aggregation_operator and evaluate required?

gnocchi/rest/aggregates/processor.py
sileht: Honestly, if InvalidOperations is raised, it's a bug too, that mean the generated object built with voluptuous is malformed. I replaced it by a RuntimeError

gnocchi/rest/aggregates/processor.py
sileht: done

gnocchi/rest/aggregates/processor.py
sileht: Right, fixed.

gnocchi/rest/aggregates/processor.py
```diff
@@ -167,16 +169,34 @@ def aggregated(timeseries, aggregation, from_timestamp=None,
...
+            granularity, times, values, is_aggregated = (
+                agg_operations.evaluate(operations, key, times, values,
+                                        False, references[key]))
+        except agg_operations.InvalidOperations as e:
+            raise UnAggregableTimeseries("Invalid operations: %s" % e.reason)
+
+        # NOTE(sileht): Check if an operations have aggregated the result.
+        # (aggregate, or binary operation between metrics, ...)
+        # Otherwise we tag each resulting serie with its original reference.
+        if len(values.shape) != 2:
```
chungg: should the following code not be part of operations.evaluate()? seems weird we are verify results in a different part of code.

sileht: I have moved the verify where the issue can occurs.

gnocchi/rest/aggregates/operations.py
sileht: Using recursive method is the more readable way to browse the a tree. I'm sure rewriting that with a while loop will look ugly and over complicated.

sileht:
`are you creating metrics_and_aggregations so in theory you can take different aggregates per metric (even though this patch doesn't actually support it). any way to do this only when we actually support it via api`

I have prefered that here to only have one patchset where I change all tests of test_aggregates.py. So the next patch that adds the new API, will just adds new tests and don't touch all other tests.

gnocchi/rest/aggregates/processor.py
sileht: About already present UnAggregableTimeseries I will remove them in an separate patch.

sileht:
Pull-request updated, HEAD is now 8cad21162731e005dbce25d164319ff066a3e26e

gnocchi/rest/aggregates/operations.py
chungg: oh, it's a tree. i had no idea

gnocchi/rest/aggregates/operations.py
```diff
@@ -0,0 +1,77 @@
...
+                                is_aggregated, references):
+    op = aggregation_operators[nodes[0]]
+    agg = nodes[1]
+    subnodes = nodes[-1]
+    args = nodes[2:-1]
+    if agg not in AGG_MAP:
+        raise carbonara.UnknownAggregationMethod(agg)
+    granularity, timestamps, values, is_aggregated = evaluate(
+        subnodes, granularity, timestamps, initial_values,
+        is_aggregated, references)
```
chungg: i'm confused what this is_aggregated is for? it's always false

sileht:
>+def handle_aggregation_operator(nodes, granularity, timestamps, initial_values,
>+                                is_aggregated, references):
>+    op = aggregation_operators[nodes[0]]
>+    agg = nodes[1]
>+    subnodes = nodes[-1]
>+    args = nodes[2:-1]
>+    if agg not in AGG_MAP:
>+        raise carbonara.UnknownAggregationMethod(agg)
>+    granularity, timestamps, values, is_aggregated = evaluate(
>+        subnodes, granularity, timestamps, initial_values,
>+        is_aggregated, references)
>
>i'm confused what this is_aggregated is for? it's always false

That's for future operators. We need to known if an operation have returned
all columns aggregated or if returned columns are just the original series modified.

We can't use the number of columns for that, in case of aggregated() have been called with
only one metrics, that's why I use a boolean.


--
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

PR #431, opened 2017-10-16
rest: allow to pass resource_id in the metric creation payload

jd:


gnocchi/rest/api.py
```diff
@@ -546,19 +546,22 @@ def _lookup(self, id, *remainder):
...
+    @staticmethod
+    def MetricSchema(definition):
+        creator = pecan.request.auth_helper.get_current_user(
+            pecan.request)
+
         # First basic validation
-        definition = cls._MetricSchema(definition)
+        schema = voluptuous.Schema({
+            "archive_policy_name": six.text_type,
+            "resource_id": functools.partial(ResourceID, creator=creator),
```
chungg: this optional?

gnocchi/rest/api.py
```diff
@@ -579,12 +582,23 @@ def MetricSchema(cls, definition):
...
+                       "detail": "name",
+                       "reason": "Name cannot be null "
+                       "if resource_id is not null"})
+            original_resource_id, resource_id = resource_id
 
         enforce("create metric", {
             "creator": creator,
             "archive_policy_name": archive_policy_name,
+            "resource_id": resource_id,
+            "original_resource_id": original_resource_id,
```
chungg: we need this for policy check?

gnocchi/rest/api.py
jd: It should be but it's not marked as such so I think it won't be. Damn, we might miss a test here.

gnocchi/rest/api.py
jd: Not with the actual policy rules, but if anyone wants to police after it, yes.

gnocchi/rest/api.py
jd: Ok, late reply, I forgot how this works and you confused me with "no test for /v1/metric". We have plenty of tests everywhere that create metrics in Gabbi with `POST /v1/metric` where no resource_id is set and it works. :)

Actually, all keys in voluptuous are optional except if `required=True` is passed when validating. We never do that here. The `voluptuous.Optional` flag is just used in that case, but it's not really used in any of the code currently. We usually use it for `PATCH` or `PUT` operations which are not supported on `/v1/metric`

chungg: word. i did a bad grep.

PR #434, opened 2017-10-18
incoming: use utils.parallel_map rather than futures

jd:


gnocchi/incoming/__init__.py
```diff
@@ -124,12 +121,11 @@ def add_measures_batch(self, metrics_and_measures):
         :param metrics_and_measures: A dict where keys
         are metrics and value are measure.
         """
-        with futures.ThreadPoolExecutor(max_workers=_NUM_WORKERS) as executor:
-            list(executor.map(
-                lambda args: self._store_new_measures(*args),
-                ((metric, self._encode_measures(measures))
-                 for metric, measures
-                 in six.iteritems(metrics_and_measures))))
+        utils.parallel_map(
```
chungg: so i'm thinking about memory usage here but does incoming need coordination? because this will load coordination code.

jd: It does not. I'll move coordination out.

jd:
Pull-request updated, HEAD is now 35b278379de9a7b0ee5b5e2c49d512834f4ab0c3

PR #436, opened 2017-10-20
Allow to rename archive policies

0livd:
This supersedes #389.

Archive policies can be renamed by issuing a PATCH request containing the new name.
The definition of an ap can also be patched along with the name within the same request.
Archive policies are now identified by id but their name must still be unique.

### DIFF
--- a/gnocchi/indexer/alembic/versions/71cfe62bcf98_use_an_id_as_primary_key_for_archive_.py
+++ b/gnocchi/indexer/alembic/versions/71cfe62bcf98_use_an_id_as_primary_key_for_archive_.py
@@ -1,0 +1,147 @@
+# Copyright 2017 The Gnocchi Developers
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+#
+
+"""Use an id as primary key for archive policy
+
+Revision ID: 71cfe62bcf98
+Revises: 1e1a63d3d186
+Create Date: 2017-09-24 19:46:22.660503
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+import sqlalchemy_utils
+import uuid
+
+
+# revision identifiers, used by Alembic.
+revision = '71cfe62bcf98'
+down_revision = '1e1a63d3d186'
+branch_labels = None
+depends_on = None
+
+archive_policy_helper = sa.Table(
+    'archive_policy',
+    sa.MetaData(),
+    sa.Column('id', sqlalchemy_utils.types.uuid.UUIDType(binary=True),
+              primary_key=True),
+    sa.Column('name', sa.String(length=255), nullable=False, unique=True),
+)
+
+metric_helper = sa.Table(
+    'metric',
+    sa.MetaData(),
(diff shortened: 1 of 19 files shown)


gnocchi/indexer/alembic/versions/71cfe62bcf98_use_an_id_as_primary_key_for_archive_.py
```diff
@@ -0,0 +1,144 @@
...
+    if connection and connection.engine.name == "postgresql":
+        op.drop_constraint('archive_policy_pkey', 'archive_policy',
+                           type_='primary')
+        op.create_primary_key('archive_policy_pkey', 'archive_policy', ['id'])
+    op.create_foreign_key(
+        'fk_metric_ap_id_ap_id',
+        'metric',
+        'archive_policy',
+        ['archive_policy_id'],
+        ['id'])
```
sileht: ondelete='RESTRICT' is missing

gnocchi/indexer/alembic/versions/71cfe62bcf98_use_an_id_as_primary_key_for_archive_.py
```diff
@@ -0,0 +1,144 @@
...
+    op.create_foreign_key(
+        'fk_apr_ap_id_ap_id',
+        'archive_policy_rule',
+        'archive_policy',
+        ['archive_policy_id'],
+        ['id'])
```
sileht: Same here.

gnocchi/indexer/sqlalchemy_base.py
```diff
@@ -150,7 +150,9 @@ def process_result_value(self, value, dialect):
 class ArchivePolicy(Base, GnocchiBase, archive_policy.ArchivePolicy):
     __tablename__ = 'archive_policy'
 
-    name = sqlalchemy.Column(sqlalchemy.String(255), primary_key=True)
+    id = sqlalchemy.Column(sqlalchemy_utils.UUIDType(), primary_key=True)
+    name = sqlalchemy.Column(sqlalchemy.String(255), unique=True,
```
sileht: You need to create a index on name with __table_args__.

gnocchi/indexer/alembic/versions/71cfe62bcf98_use_an_id_as_primary_key_for_archive_.py
```diff
@@ -0,0 +1,144 @@
...
+    op.create_index('ix_ap_name', 'archive_policy', ['name'], unique=True)
```
sileht: According the failing postgresql tests, unique=True seems doesn't work.

0livd:
@sileht Thx for the review, tests are passing now :)

gnocchi/tests/functional/gabbits/base.yaml
```diff
@@ -61,14 +61,14 @@ tests:
   GET: $LOCATION
   status: 200
   response_json_paths:
-      $.archive_policy.name: test1
+      $.archive_policy_name: test1
```
jd: This changes the API so that's not good. Is there any reason you can't fetch the archive policy details with the metric and return it like before?

0livd: Some tests use ``archive_policy_name`` and others ``archive_policy.name`` but I removed the conditional block that chooses between both of them and arbitrarily chose to only keep archive_policy_name  https://github.com/gnocchixyz/gnocchi/pull/436/files#diff-9efde1e460b5fbbfdced1a5cd0b95814L205 . I first tried to replace it with the following block but with the other modifications I made, it is now always hitting the ``else``. I didn't find any way to keep the behavior of the previous conditional block... if you have any clue on that.

    if 'archive_policy' in unloaded:
       d['archive_policy_name'] = self.archive_policy.name
    else:
        d['archive_policy'] = self.archive_policy

jd:
This has not been updated in a while and needs rebasing. I'll close this PR as we want to leave the archive policy in the long run anyway.

PR #437, opened 2017-10-20
aggregates: search-and-fetch api

sileht:
This change adds the /v1/aggregates/search-and-fetch API.

It allows to search for metrics with the same format as
resource search and cross metric aggregation.

And then it applies operations on them like /v1/aggregates/fetch API.

It also re-add transformation tests related to cross metric aggregation
endpoint but with new aggregates API.

Related #419

### DIFF
--- a/doc/source/rest.j2
+++ b/doc/source/rest.j2
@@ -602,7 +602,16 @@ least one of the time series is returned.
 
 {{ scenarios['get-aggregates-by-metric-ids-fill']['doc'] }}
 
+It's also possible to do that aggregation on |metrics| linked to |resources|.
+In order to select these |resources|, the following endpoint accepts a query
+such as the one described in `Searching for resources`_.
 
+{{ scenarios['get-aggregates-by-attributes-lookup']['doc'] }}
+
+It is possible to group the |resource| search results by any attribute of the
+requested |resource| type, and then compute the aggregation:
+
+{{ scenarios['get-aggregates-by-attributes-lookup-groupby']['doc'] }}
 
 List of supported <operations>
 ------------------------------
--- a/gnocchi/rest/aggregates/api.py
+++ b/gnocchi/rest/aggregates/api.py
@@ -14,12 +14,15 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import itertools
+
 import pecan
 from pecan import rest
 import pyparsing
 import six
 import voluptuous
 
+from gnocchi import indexer
 from gnocchi.rest.aggregates import exceptions
 from gnocchi.rest.aggregates import operations as agg_operations
 from gnocchi.rest.aggregates import processor
@@ -155,15 +158,28 @@ def get_measures_or_abort(metrics_and_aggregations, operations, start,
                         "detail": [(str(e.metric.id), e.method)]})
 
 
-class FetchController(rest.RestController):
+def ResourceTypeSchema(resource_type):
+    try:
+        pecan.request.indexer.get_resource_type(resource_type)
+    except indexer.NoSuchResourceType as e:
+        api.abort(400, e)
+    return resource_type
 
-    FetchSchema = {
+
+class AggregatesController(rest.RestController):
+
+    FetchSchema = voluptuous.Any({
         "operations": OperationsSchema
-    }
+    }, {
+        "operations": OperationsSchema,
+        "resource_type": ResourceTypeSchema,
+        "search": voluptuous.Any(api.ResourceSearchSchema,
+                                 api.QueryStringSearchAttrFilter.parse),
+    })
 
     @pecan.expose("json")
     def post(self, start=None, stop=None, granularity=None,
-             needed_overlap=100.0, fill=None):
+             needed_overlap=100.0, fill=None, groupby=None):
         start, stop, granularity, needed_overlap, fill = api.validate_qs(
             start, stop, granularity, needed_overlap, fill)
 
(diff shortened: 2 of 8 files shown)


sileht:
Pull-request updated, HEAD is now 02b078baafdf0de80d59c137053a5ada1e7f0c26

doc/source/rest.j2
```diff
@@ -602,7 +602,16 @@ least one of the time series is returned.
 
 {{ scenarios['get-aggregates-by-metric-ids-fill']['doc'] }}
 
+It's also possible to do that aggregation on |metrics| linked to |resources|.
+In order to select these |resources|, the following endpoint accepts a query
+such as the one described in `Searching for resources`_.
```
jd: I'm thinking we'll have to rewrite this as a string like we did for operations, e.g., `(eq "server_group" "my_autoscaling_group")`

gnocchi/rest/aggregates/api.py
```diff
@@ -207,5 +210,94 @@ def post(self, start=None, stop=None, granularity=None,
...
+    @pecan.expose('json')
+    def post(self, start=None, stop=None, granularity=None,
+             needed_overlap=100.0, fill=None, groupby=None):
+        start, stop, granularity, needed_overlap, fill = api.validate_qs(
+            start, stop, granularity, needed_overlap, fill)
+        body = api.deserialize_and_validate(self.SearchAndFetchSchema)
+
+        references = list(extract_references(body["operations"]))
+        if not references:
+            api.abort(400, {"cause": "operations is invalid",
```
jd: `Operation is invalid`

gnocchi/rest/aggregates/api.py
```diff
@@ -207,5 +210,94 @@ def post(self, start=None, stop=None, granularity=None,
...
+                            "reason": "at least one 'metric' is required",
```
jd: `At least…`

sileht:
Pull-request updated, HEAD is now 21a91a0c2bbe8705f619d15b7ad6733c6ef8f4a6

doc/source/rest.j2
sileht: That looks a good idea !

sileht:
Pull-request updated, HEAD is now ad345bb01b57339e935c511c308b2c49c4bdaa06

sileht:
Pull-request updated, HEAD is now 21a91a0c2bbe8705f619d15b7ad6733c6ef8f4a6

gnocchi/rest/aggregates/api.py
```diff
@@ -207,5 +210,94 @@ def post(self, start=None, stop=None, granularity=None,
...
+                                     start, stop, granularity, needed_overlap,
+                                     fill, ref_identifier="name")
+
+
 class AggregatesController(object):
-    fetch = FetchController()
+    def __init__(self):
+        ctrls = {
+            "fetch": FetchController(),
+            "search-and-fetch": SearchAndFetchController(),
```
chungg: tbh, new endpoint seems a little strange. should we maybe just combine the two?

gnocchi/tests/functional/gabbits/aggregates-search-and-fetch.yaml
```diff
@@ -0,0 +1,190 @@
...
+        id: b1409ec6-3909-4b37-bbff-f9a5448fe328
+        user_id: 70b5b732-9d81-4dfb-a8a1-a424ef3eae6b
+        project_id: ee4cfc41-1cdc-4d2f-9a08-f94111d80171
+        metrics:
+          cpu.util:
+            archive_policy_name: low
+      status: 201
+
+    - name: post cpuutil measures 4
+      POST: /v1/resource/generic/33333BC5-5948-4F29-B7DF-7DE607660452/metric/cpu.util/measures
```
chungg: this is posting to resource 3. resource 4 doesn't seem to be used ever.

gnocchi/rest/aggregates/api.py
chungg: maybe we should call it transform, translate, or mutate instead of fetch now that i think about it.

sileht:
>>  class AggregatesController(object):
>-    fetch = FetchController()
>+    def __init__(self):
>+        ctrls = {
>+            "fetch": FetchController(),
>+            "search-and-fetch": SearchAndFetchController(),
>
>tbh, new endpoint seems a little strange. should we maybe just combine the two?

I guess we can check is "search" is set or not to known if we have to
query resources.

@gordc, @jdanjou WDT ?

-- 
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

jd:
@sileht yeah sounds like a good idea to me. I don't see any downside to it.

chungg:
@sileht yeah, that seems better if possible.

sileht:
Pull-request updated, HEAD is now dc05a22af99d4089cb10eb63276f81bd3000f2f1

sileht:
Pull-request updated, HEAD is now 02f66294a0c97caeb8c7c067f51ef9ff6e091aa5

sileht:
Pull-request updated, HEAD is now 0bccb18578f3acea3ce18c677419048e20015f81

sileht:
Pull-request updated, HEAD is now 228cedfbacf865298300abdc8b5b6c4760e1f5c6

sileht:
Pull-request updated, HEAD is now a59573354a91e7a3bf5bcc5d43320f8d2783f998

sileht:
Pull-request updated, HEAD is now 8e254cd66c66403aad45d8401c1ff275da695040

gnocchi/tests/functional/gabbits/aggregates-with-resources.yaml
```diff
@@ -0,0 +1,190 @@
...
+        id: b1409ec6-3909-4b37-bbff-f9a5448fe328
+        user_id: 70b5b732-9d81-4dfb-a8a1-a424ef3eae6b
+        project_id: ee4cfc41-1cdc-4d2f-9a08-f94111d80171
+        metrics:
+          cpu.util:
+            archive_policy_name: low
+      status: 201
+
+    - name: post cpuutil measures 4
+      POST: /v1/resource/generic/33333BC5-5948-4F29-B7DF-7DE607660452/metric/cpu.util/measures
```
chungg: still posting to resource 3 and not resource 4

sileht:
Pull-request updated, HEAD is now 32a61aa3dc35a0aa9cfc1c1bd6c0f611565edace

gnocchi/tests/functional/gabbits/aggregates-with-resources.yaml
sileht: lol, the test that use it is also true when no measure are inside metric4

chungg: ooohhh. metric4 is there to prove it's not being picked up by search query. just noticed it was a different user_id

Asu4ni:
I was browsing through the release note 4.1, and found that the deprecation of Aggregation across metrics endpoint was not mentioned there. Is it okay if there isn't one?

jd:
It's marked as deprecated in the documentation of the REST API. We don't have any programmatic way to deprecate things, so we're doing our best. If you want to add a release note, feel free, we can backport it.

PR #438, opened 2017-10-20
aggregtated: remove useless resample argument

sileht:
"resample" can be done within the "operations".

This change does that.


sileht:
Pull-request updated, HEAD is now 1a78074bcb2415d34bd2ccad16e073dad1686932

sileht:
Pull-request updated, HEAD is now d22d26a9d32565d1a62274c874a058da68b94a9c

sileht:
On Mon, Oct 23, 2017 at 04:54:46AM -0700, gord chung wrote:
>should we deprecate support for resample querystring in api?

No, the new aggregates API doesn't have resample in querystring.

This changes just change internal of cross metrics aggregation not the
API.

-- 
Mehdi Abaakouk
mail: sileht@sileht.net
irc: sileht

PR #440, opened 2017-10-23
simplify operations interface

chungg:
it's very cluttered to pass around is_aggregated when it's only
ever used for end result formatting.

we'll only ever have two scenarios:
- everything is aggregated into a single result
- batch retrieval and return all series individually (maybe modified)

there doesn't seem to be path which has a partially aggregated
result. if there is, it doesn't seem like this code would even handle
it.

this patch:
- removes is_aggregated and just assumes, if there's one series
returned and there was more than one given, it's aggregated.
- just transpose once if multiple series returned.

### DIFF
--- a/gnocchi/rest/aggregates/processor.py
+++ b/gnocchi/rest/aggregates/processor.py
@@ -156,21 +156,22 @@ def aggregated(refs_and_timeseries, operations, from_timestamp=None,
                     'timespan (%.2f%%)' % (needed_percent_of_overlap,
                                            percent_of_overlap))
 
-        granularity, times, values, is_aggregated = (
+        granularity, times, values = (
             agg_operations.evaluate(operations, key, times, values,
-                                    False, references[key]))
+                                    references[key]))
 
-        if is_aggregated:
+        values = values.T
+        if len(values) == 1 and len(references[key]) > 1:
             result["aggregated"]["timestamps"].extend(times)
             result["aggregated"]['granularity'].extend([granularity] *
                                                        len(times))
-            result["aggregated"]['values'].extend(values.T[0])
+            result["aggregated"]['values'].extend(values[0])
         else:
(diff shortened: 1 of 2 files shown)


gnocchi/rest/aggregates/processor.py
```diff
@@ -156,21 +156,22 @@ def aggregated(refs_and_timeseries, operations, from_timestamp=None,
...
 
-        granularity, times, values, is_aggregated = (
+        granularity, times, values = (
             agg_operations.evaluate(operations, key, times, values,
-                                    False, references[key]))
+                                    references[key]))
 
-        if is_aggregated:
+        values = values.T
+        if len(values) == 1 and len(references[key]) > 1:
```
sileht: If I just do: "(* 2 (metric id1 mean))" that doesn't aggregated anything, but with your change I got "aggregated": .... instead of "id1_mean" as key.

sileht: I have added some tests to cover this: https://github.com/gnocchixyz/gnocchi/pull/442

chungg: no, it'll give id1_mean. it'll only give "aggregated" if the number of original metrics was greater than 1 and shrinks to 1. i know this because you added tests for this already. :)

this case doesn't cover a scenario where you only aggregate a few of the series...  that said, i don't think the code supports that.

chungg: ah, nevermind. i see your new test for reaggregate a single series returns aggregated

Asu4ni:
I'm thinking about another way around. Checking if it works #447

Asu4ni:
@chungg I guess #447 might work.(except some strange test errors I cannot figure them out)

chungg:
closing. fix it later if needed.

PR #445, opened 2017-10-24
aggregates: remove useless checks

sileht:
It's not necessary useful to forbid fill without granularity.

Of course, some "operations" doesn't make sense. Like resample without
granularity. But it's up to the user to understand that.


sileht:
Pull-request updated, HEAD is now cb7190f6ad506fc81ae095d65702029bec333f87

PR #447, opened 2017-10-25
[Testing] Avoid passing `is_aggreagated` in and out in rest/aggregates/operations

Asu4ni:
test if it works
Remove is_aggreagated from function in rest/aggregates/operation

### DIFF
--- a/gnocchi/rest/aggregates/operations.py
+++ b/gnocchi/rest/aggregates/operations.py
@@ -100,23 +100,23 @@ unary_operators = {
 
 
 def handle_unary_operator(nodes, granularity, timestamps, initial_values,
-                          is_aggregated, references):
+                          references):
     op = nodes[0]
     granularity, timestamps, values, is_aggregated = evaluate(
         nodes[1], granularity, timestamps, initial_values,
-        is_aggregated, references)
+        references)
 
     values = unary_operators[op](values)
     return granularity, timestamps, values, is_aggregated
 
 
 def handle_binary_operator(nodes, granularity, timestamps,
-                           initial_values, is_aggregated, references):
+                           initial_values, references):
     op = nodes[0]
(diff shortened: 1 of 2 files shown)


gnocchi/rest/aggregates/operations.py
```diff
@@ -183,20 +181,22 @@ def handle_resample(agg, granularity, timestamps, values, is_aggregated,
...
     subnodes = nodes[-1]
     args = nodes[2:-1]
     granularity, timestamps, values, is_aggregated = evaluate(
         subnodes, granularity, timestamps, initial_values,
-        is_aggregated, references)
-    return op(agg, granularity, timestamps, values, is_aggregated,
-              references, *args)
+        references)
+    if op == handle_aggregate:
+        is_aggregated = True
```
sileht: I preferred the version where each methods return False or True.

Asu4ni: Do you mean that the `handle_*` functions still has to be called with `is_aggregated`? I modified this part since only `handle_aggregate` returns `True` all the time and the others just return back `is_aggregated` without any modification.

Asu4ni:
@sileht sorry about dismising the review. I was modifying the commit message to make it more clear. And  if you strongly prefer the `handle_*` functions remain the same as before, just tell me and I'll change it back.

gnocchi/rest/aggregates/operations.py
chungg: yeah, i'm not sure this patch cleans up the interface any better. it's probably more confusing  because there's just random false/true parameters passed around.

we could remove it completely if 'operations' was a string... that way we just have to check if there's an aggregate operation or if the number of result series < number of input series.

jd:
Closing since it does not seem like we want to change that.

PR #454, opened 2017-10-27
make metric rest docs more consumable

chungg:
- divide metric endpoint into crud sections


doc/source/rest.j2
```diff
@@ -187,7 +222,7 @@ By default, new |measures| can only be processed if they have timestamps in the
 future or part of the last aggregation period. The last aggregation period size
 is based on the largest |granularity| defined in the |archive policy|
 definition. To allow processing |measures| that are older than the period, the
-`back_window` parameter can be used to set the number of coarsest periods to
+`back_window` parameter can be used to set the number of historical periods to
```
jd: Except you don't need the number of historical period.
We need to change that unit anyway. Logging a big.

chungg: i'm not entirely sure what you mean by

> We need to change that unit anyway. Logging a big.

you want me to change something in this patch?

jd: s/big/bug/

I opened #459 to track that. I don't think your change here makes sense.
That does not mean I think the original text is easy to understand. But this line change remove information.
So it makes from *hard* to *impossible* to understand the parameter. :)

chungg: done

doc/source/rest.j2
```diff
@@ -119,24 +162,25 @@ timestamp:
 
 {{ scenarios['get-measures-from']['doc'] }}
 
+Aggregate
```
Asu4ni: Maybe `Aggregation` is better. not to be confused with 'aggregates'. and the parameter in url uses `aggregation` as well

doc/source/rest.j2
```diff
@@ -108,9 +149,11 @@ to be handled:
    Depending on the amount of data that is unprocessed, `refresh` may add
    some overhead to your query.
 
-The list of points returned is composed of tuples with (timestamp,
-|granularity|, value) sorted by timestamp. The |granularity| is the |timespan|
-covered by aggregation for this point.
+Filter
```
Asu4ni: I consider `Filter` should be the subsection of `Read` since it is used when reading aggregates

doc/source/rest.j2
```diff
@@ -87,18 +101,45 @@ status code. It is possible to provide any number of |measures|.
...
+
+Or using named |metrics| of |resources|:
+
+{{ scenarios['post-measures-batch-named']['doc'] }}
+
+If some named |metrics| specified in the batch request do not exist, Gnocchi
+can try to create them as long as an |archive policy| rule matches:
+
+{{ scenarios['post-measures-batch-named-create']['doc'] }}
+
```
Asu4ni: Since that measures and aggregates have been distinguished now, maybe there should be a new section:
```
Aggregates
==========
```

doc/source/rest.j2
```diff
@@ -532,112 +663,71 @@ into by using the `metric_id` query parameter several times.
...
-argument, and in this case the second argument passed is the value, or it.
-
-The operators or (`or` or `∨`), and (`and` or `∧`) and `not` are also
-supported, and take a list of arguments as parameters.
 
 .. _aggregates:
 
-Aggregates: on the fly, measurements modification and aggregation
-===================================================================
+Aggregates
```
Asu4ni: `on the fly` should be retained. Otherwise, it'll be confused with the stored `Aggregates` above.

doc/source/rest.j2
Asu4ni: Since that measures and aggregates have been distinguished now, maybe there should be a new section:
```
Aggregates
==========
```

doc/source/rest.j2
Asu4ni: I consider `Filter` should be the subsection of `Read` since it is used when reading aggregates

doc/source/rest.j2
Asu4ni: Maybe `Aggregation` is better. not to be confused with 'aggregates'. and the parameter in url uses `aggregation` as well

doc/source/rest.j2
Asu4ni: `on the fly` should be retained. Otherwise, it'll be confused with the stored `Aggregates` above.

doc/source/rest.j2
chungg: sorry, i don't understand your comment. i already have an aggregates section below?

doc/source/rest.j2
chungg: yeah, i considered this. i really wanted filter to be top level but ok.

doc/source/rest.j2
chungg: sure.

doc/source/rest.j2
chungg: how about dynamic aggregates? 'on the fly' is more slang i feel.

doc/source/rest.j2
Asu4ni: At some point before, the incoming data points and the processed(aggregated) ones are both called `measures`. But since one of my commit and the discussion with jd, the processed ones are now called `aggregates`. To explain in another way, now we have 'measure storage', 'aggregate storage', 'index storage', and the `Aggregates` I mentioned is the one in 'aggregate' storage.
To make the document more consistent, measures can only be pushed(POST), and the ones get read(GET) are actually `aggregates`. So I hope that the `Read` part here should belong to `Aggregates` section rather than `Measures` section.
BTW, I know that for the endpoint, `measures` is still used when getting aggregates, but I'm planning to deprecate that and create a new `aggregates` endpoint. #421

chungg: so i understand the comment about terminology but i would probably have a discussion first about moving all GET actions to another endpoints. imo, it's a bad idea. people are use to CRUD concept and gnocch itself is already built on principle that everything is aggregated.

to me, aggregates endpoint is a custom-retrieval endpoint, not a simple read.

jd: To circle back on this discussion, I lean toward @Asu4ni about terminology because I encountered a lot of people confused about measures vs. aggregates and having it clearer is better. The changes proposed in #421 are orthogonal to this though – they could happen one day, but for now we should just focus on what exists. :)

That being said, that seems good enough for me now. The doc really talks about retrieving *aggregates* and not *measures*. :)

PR #469, opened 2017-11-08
doc: add some note about resample

sileht:
People should leverage archive policy system and using
resample at last chance.

This change adds some notes about that.

### DIFF
--- a/doc/source/rest.j2
+++ b/doc/source/rest.j2
@@ -187,6 +187,13 @@ can be resampled to a new |granularity|.
 
 {{ scenarios['get-measures-resample']['doc'] }}
 
+
+.. note::
+
+   If you plan to execute the query often, it is recommended for performance
+   to leverage an |archive policy| with the needed |granularity| instead of
+   resampling the time series on each query.
+
 .. note::
 
    Depending on the |aggregation method| and frequency of |measures|, resampled
@@ -735,6 +742,13 @@ Resample
                        (mean, median, std, min, max, sum, var, count)
 
    granularity: the granularity (e.g.: 1d, 60s, ...)
+
+.. note::
+
+   If you plan to execute the query often, it is recommended for performance
+   to leverage an |archive policy| with the needed |granularity| instead of
+   resampling the time series on each query.
+
 
 Math operations
 ~~~~~~~~~~~~~~~


doc/source/rest.j2
```diff
@@ -187,6 +187,13 @@ can be resampled to a new |granularity|.
 
 {{ scenarios['get-measures-resample']['doc'] }}
 
+
+`.. note::
+
+   Resample should be avoid because it have a performance impact on each
+   queries. It strongly recommended to define `archive policies` with all
```
jd: It is strongly…

you meant `|archive policies|`

doc/source/rest.j2
```diff
@@ -187,6 +187,13 @@ can be resampled to a new |granularity|.
...
+   `granularities` you will need at query times.
```
jd: |granularities| ?

doc/source/rest.j2
```diff
@@ -187,6 +187,13 @@ can be resampled to a new |granularity|.
 
 {{ scenarios['get-measures-resample']['doc'] }}
 
+
+`.. note::
+
+   Resample should be avoid because it have a performance impact on each
```
jd: it has

sileht:
Pull-request updated, HEAD is now 6b190219100099c3d2da13e13b74890464d2955b

sileht:
Pull-request updated, HEAD is now 793a2025dc058ab6581b08f2a0635435cf484802

doc/source/rest.j2
```diff
@@ -187,6 +187,13 @@ can be resampled to a new |granularity|.
 
 {{ scenarios['get-measures-resample']['doc'] }}
 
+
+.. note::
+
+   Resample should be avoided because it has a performance impact on each
+   queries. It is strongly recommended to define |archive policy| with all
```
jd: each query
it's singular

and it's "an archive policy" or "archive policies"

doc/source/rest.j2
```diff
@@ -187,6 +187,13 @@ can be resampled to a new |granularity|.
...
+   |granularities| you will need at query times.
```
jd: time

chungg:
i'm on the fence. understandly, it will add a performance impact and may return results that aren't exactly accurate. that said, i don't know if we want to encourage policies that compute granularities for rare queries.

i'm pretty sure any granularity above a day will have a negative performance affect on metricd as it now needs to store/retreive a (really) big back_window and resulting computation. for example, a monthly time-series will probably only be queried very rarely but refreshing the the monthly timeseries on every datapoint is a lot of work.

sileht:
Pull-request updated, HEAD is now 6d08a628cf78954aa40611bbdb75e1a9b2040701

sileht:
@gordc I have reworked the wording to be less strong.

doc/source/rest.j2
```diff
@@ -187,6 +187,13 @@ can be resampled to a new |granularity|.
 
 {{ scenarios['get-measures-resample']['doc'] }}
 
+
+.. note::
+
+   If the quey is planned to be done often, for better performance, it is
```
jd: If you plan to execute the query often, it is recommended for performances to leverage an |archive policy| with the needed |granularity| instead of resampling the time series on each query.

sileht:
Pull-request updated, HEAD is now 2caa726774e76c65af079264d7472f03c24c9d0b

doc/source/rest.j2
```diff
@@ -187,6 +187,13 @@ can be resampled to a new |granularity|.
 
 {{ scenarios['get-measures-resample']['doc'] }}
 
+
+.. note::
+
+   If you plan to execute the query often, it is recommended for performances
```
chungg: s/performances/performance/

doc/source/rest.j2
```diff
@@ -736,6 +743,13 @@ Resample
 
    granularity: the granularity (e.g.: 1d, 60s, ...)
 
+.. note::
+
+   If you plan to execute the query often, it is recommended for performances
```
chungg: s/performances/performance/

sileht: done

doc/source/rest.j2
sileht: done

sileht:
Pull-request updated, HEAD is now 2aec76230d30d5c236c67924c6fb4f2e9a9303d0

Issue #465, opened 2017-11-03
Add the possibility to get TOP N groups when aggregating metrics


scolinas:
When the selected grouping has many results it will be usefull to get top N and others alltogether aggregated.

chungg:
can you provide an example of what you mean?

xuyao18:
wow!  @scolinas   I have same requirement  like you .

I wrote some code using aggregation to implement this ,but not efficient. 

I think we should add some API like this :

method: `POST`
url: `/v1/top?metric_name=cpu_util&limit=10&aggregation=last&start=201701108T02:00:00`

body:
```
{
     "=": {"project_id": "project_1"}
}
```

then , the result should be like this:
```
{
     [
        {"instance_uuid1": 
              {} # some information about resource
         "cpu_util": {}  # some information about metric
         "value": "1.5"     #value to display?
        }
....
      ]
}
```

sileht:
The rule for ordering metrics have a good chance to be different depending of what the metric represent (e.g: a cost, cpu_util, a rate), and also if the meter is a gauge, a delta, ....

The API already allows to group by resource attributes retrieved with aggregates API.

For example:

> I want the top N instances for `project_1` that consume the most cpu this week.

Request:
```
POST /v1/aggregates?start=20171108T02:00:00&stop=20171115T02:00:00&group_by=original_resource_id
{
  resource_type: instance,
  search: {"=": {"project_id": "project_1"}}
  operations: "(resample 1w mean (metric cpu_util mean))"
}
```

Answer:
```
[
{"group": { "original_resource_id": "<id1>" },
 "measures": { "cpu_util_mean": [["2017-11-15T02:00:00", 1w, 58.4  ]]}}
{"group": { "original_resource_id": "<id2>" },
 "measures": { "cpu_util_mean": [["2017-11-15T02:00:00", 1w, 12.5  ]]}}
{"group": { "original_resource_id": "<id2>" },
 "measures": { "cpu_util_mean": [["2017-11-15T02:00:00", 1w, 1.1  ]]}}
{"group": { "original_resource_id": "<id3>" },
 "measures": { "cpu_util_mean": [["2017-11-15T02:00:00", 1w, 42.42  ]]}}
]
```

Note that `start - stop` is exactly one week and we resample series to 1w too. This will ensure, each resulting series have only one point.

Now we have this list, it's easy to order it as you like and pick to top N.

[...]

PR #471, opened 2017-11-08
doc: introduces a REST use-cases section

sileht:
We should list all REST API use-case that people is asking for, to
not get issue opened for new feature when it's already possible to get
the data.

It's not always clear what API call I need to do to get all informations
I need for a use-cases.

This change starts a new section and add one example.

Related #465.

### DIFF
--- a/doc/source/rest.j2
+++ b/doc/source/rest.j2
@@ -852,6 +852,17 @@ Operations between metrics can also be done, such as:
 
 {{ scenarios['get-aggregates-between-metrics']['doc'] }}
 
+List the top N resources that consume the most CPU during the last hour
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following is configured so that `stop` - `start` = `granularity` will get
+only one point per instance.
+
+This will give all information needed to order by `cpu.util` timeseries which
+can be filtered down to N results.
+
+
+{{ scenarios['use-case1-top-cpuutil-per-instances']['doc'] }}
 
 Aggregation across metrics (deprecated)
 =======================================
@@ -944,7 +955,6 @@ reporting values such as the number of new |measures| to process for each
 |metric|:
 
 {{ scenarios['get-status']['doc'] }}
-
 
 Timestamp format
 ================
--- a/doc/source/rest.yaml
+++ b/doc/source/rest.yaml
@@ -8,6 +8,10 @@
       "back_window": 0,
       "definition": [
         {
+          "granularity": "1h",
+          "timespan": "7 day"
+        },
+        {
           "granularity": "1s",
           "timespan": "1 hour"
         },
@@ -52,6 +56,10 @@
 
     {
       "definition": [
+        {
+          "granularity": "1h",
+          "timespan": "7 day"
+        },
         {
           "granularity": "1s",
           "timespan": "1 hour"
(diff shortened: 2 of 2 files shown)


doc/source/rest.j2
```diff
@@ -931,6 +931,22 @@ reporting values such as the number of new |measures| to process for each
 
 {{ scenarios['get-status']['doc'] }}
 
+Use-cases examples
+==================
+
+1. I want the top N instances for project_id "BD3A1E52-1C62-44CB-BF04-660BD88CD74D" that consume the most cpu last hour.
```
chungg: s/I want/List/

doc/source/rest.j2
```diff
@@ -931,6 +931,22 @@ reporting values such as the number of new |measures| to process for each
 
 {{ scenarios['get-status']['doc'] }}
 
+Use-cases examples
```
chungg: s/Use-cases/Use case/

doc/source/rest.j2
```diff
@@ -931,6 +931,22 @@ reporting values such as the number of new |measures| to process for each
...
+==================
+
+1. I want the top N instances for project_id "BD3A1E52-1C62-44CB-BF04-660BD88CD74D" that consume the most cpu last hour.
+
+To do so, I will use the :ref:`Dynamic aggregates API <aggregates>`. I will use
+|start|, |stop|, |granularity| and |groupby| parameters. The entire query will
+looke like:
```
chungg: i imagine everything will probably go through aggregates endpoint so maybe drop this intro.

doc/source/rest.j2
```diff
@@ -931,6 +931,22 @@ reporting values such as the number of new |measures| to process for each
...
+
+{{ scenarios['use-case1-top-cpuutil-per-instances']['doc'] }}
+
+I have ensured that |stop| - |start| = |granularity| to get only one point per
```
chungg: The following is configured so that |stop| - |start| = |granularity| will get only one point per instance.

doc/source/rest.j2
```diff
@@ -931,6 +931,22 @@ reporting values such as the number of new |measures| to process for each
...
+instances.
+
+This will give all informations we need to order by cpu.util timeseries and
+pick only the N we are interrested in.
```
chungg: This will give all information needed to order by cpu.util timeseries which can be filtered down to N results.

sileht:
Pull-request updated, HEAD is now 6ac38f8e458237abf2ca1df6beb22b5415a1f96e

doc/source/rest.j2
```diff
@@ -838,6 +838,17 @@ Operations between metrics can also be done, such as:
...
+List top N instances for that consume the most cpu last hour
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+{{ scenarios['use-case1-top-cpuutil-per-instances']['doc'] }}
+
+The following is configured so that `stop` - `start` = `granularity` will get
+only one point per instance.
+
+This will give all information needed to order by cpu.util timeseries which can
+be filtered down to N results..
```
chungg: s/.././

we can maybe move the example REST below the text but it's not important.

sileht:
Pull-request updated, HEAD is now 7d5755d7ebc9473aa0be388e07d5e78e69a46457

doc/source/rest.j2
```diff
@@ -838,6 +838,17 @@ Operations between metrics can also be done, such as:
 
 {{ scenarios['get-aggregates-between-metrics']['doc'] }}
 
+List top N instances for that consume the most cpu last hour
```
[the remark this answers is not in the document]
chungg: s/during last/during the last/

doc/source/rest.j2
```diff
@@ -838,6 +838,17 @@ Operations between metrics can also be done, such as:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following is configured so that `stop` - `start` = `granularity` will get
+only one point per instance.
+
+This will give all information needed to order by cpu.util timeseries which can
```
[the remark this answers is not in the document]
chungg: with or without 'the' works. i guess with 'the' is more correct.

sileht:
Pull-request updated, HEAD is now 380e355fb199ac38b70aba08be8a552943e005c1

PR #473, opened 2017-11-09
tests: Assert on NAN

sileht:
gabbi allows to write custom value checker since 1.37.0.

This change uses that to check that .NAN is NAN.


sileht:
Pull-request updated, HEAD is now 2587342bdaf9f43af24cac7b5b9a2e00913324cc

PR #476, opened 2017-11-10
WIP: aggregates: showup API design bug

sileht:
The added tests will regenerate only one timeserie

$.cpu.util_mean:
  MEASURES

while it should dump two timeseries...


sileht:
Pull-request updated, HEAD is now 4e2e83b28da348d1368d699a6170d15360a6477e

sileht:
Pull-request updated, HEAD is now 6214311393245eb9d38a182f18b0bb874b221dcb

Issue #477, opened 2017-11-10
aggregates API output format is not good enough


sileht:
We currently dump `{"<metric_id/name>_<aggregation>": measures}` or `{"aggregated": measures}`.

This design have some issue when two timeseries have the same names, only the last one is dumped...

Also the internal retrieves the metrics details and resource details, it would be a good enhance the output to provide them, maybe something like :

`{"reference": {"aggregation": "<aggregation>", "metric": <metric_detail>}, "measures": measures}`

sileht:
It's done

PR #479, opened 2017-11-13
aggregates: rework API output

sileht:
aggregates: rework API output
This design have some issue when two timeseries have the same names,
only the last one is dumped. Or if the same metric is asked
twice for whatever reason, measures have all timestamps duplicated.

This change the result to a new format:

For resources:
{
  "measures": { resource_id: {metric_name: {aggregation: MEASURES}}},
  "references": [RESOURCES_DETAILS]
}

For metrics
{
  "measures": {metric_id: {aggregation: MEASURES}},
  "references": [METRICS_DETAILS]
}

When the result is aggregtated we got:

For resources:
{
  "measures": { "aggregated": MEASURES},
  "references": [RESOURCES_DETAILS]
}
For metrics
{
  "measures": { "aggregated": MEASURES},
  "references": [METRICS_DETAILS]
}

Closes-bug: #477

### DIFF
--- a/gnocchi/rest/aggregates/api.py
+++ b/gnocchi/rest/aggregates/api.py
@@ -189,7 +189,9 @@ class AggregatesController(rest.RestController):
 
     @pecan.expose("json")
     def post(self, start=None, stop=None, granularity=None,
-             needed_overlap=None, fill=None, groupby=None):
+             needed_overlap=None, fill=None, groupby=None, **kwargs):
+        details = api.get_details(kwargs)
+
         if fill is None and needed_overlap is None:
             fill = "dropna"
         start, stop, granularity, needed_overlap, fill = api.validate_qs(
@@ -226,7 +228,7 @@ class AggregatesController(rest.RestController):
             if not groupby:
                 return self._get_measures_by_name(
                     resources, references, body["operations"], start, stop,
-                    granularity, needed_overlap, fill)
+                    granularity, needed_overlap, fill, details=details)
 
             def groupper(r):
                 return tuple((attr, r[attr]) for attr in groupby)
@@ -237,7 +239,7 @@ class AggregatesController(rest.RestController):
                     "group": dict(key),
                     "measures": self._get_measures_by_name(
                         resources, references, body["operations"], start, stop,
-                        granularity, needed_overlap, fill)
+                        granularity, needed_overlap, fill, details=details)
                 })
             return results
 
@@ -268,13 +270,19 @@ class AggregatesController(rest.RestController):
             metrics_by_ids = dict((six.text_type(m.id), m) for m in metrics)
             references = [processor.MetricReference(metrics_by_ids[m], a)
                           for (m, a) in references]
-            return {"references": references,
-                    "measures": get_measures_or_abort(
-                        references, body["operations"],
-                        start, stop, granularity, needed_overlap, fill)}
+
+            response = {
+                "measures": get_measures_or_abort(
+                    references, body["operations"],
+                    start, stop, granularity, needed_overlap, fill)
+            }
+            if details:
+                response["references"] = references
+            return response
 
     def _get_measures_by_name(self, resources, metric_names, operations,
-                              start, stop, granularity, needed_overlap, fill):
+                              start, stop, granularity, needed_overlap, fill,
+                              details):
 
         references = [
             processor.MetricReference(r.get_metric(metric_name), agg, r)
@@ -286,7 +294,11 @@ class AggregatesController(rest.RestController):
             api.abort(400, {"cause": "Metrics not found",
                             "detail": set((m for (m, a) in metric_names))})
 
-        return {"references": references,
-                "measures": get_measures_or_abort(
-                    references, operations, start, stop, granularity,
-                    needed_overlap, fill)}
+        response = {
+            "measures": get_measures_or_abort(
+                references, operations, start, stop, granularity,
+                needed_overlap, fill)
+        }
+        if details:
+            response["references"] = references
+        return response
--- a/gnocchi/rest/aggregates/api.py
+++ b/gnocchi/rest/aggregates/api.py
@@ -119,32 +119,42 @@ def OperationsSchema(v):
                              required=True)(v)
 
 
+class ReferencesList(list):
+    "A very simplified OrderedSet with list interface"
+
+    def append(self, ref):
+        if ref not in self:
+            super(ReferencesList, self).append(ref)
+
+    def extend(self, refs):
+        for ref in refs:
+            self.append(ref)
+
+
 def extract_references(nodes):
-    references = set()
+    references = ReferencesList()
     if nodes[0] == "metric":
         if isinstance(nodes[1], list):
             for subnodes in nodes[1:]:
-                references.add(tuple(subnodes))
+                references.append(tuple(subnodes))
         else:
-            references.add(tuple(nodes[1:]))
+            references.append(tuple(nodes[1:]))
     else:
         for subnodes in nodes[1:]:
             if isinstance(subnodes, list):
-                references |= extract_references(subnodes)
+                references.extend(extract_references(subnodes))
     return references
 
 
-def get_measures_or_abort(metrics_and_aggregations, operations, start,
-                          stop, granularity, needed_overlap, fill,
-                          ref_identifier):
+def get_measures_or_abort(references, operations, start,
+                          stop, granularity, needed_overlap, fill):
     try:
         return processor.get_measures(
             pecan.request.storage,
-            metrics_and_aggregations,
+            references,
             operations,
             start, stop,
-            granularity, needed_overlap, fill,
-            ref_identifier=ref_identifier)
+            granularity, needed_overlap, fill)
     except exceptions.UnAggregableTimeseries as e:
         api.abort(400, e)
     # TODO(sileht): We currently got only one metric for these exceptions but
@@ -187,7 +197,7 @@ class AggregatesController(rest.RestController):
 
         body = api.deserialize_and_validate(self.FetchSchema)
 
-        references = list(extract_references(body["operations"]))
+        references = extract_references(body["operations"])
         if not references:
             api.abort(400, {"cause": "Operations is invalid",
                             "reason": "At least one 'metric' is required",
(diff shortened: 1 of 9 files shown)


sileht:
Pull-request updated, HEAD is now 1e103f65bffd7dcd22bbc7169964810d0d167596

sileht:
Pull-request updated, HEAD is now 6254b9eca41f98cbcbdd4b6adc971c514db016c7

sileht:
Pull-request updated, HEAD is now f97f9e3edd81ff254716f9d69c758a1bec819295

sileht:
Pull-request updated, HEAD is now 06058c4fc6d41f60be75ffe8885b22af3c382343

sileht:
Pull-request updated, HEAD is now 06058c4fc6d41f60be75ffe8885b22af3c382343

sileht:
This is API breakage of the new /v1/aggregates, but the API in 4.1 is not good enough to be used without additional API requests. Redoing queries to got metrics and resources details while the aggreages API already have them it's a waste of times and resources, so the change expose them.
I propose to backport this in 4.1 when it's merged, WDYT ?

This is ready to review, but not to merged, I need to figure out how to load metric.resource instead of metric.resource_id with sqlalchemy first.

gnocchi/rest/aggregates/api.py
```diff
@@ -232,8 +233,8 @@ def groupper(r):
 
         else:
             try:
-                metric_ids = [six.text_type(utils.UUID(m))
-                              for (m, a) in references]
+                metric_ids = set([six.text_type(utils.UUID(m))
```
jd: You should remove the `[]` to build the set without an intermediate list.

sileht:
Pull-request updated, HEAD is now 515891aa05b54af62d93217d9783eea41e1d7a71

sileht:
Pull-request updated, HEAD is now e1d7c28ab0614ed8b5c1e0efb137fadc150579cc

gnocchi/rest/aggregates/api.py
sileht: done

sileht:
> One thing that I just wonder: you mention only one of the timeseries is outputed when the names are the same; maybe having a different name with a default suffix of _1 or whatever could solve it?

Yeah we could also do that. But <metric.name>_<aggregation>_N is still a bad format. On client side, I have to split on '_'. Maybe metric.name will have "_" inside... 

Also we retrieve, internally, all resources and metrics related to the user query. And just return a (obscur) ID. I think all applications will have to get the metric details at some point.

A Grafana use case is, I show cpu usage of all instances of the server group "my_autoscaling_group".

Grafana user have the ability to use resource attributes to label displayed series, he can write thing like "$display_name [$flavor]". Current Grafana version (without aggregates API) just do an API call "a resource search" to get all metrics IDs and one call to /measures for each metric.

If I keep the current format (<metric.name>_<aggregation>_N), I will do one API call to get all timeseries, then for each timeseries I need to get the metric, and then to get the associated resources.

Since Gnocchi API already have everything computed under the hood. I really need/want to expose it.

> We talked about adding a as operator to have (as (metric <uuid>) "foobar") to have a defined name. Wouldn't that too help?

A custom name will not help, my use case is about having metric/resource details related the retrieved timeseries.

chungg:
> A custom name will not help, my use case is about having metric/resource details related the retrieved timeseries.

did you ever test how much overhead it was to add all this resource/metric details? maybe try an edge case of result with 100metrics?

gnocchi/indexer/sqlalchemy.py
```diff
@@ -1104,7 +1105,18 @@ def list_resources(self, resource_type='generic',
...
-            return all_resources
+            if detailed_metrics:
+                return list(map(self._load_detailed_metrics, all_resources))
+            else:
+                return all_resources
+
+    @staticmethod
+    def _load_detailed_metrics(r):
+        # NOTE(sileht): This moves the r in metric.resource without
+        # querying the database
```
chungg: just a question, but this is true only becasue we use joinedload above right?

gnocchi/rest/aggregates/processor.py
```diff
@@ -175,11 +179,25 @@ def aggregated(refs_and_timeseries, operations, from_timestamp=None,
...
-                                            result[ident]['values'])))
-                 for ident in result))
+            result[i]["timestamps"].extend(t)
+            result[i]['granularity'].extend([granularity] * len(t))
+            result[i]['values'].extend(v)
+
+    # NOTE(sileht): the output list is ordered like the input metric list,
+    # this makes the output always ordered in the same maner.
+    output = []
+    for i, ident in enumerate(idents):
```
chungg: this seems like it's only handling the last sampling key. i'd much prefer we not have multiple loops and this be moved with previous

`for i, ident in enumerate(idents):`

gnocchi/indexer/sqlalchemy.py
sileht: yes

gnocchi/rest/aggregates/processor.py
sileht: This is something ugly I wrote... but idents are always the same whatever the sampling is.

I will refactor the code a bit to remove this oddity. Maybe into a separate patch if the change is too big.

sileht:
> did you ever test how much overhead it was to add all this resource/metric details? maybe try an edge case of result with 100metrics?

The payload will be a lot bigger for sure, but nobody can use an output like:

```
cpu.util_mean_1: MEASURES
cpu.util_mean_2: MEASURES
cpu.util_mean_3: MEASURES
cpu.util_mean_4: MEASURES
```

My current version is not good enough because the resource detail is repeated for each timeseries.

Also I don't like to create a special syntax to name things. After some thinking I plan to change the format  to:

For queries with metric ids:

```
{
    "references": list-of-metrics,

    "aggregated": MEASURES,
      - OR -
     "measures": {
         "metric-id": {"mean": MEASURES}
    }
}
```

For queries with resource search:

```
{
    "references": list-of-resource-with-metric-detail, 

    "aggregated": MEASURES,
      - OR -
    "measures": {
             "resource-id": {"metric-name": {"mean": MEASURES, "max": MEASURES}}
    }
}
```

In this new version, I can make `references` optional with something like &detail=true.

sileht:
Pull-request updated, HEAD is now 170ee93810c2f0d4101fa6aabd32608945148d29

sileht:
Pull-request updated, HEAD is now 42dcff15463a1fdae77688f475b4e0d598f18e9b

sileht:
Pull-request updated, HEAD is now ef4f41ce2f3e8ab62e83d9e74fb3ea1b2c2cd00c

requirements.txt
```diff
@@ -4,6 +4,7 @@ iso8601
 oslo.config>=3.22.0
 oslo.policy>=0.3.0
 oslo.middleware>=3.22.0
+orderedset
```
jd: unused?

sileht:
Pull-request updated, HEAD is now 82f3ab36de8fda4d1d10fb51f416f1d95a036dbb

gnocchi/rest/aggregates/api.py
```diff
@@ -255,24 +268,37 @@ def groupper(r):
...
     def _get_measures_by_name(self, resources, metric_names, operations,
-                              start, stop, granularity, needed_overlap, fill):
+                              start, stop, granularity, needed_overlap, fill,
+                              details):
+
+        references = list(map(lambda x: processor.MetricReference(*x),
+                              filter(lambda x: x[0] is not None,
+                                     ((r.get_metric(metric_name), agg, r)
+                                      for (metric_name, agg) in metric_names
+                                      for r in resources))))
```
chungg: do we need this filter or can we just do:

```
processor.MetricReference(r.get_metric(metric_name), agg, r)
for (metric_name, agg) in metric_names
for r in resources if r.get_metric(metric_name) is not None
```

gnocchi/rest/aggregates/processor.py
```diff
@@ -75,62 +100,62 @@ def get_measures(storage, metrics_and_aggregations,
...
                               for g in granularities_in_common])
 
     return aggregated(tss, operations, from_timestamp, to_timestamp,
                       needed_overlap, fill)
 
 
 def aggregated(refs_and_timeseries, operations, from_timestamp=None,
-               to_timestamp=None, needed_percent_of_overlap=100.0, fill=None):
+               to_timestamp=None, needed_percent_of_overlap=100.0, fill=None,
+               ref_identifier="id"):
```
chungg: ref_identifier unused

gnocchi/tests/test_aggregates.py
```diff
@@ -1452,7 +1535,7 @@ def test_bool(self):
         ], values)
 
     def test_unary_operator(self):
-        metric2, __ = self._create_metric()
+        metric2, metric2_sql = self._create_metric()
```
chungg: not needed?

requirements.txt
sileht: done

gnocchi/rest/aggregates/api.py
sileht: That's work too and this is more readable.

gnocchi/rest/aggregates/processor.py
sileht: cleaned

gnocchi/tests/test_aggregates.py
sileht: cleaned

sileht:
Pull-request updated, HEAD is now b6f78fcc97bdff6e48624a63daa70adf3c2e63dc

sileht:
Pull-request updated, HEAD is now 14c86397b0fb2cc13b62693813cc07310b9a9a8e

sileht:
Pull-request updated, HEAD is now 34818e46cc12331936af14da6c1df16d8d1dd502

sileht:
Pull-request updated, HEAD is now 3f9932083fc45a9e3b9fd8081e8d787f192a6c7e

releasenotes/notes/aggregates-api-output-change-2bc6620c7f595925.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Aggregates API output introduced in 4.1.0 doesn't permit to easly identify which timeseries
```
chungg: s/permit to easly identify which timeseries/allow for easy identification of which timeseries/

PR #480, opened 2017-11-13
cap pandas

chungg:
pandas 0.21.0 does not work with numpy<1.13.0. we remove pandas
in master already so let's just cap pandas instead of bumping numpy


jd:
Yeah this is not the correct fix but I guess we don't have a choice because of pip limitations. :(

chungg:
@jd ok. i see your point. let's try something else.

PR #482, opened 2017-11-14
Pull request for influxdb

jd:
api: add an InfluxDB API endpoint
api: rename refresh_timeout option to operation_timeout
cli: use --http-socket rather than --http for uwsgi

### DIFF
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -1925,7 +1925,8 @@ class BatchController(object):
          indexer.ResourceTypeAlreadyExists,
          indexer.NamedMetricAlreadyExists)))
 def get_or_create_resource_and_metrics(
-        creator, rid, original_resource_id, job, instance, metric_names,
+        creator, rid, original_resource_id, metric_names,
+        resource_attributes,
         resource_type, resource_type_attributes=None):
     try:
         r = pecan.request.indexer.get_resource(resource_type, rid,
@@ -1973,19 +1974,17 @@ def get_or_create_resource_and_metrics(
             "resource_type": resource_type,
             "creator": creator,
             "original_resource_id": original_resource_id,
-            "job": job,
-            "instance": instance,
             "metrics": metrics,
         }
+        target.update(resource_attributes)
         enforce("create resource", target)
 
         try:
             return pecan.request.indexer.create_resource(
                 resource_type, rid, creator,
                 original_resource_id=original_resource_id,
-                job=job,
-                instance=instance,
-                metrics=metrics
+                metrics=metrics,
+                **resource_attributes,
             ).metrics
         except indexer.ResourceAlreadyExists:
             # NOTE(sileht): ensure the rid is not registered whitin another
@@ -2033,7 +2032,8 @@ class PrometheusWriteController(rest.RestController):
             rid = ResourceUUID(original_rid, creator=creator)
             metric_names = list(measures.keys())
             metrics = get_or_create_resource_and_metrics(
-                creator, rid, original_rid, job, instance, metric_names,
+                creator, rid, original_rid, metric_names,
+                dict(job=job, instance=instance),
                 "prometheus", self.PROMETHEUS_RESOURCE_TYPE)
 
             for metric in metrics:
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -500,7 +500,7 @@ class MetricController(rest.RestController):
             try:
                 pecan.request.storage.refresh_metric(
                     pecan.request.indexer, pecan.request.incoming, self.metric,
-                    pecan.request.conf.api.refresh_timeout)
+                    pecan.request.conf.api.operation_timeout)
             except storage.SackLockTimeoutError as e:
                 abort(503, six.text_type(e))
         try:
@@ -1816,7 +1816,7 @@ class AggregationController(rest.RestController):
                     try:
                         pecan.request.storage.refresh_metric(
                             pecan.request.indexer, pecan.request.incoming, m,
-                            pecan.request.conf.api.refresh_timeout)
+                            pecan.request.conf.api.operation_timeout)
                     except storage.SackLockTimeoutError as e:
                         abort(503, six.text_type(e))
             if number_of_metrics == 1:
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -1940,22 +1940,25 @@ class PrometheusWriteController(rest.RestController):
              indexer.NamedMetricAlreadyExists)))
     def get_or_create_resource_and_metrics(
             creator, rid, original_resource_id, job, instance, metric_names,
-            resource_type, resource_type_attributes):
+            resource_type, resource_type_attributes=None):
         try:
             r = pecan.request.indexer.get_resource(resource_type, rid,
                                                    with_metrics=True)
         except indexer.NoSuchResourceType:
-            enforce("create resource type", {
-                'name': resource_type,
-                'state': 'creating',
-                'attributes': resource_type_attributes,
-            })
+            if resource_type_attributes:
+                enforce("create resource type", {
+                    'name': resource_type,
+                    'state': 'creating',
+                    'attributes': resource_type_attributes,
+                })
 
-            schema = pecan.request.indexer.get_resource_type_schema()
-            rt = schema.resource_type_from_dict(
-                resource_type, resource_type_attributes, 'creating')
-            pecan.request.indexer.create_resource_type(rt)
-            raise tenacity.TryAgain
+                schema = pecan.request.indexer.get_resource_type_schema()
+                rt = schema.resource_type_from_dict(
+                    resource_type, resource_type_attributes, 'creating')
+                pecan.request.indexer.create_resource_type(rt)
+                raise tenacity.TryAgain
+            else:
+                raise
         except indexer.UnexpectedResourceTypeState as e:
             # NOTE(sileht): Currently created by another thread
             if not e.state.endswith("_error"):
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -1917,6 +1917,85 @@ class BatchController(object):
     resources = ResourcesBatchController()
 
 
+# Retry with exponential backoff for up to 1 minute
+@tenacity.retry(
+    wait=tenacity.wait_exponential(multiplier=0.5, max=60),
+    retry=tenacity.retry_if_exception_type(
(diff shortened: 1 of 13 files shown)


jd:
Pull-request updated, HEAD is now 0d8f2b7fd2925682c1db6830d55b171e4ca238a6

jd:
Pull-request updated, HEAD is now 4f5e4d71601486d9779914b83bdb34b5eeea5928

jd:
Pull-request updated, HEAD is now ad617be414323e357efb02ea4853df33e5fa6bf1

jd:
Pull-request updated, HEAD is now 790ff3303da9caed3134e3bc907e4776060d00d7

jd:
Pull-request updated, HEAD is now 1c6405f7f0d9127a1924ebc2ebe96c2a990c3e1a

jd:
Pull-request updated, HEAD is now 31a2b7b798d112f3bb93efb82026f4ed9e2c943b

gnocchi/rest/influxdb.py
```diff
@@ -0,0 +1,316 @@
...
+            pecan.request.incoming.add_measures_batch(measures_to_batch)
+            pecan.response.status = 204
+
+            if encoding != "chunked":
+                return
+
+    @staticmethod
+    @tenacity.retry(retry=tenacity.retry_if_exception_type(
+        (indexer.NoSuchResource, indexer.NamedMetricAlreadyExists)))
+    def get_or_create_resource_and_metrics(resource_type, creator, rid,
```
sileht: Maybe we can share it with the one in Prometheus

gnocchi/tests/indexer/sqlalchemy/test_migrations.py
```diff
@@ -49,6 +49,20 @@ def setUp(self):
         self.index = indexer.get_driver(self.conf)
         self.index.upgrade(nocreate=True)
         self.addCleanup(self._drop_database)
+        # NOTE(sileht): remove tables dynamically created by other tests
```
sileht: That may need to be rebased, mine have been merged.

gnocchi/rest/influxdb.py
jd: Yeah I can try to do that now that Prometheus is merged.

gnocchi/tests/indexer/sqlalchemy/test_migrations.py
jd: ack

jd:
Pull-request updated, HEAD is now 53989ae44aa27bfea8cf47a2f59be07a152f779c

jd:
Pull-request updated, HEAD is now 439d1d367fa241c8c895e8719e327f154ad501e9

jd:
Pull-request updated, HEAD is now 439d1d367fa241c8c895e8719e327f154ad501e9

jd:
Pull-request updated, HEAD is now aa33aeeed592d1ac00d0d4fa4a69ac32d12bde4d

gnocchi/rest/api.py
```diff
@@ -2055,6 +2057,7 @@ def __init__(self):
             "capabilities": CapabilityController(),
             "status": StatusController(),
             "aggregates": agg_api.AggregatesController(),
+            "influxdb": influxdb.InfluxDBController(),
```
chungg: can we enable this on demand?

gnocchi/rest/influxdb.py
```diff
@@ -0,0 +1,267 @@
...
+
+    @pecan.expose('json')
+    def post_query(self, q=None):
+        if q is not None:
+            try:
+                query = query_parser.parseString(q)
+            except pyparsing.ParseException:
+                api.abort(501, {"cause": "Not implemented error",
+                                "detail": "q",
+                                "reason": "Querying is not implemented"})
```
chungg: this error is confusing... i don't see how 'querying is not implemented', this is literally what the endpoint is. is it something 'unsupported' that will throw this?

gnocchi/rest/influxdb.py
```diff
@@ -0,0 +1,267 @@
...
+                pecan.request.indexer.create_resource_type(rt)
+            except indexer.ResourceTypeAlreadyExists:
+                pass
+            pecan.response.status = 204
+
+    @staticmethod
+    def _write_get_lines():
+        encoding = pecan.request.headers.get('Transfer-Encoding', "").lower()
+        if encoding == "chunked":
+            if uwsgi is None:
```
chungg: how usable is this without chunked? if it isn't, i don't think we should support this especially since we're checking this constantly.

gnocchi/rest/influxdb.py
```diff
@@ -0,0 +1,267 @@
...
+                            stop=tenacity.stop_after_delay(timeout))(
+                                creator, resource_id, resource_name,
+                                metrics_and_measures.keys(),
+                                {}, db)
+                    )
+                except indexer.ResourceAlreadyExists as e:
+                    # If this function raises ResourceAlreadyExists it means
+                    # the resource might already exist as another type, we
+                    # can't continue.
+                    LOG.error("Unable to get resource `%s' for InfluxDB, "
```
chungg: s/get/create/

gnocchi/tests/functional/gabbits/influxdb.yaml
```diff
@@ -0,0 +1,111 @@
...
+      GET: /v1/resource/influxdbtest/foobar
+      status: 200
+      response_json_paths:
+        $.original_resource_id: foobar
+        $.id: b4d568e4-7af1-5aec-ac3f-9c09fa3685a9
+        $.type: influxdbtest
+        $.creator: admin
+
+    - name: check metric created
+      GET: /v1/resource/influxdbtest/foobar/metric/mymetric.field@mytag=myvalue
```
chungg: this should assert something.

gnocchi/tests/functional/gabbits/influxdb.yaml
```diff
@@ -0,0 +1,111 @@
...
+      GET: /v1/resource/influxdbtest/myvalue
+      status: 200
+      response_json_paths:
+        $.original_resource_id: myvalue
+        $.id: 6b9e2039-98d0-5d8d-9153-2d7491cf13e5
+        $.type: influxdbtest
+        $.creator: admin
+
+    - name: check metric created different tag resource id
+      GET: /v1/resource/influxdbtest/myvalue/metric/mymetric.field@host=foobar
```
chungg: assert something

gnocchi/tests/functional/gabbits/influxdb.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+    - name: check metric created different tag resource id and slash replaced
+      GET: /v1/resource/influxdbtest/myvalue/metric/cpu.field@path=_foobar
```
chungg: assert something

gnocchi/rest/api.py
jd: Probably, though we don't do that for Prometheus either. So what would be the usecase?

gnocchi/rest/influxdb.py
jd: We don't support InfluxDB query language. I changed to "query not implemented" to be more precise about the fact that this query is not implemented.
The only supported is `CREATE DATABASE` that's not what I'd call a query ;)

gnocchi/rest/influxdb.py
jd: it's completely usable without chunked encoding. That's what the tests do.

But real InfluxDB client sends chunked encoding.

gnocchi/rest/influxdb.py
jd: fixed

gnocchi/tests/functional/gabbits/influxdb.yaml
jd: it does, it asserts there's no error being returned (i.e. no 404).

it does not check any content since it's not the purpose of this test. It's not testing the format, it's testing the metric creation.

gnocchi/tests/functional/gabbits/influxdb.yaml
jd: ditto

gnocchi/tests/functional/gabbits/influxdb.yaml
jd: ditto

gnocchi/rest/api.py
chungg: we do, it's only enabled if the right libs are there although we can't do the same here... i just don't want to load it unless explicit.

gnocchi/rest/influxdb.py
chungg: so unchunked is only for tests?

jd:
Pull-request updated, HEAD is now 8b0b52f8cb7ed86f88269974c20014c463925d95

gnocchi/tests/functional/gabbits/influxdb.yaml
chungg: word. didn't realise gabbi would auto fail  for these cases. i'd prefer it be explicit and we validate more than just 'it didn't explode' but your call.

gnocchi/rest/api.py
jd: If I have snappy installed on my system for whatever reason then I cannot disable Prometheus support.
This isn't a correct response. :/

That being said, we need something more than just a switch IMHO, see #499. So I'd postponed any switch until we solve #499.

gnocchi/rest/influxdb.py
jd: In real life, really likely, until someone starts to talk InfluxDB to Gnocchi in regular encoding (but I don't think InfluxDB does that?)

gnocchi/tests/functional/gabbits/influxdb.yaml
jd: yeah it's always a trade-off: asserting `200` for example would break if we ever change the code to be `202` for example, whereas here we don't really care.

Though I agree in this case there's 0% chance we ever change from 200 :)

jd:
Fix test after 'Query' word change

jd:
Pull-request updated, HEAD is now b4264aeadeb53bfa4228bd82dc02d3ae7258f632

jd:
> You finally not sharing get_or_create_resource_and_metrics() with prometheus ?

@sileht of course it is shared. did I miss something?

sileht:
@jd, no it's OK I have misreading

PR #489, opened 2017-11-15
indexer: make sqlalchemy Metadata thread-safe

sileht:
sqlalchemy provides a threadsafe version of Metadata() object.

Since we load resource type schema dynamically this will avoid race
where our cache is not synced.


sileht:
I hold this patch since it doesn't solve the sql migration tests issue we have in prometheus and influx db patch

PR #490, opened 2017-11-16
rest: don't return body on 202

### LINKED ISSUE #463: `POST /v1/batch/resources/metrics/measures` returns `null`
From the doc:
```
POST /v1/batch/resources/metrics/measures HTTP/1.1
Content-Length: 585

{
  "15e9c872-7ca9-11e4-a2da-2fb4032dfc09": {
    "cpu.util": [
      {
        "timestamp": "2014-10-06T14:34:12",
        "value": 12
      },
      {
        "timestamp": "2014-10-06T14:34:20",
        "value": 2
      }
    ]
  },
  "6f24edd9-5a2f-4592-b708-ffbed821c5d2": {
    "cpu.util": [
      {
        "timestamp": "2014-10-06T14:34:12",
        "value": 6
      },
      {
        "timestamp": "2014-10-06T14:34:20",
        "value": 25
      }
    ]
  },
  "ab68da77-fa82-4e67-aba9-270c5a98cbcb": {
    "temperature": [
      {
        "timestamp": "2014-10-06T14:34:12",
        "value": 17
      },
      {
        "timestamp": "2014-10-06T14:34:20",
        "value": 18
      }
    ]
  }
}
HTTP/1.1 202 Accepted
Content-Length: 4

null
```

Should probably be empty, not null. I guess pecan converts the `None` to `null`.

sileht:
Batch endpoint tell pecan body is a json, but we don't return body.
So pecan generates the body "null"

This change fixes that.

It also adds tests to ensure the all 202 don't have body set.

Closes #463

### DIFF
--- a/gnocchi/rest/app.py
+++ b/gnocchi/rest/app.py
@@ -23,6 +23,7 @@ from oslo_policy import policy
 from paste import deploy
 import pecan
 from pecan import jsonify
+from pecan import templating
 from stevedore import driver
 import webob.exc
 
@@ -73,6 +74,17 @@ class NotImplementedMiddleware(object):
             raise webob.exc.HTTPNotImplemented(
                 "Sorry, this Gnocchi server does "
                 "not implement this feature 😞")
+
+
+class JsonRenderer(templating.JsonRenderer):
+    def render(self, template_path, namespace):
+        # NOTE(sileht): Unlike the builtin renderer of pecan
+        # we don't want to return "null" for None. Our API
+        # returns only empty, list or dict.
+        if namespace is None:
+            return ""
+        return super(JsonRenderer, self).render(template_path, namespace)
+
 
 # NOTE(sileht): pastedeploy uses ConfigParser to handle
 # global_conf, since python 3 ConfigParser doesn't
@@ -130,6 +142,7 @@ def _setup_app(root, conf, indexer, storage, incoming,
         root,
         hooks=(GnocchiHook(storage, indexer, incoming, conf),),
         guess_content_type_from_ext=False,
+        custom_renderers={"json": JsonRenderer}
     )
 
     if not_implemented_middleware:
--- a/gnocchi/gendoc.py
+++ b/gnocchi/gendoc.py
@@ -41,6 +41,9 @@ def _format_json(txt):
 
 def _extract_body(req_or_resp):
     # TODO(jd) Make this a Sphinx option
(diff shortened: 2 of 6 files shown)


sileht:
`So I think this does not work in the case where an error is returned, because the error won't be returned as JSON. You can prove me wrong by adding a test that e.g. hit L1558`

This is what we do on other endpoints where we returns 202. But tests tell me that will not work :(

sileht:
Pull-request updated, HEAD is now e8823a565b988a117d5b47bc21129ce431f19efa

sileht:
Pull-request updated, HEAD is now a0265c3ad6323e8c81d5f5486bf4e12eb7a8d8b6

gnocchi/rest/app.py
```diff
@@ -74,6 +74,20 @@ def __call__(self, environ, start_response):
                 "Sorry, this Gnocchi server does "
                 "not implement this feature 😞")
 
+
+class JsonRenderer(object):
+    def __init__(self, path, extra_vars):
+        pass
+
+    def render(self, template_path, namespace):
```
jd: @staticmethod

gnocchi/rest/app.py
```diff
@@ -74,6 +74,20 @@ def __call__(self, environ, start_response):
                 "Sorry, this Gnocchi server does "
                 "not implement this feature 😞")
 
+
+class JsonRenderer(object):
+    def __init__(self, path, extra_vars):
```
jd: @staticmethod

gnocchi/rest/app.py
```diff
@@ -74,6 +74,20 @@ def __call__(self, environ, start_response):
...
+        pass
+
+    def render(self, template_path, namespace):
+        # NOTE(sileht): Unlike the builtin renderer of pecan
+        # we don't want to return "null" for None. Our API
+        # returns only empty, list or dict.
+        if namespace is None:
+            return ""
+        return jsonify.encode(namespace)
```
jd: any reason you don't `return super(JsonRenderer, self).render()`?

sileht:
Pull-request updated, HEAD is now 4928b6868da68ad38967b855fdbf60f163ea0eb3

sileht:
Pull-request updated, HEAD is now ccfddee0fa058fe1cd7e0936487ae0cf85ae7348

Issue #314, opened 2017-08-31
Metric listing do not include accessible metrics


david-hill:
Hi guys,

   Gnocchi metric list is pretty much broken.  Let me explain this: 
1) When ceilometer creates a metric, it uses the ceilometer user / project
2) When "gnocchi metric list" is ran, it takes the user and project of the current user
3) This will always be empty for non-admin users

Also, for some reasons, I can't get more than 1004 metrics on a given system.   I tried finding a hard coded limit and the only one close to that is max_limit that is set at 1000.   Does this make sense?

Thank you very much,

David Hill

jd:
Hi David,

The metric limit might be due to the CLI not using paging, which is an issue opened here https://github.com/gnocchixyz/python-gnocchiclient/issues/4

Gnocchi metric list is supposed to list metric created by users, which as you correctly noted, are not created by the user but by Ceilometer.

The resource created by Ceilometer are however exposed to users and their metric are also accessible to users. They are easy to see by using `gnocchi resource list` and `gnocchi resource show`.

It's not really possible to list metric accessible by users right now, because the ACL engine is based on _oslo.policy_ which does not allow that. We'll probably need to replace that engine at some time to allow to express such permissions.

david-hill:
Hello Julien,

    That's where I started my investigation and found out the database counted 550k entries from the ceilimeter user alone ... and no other user ids.  Would you have an idea of why the "gnocchi metric list" would only return 1004 elements put of 550k ?

David Hill

On Sep 1, 2017, at 04:50, Julien Danjou <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi David,

The metric limit might be due to the CLI not using paging, which is an issue opened here gnocchixyz/python-gnocchiclient#4<https://github.com/gnocchixyz/python-gnocchiclient/issues/4>

Gnocchi metric list is supposed to list metric created by users, which as you correctly noted, are not created by the user but by Ceilometer.

The resource created by Ceilometer are however exposed to users and their metric are also accessible to users. They are easy to see by using gnocchi resource list and gnocchi resource show.

It's not really possible to list metric accessible by users right now, because the ACL engine is based on oslo.policy which does not allow that. We'll probably need to replace that engine at some time to allow to express such permissions.

jd:
> Would you have an idea of why the "gnocchi metric list" would only return 1004 elements put of 550k ?

Yes, I just told you:

[...]

PR #498, opened 2017-11-21
Pull request for metric-list-attribute-filter

jd:
indexer: replace ids argument with attribute filter in list_metrics()
indexer: replace names filter in list_metric by attribute_filter
indexer: replace kwargs filtering by regular attribute filter

### DIFF
--- a/gnocchi/tests/test_indexer.py
+++ b/gnocchi/tests/test_indexer.py
@@ -348,7 +348,8 @@ class TestIndexerDriver(tests_base.TestCase):
         self.assertRaises(indexer.NoSuchResource,
                           self.index.delete_resource,
                           r1)
-        metrics = self.index.list_metrics(ids=[e1, e2])
+        metrics = self.index.list_metrics(
+            attribute_filter={"in": {"id": [e1, e2]}})
         self.assertEqual([], metrics)
 
     def test_delete_resource_non_existent(self):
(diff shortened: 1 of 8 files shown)


jd:
Pull-request updated, HEAD is now f934488054169310f2deaa2d577988651a6b13a1

jd:
This is a first step toward fixing https://github.com/gnocchixyz/gnocchi/issues/314

gnocchi/tests/test_indexer.py
```diff
@@ -1121,9 +1124,6 @@ def test_get_metric_with_bad_uuid(self):
         self.assertEqual([], self.index.list_metrics(
             attribute_filter={"=": {"id": e1}}))
 
-    def test_get_metric_empty_list_uuids(self):
```
chungg: what happens in this scenario now?

jd: Same things. it's basically `{"in" : {"id": []}}`

it's not a special case anymore that's why I removed the test

chungg: does it crash? i'm pretty sure it was there to make sure it didn't throw some SQL error which it did previously with empty list

jd:
Pull-request updated, HEAD is now 1de30ae5e448c16d46aa9a453da47858ee76c012

jd:
Pull-request updated, HEAD is now 5395f7481aa1a702b6078b686983bc06bc95342e

PR #500, opened 2017-11-22
aggregates: return resources/metrics only once

sileht:
We currently returns all references, but when many of them point to the
same resource, we output each time the resource.

This change removes this resource duplication from the output.

### DIFF
--- a/gnocchi/indexer/__init__.py
+++ b/gnocchi/indexer/__init__.py
@@ -69,6 +69,8 @@ class Resource(object):
         return self.revision_start.replace(microsecond=0,
                                            tzinfo=iso8601.iso8601.UTC)
 
+    __hash__ = object.__hash__
+
 
 class Metric(object):
     def __init__(self, id, archive_policy, creator=None,
--- a/gnocchi/tests/functional/gabbits/aggregates-with-metric-ids.yaml
+++ b/gnocchi/tests/functional/gabbits/aggregates-with-metric-ids.yaml
@@ -36,24 +36,28 @@ tests:
     - name: create metric1
       POST: /v1/metric
       data:
+          name: metric1
+          archive_policy_name: cookies
+      status: 201
+
+    - name: create metric2
+      POST: /v1/metric
+      data:
+          name: metric2
           archive_policy_name: cookies
       status: 201
 
     - name: create metric3
       POST: /v1/metric
       data:
+          name: metric3
           archive_policy_name: cake
-      status: 201
-
-    - name: create metric2
-      POST: /v1/metric
-      data:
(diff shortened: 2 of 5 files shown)


sileht:
Pull-request updated, HEAD is now 1569a8957d3379e08bb4b7b86faeb2e5204f81f5

sileht:
Pull-request updated, HEAD is now b36e6bc45e171c77a711290795fe3ff591ff8655

sileht:
> I'll assume this does not change the returned API? :(

No `references` is still a `list` of `Metric`, or `Resource` for resources search. Just now we will not have duplicate items into the list that was just waste space.

gnocchi/indexer/__init__.py
```diff
@@ -69,6 +69,8 @@ def lastmodified(self):
         return self.revision_start.replace(microsecond=0,
                                            tzinfo=iso8601.iso8601.UTC)
 
+    __hash__ = object.__hash__
```
chungg: uh, what's the point of this hash? is this overriding anything?

gnocchi/tests/functional/gabbits/aggregates-with-metric-ids.yaml
```diff
@@ -170,10 +174,10 @@ tests:
       response_json_paths:
         $.`len`: 2
         $.references.`len`: 2
-        $.references[0].id: $HISTORY['create metric1'].$RESPONSE['$.id']
-        $.references[1].id: $HISTORY['create metric2'].$RESPONSE['$.id']
-        $.references[0].archive_policy.name: cookies
-        $.references[1].archive_policy.name: cookies
+        $.references[/name][0].id: $HISTORY['create metric1'].$RESPONSE['$.id']
```
chungg: today i learned '/' means sort

gnocchi/indexer/__init__.py
sileht: On python3, when __eq__ is defined we must defined __hash__ otherwise hashing is rejected.

> If a class does not define an __eq__() method it should not define a __hash__() operation either; if it defines __eq__() but not __hash__(), its instances will not be usable as items in hashable collections

Since I use Resource in set(), I have to define __eq__ and __hash__.

chungg: is this just a random hash though or is there some logic to how hash is generated? if it's random, i feel like we should be explicit on how the hash is generated ie. hash the same attributes as __eq__

sileht: By default, it's just random. In my case I just want to identify if two references point to the same object. I don't want to consider equal two different objects describing the same resource.

PR #505, opened 2017-11-22
Simplify Ceph flavor

jd:
Now that the `ceph` flavor is gone, let's use `ceph` for the recommended
approach and `ceph_alternative` for the other one.


jd:
Pull-request updated, HEAD is now ac73147d256ceca993d15f26e7bd9770e89c0ea1

chungg:
i'll be honest, i'm not entirely sure it's easy to understand why there are two ceph entries... regardless of what the name is.

jd:
it's more likely that users will use `ceph` by accident and make it work.
it's unlikely they will type `ceph_recommended_lib` by accident.

PR #508, opened 2017-11-22
redis: enhance error message is redis is not installed

jd:
There's is a python-redis on PyPI, and that's not what we want.
Let's not be confusing.

### DIFF
--- a/gnocchi/common/redis.py
+++ b/gnocchi/common/redis.py
@@ -73,7 +73,7 @@ CLIENT_INT_ARGS = frozenset([
 
 def get_client(conf):
     if redis is None:
-        raise RuntimeError("python-redis unavailable")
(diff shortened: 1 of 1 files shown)


gnocchi/common/redis.py
```diff
@@ -73,7 +73,7 @@
 
 def get_client(conf):
     if redis is None:
-        raise RuntimeError("python-redis unavailable")
+        raise RuntimeError("Redis Python module is unavailable")
```
chungg: why not just specify the exact lib?

jd: It's `redis`. If you can find a way to say that without being confusing about the redis-server, please halp meeeee

chungg: *shrugs*

PR #511, opened 2017-11-24
doc: move comparison in intro

jd:


chungg:
you have a reason :)... not that i care if it's in intro or index.

jd:
It's just not to have this on the frontpage

jd:
Pull-request updated, HEAD is now 72c637627c8636911e7f5fbea86779b80943b8b4

PR #512, opened 2017-11-24
doc: fix heading level for titles in intro

jd:
There are at level 1 where they should be level 2

### DIFF
--- a/doc/source/intro.rst
+++ b/doc/source/intro.rst
@@ -1,53 +1,8 @@
 Getting started
 ---------------
-Gnocchi uses three different back-ends for storing data: one for storing new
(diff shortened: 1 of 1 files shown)


doc/source/intro.rst
```diff
@@ -47,7 +47,7 @@ relationships of |resources|..
 .. _MySQL: http://mysql.org
 
 Architecture overview
----------------------
+~~~~~~~~~~~~~~~~~~~~~
```
chungg: tbh, it feels strange this is the same level as drivers... i feel like drivers should be lower

jd:
Pull-request updated, HEAD is now 3f0b38c048c18649e3b74026decc9e4d27e38480

PR #513, opened 2017-11-27
metricd: expose respect_processing_delay option

### LINKED ISSUE #507: Redis incoming driver does not respect 'delay' parameter for processing.
4.1

The Redis incoming driver should respect the 'delay' processing setting.

The incoming data in Redis is immediately processed even with a non-zero delay

The incoming data would be processed after 'delay' seconds and with a period of 'delay' seconds like the other incoming drivers.

jd:
This allows to disable eager processing of new measures.

Fixes #507

### DIFF
--- a/doc/source/operating.rst
+++ b/doc/source/operating.rst
@@ -190,8 +190,23 @@ uses the 6 default |aggregation methods| (mean, min, max, sum, std, count) with
 the same "one year, one minute aggregations" resolution, the space used will go
 up to a maximum of 6 × 4.1 MiB = 24.6 MiB.
 
-How many metricd workers do we need to run
-==========================================
+Metricd
+=======
+
+Metricd is the daemon responsible for processing measures, computing their
+aggregates and storing them into the aggregate storage. It also handles a few
+other cleanup tasks, such as deleting metrics marked for deletion.
+
+Metricd therefore is responsible for most of the CPU usage and I/O job in
+Gnocchi. The archive policy of each metric will influence how fast it performs.
+
+In order to process new measures, metricd checks the incoming storage for new
+measures from time to time. The delay between each check is can be configured
+by changing the `[metricd]metric_processing_delay` configuration option.
+
+
+How many metricd workers do I need to run
+-----------------------------------------
 
 By default, `gnocchi-metricd` daemon spans all your CPU power in order to
 maximize CPU utilisation when computing |metric| aggregation. You can use the
@@ -205,7 +220,7 @@ increase the number of `gnocchi-metricd` daemons. You can run any number of
 metricd daemon on any number of servers.
 
 How to scale measure processing
-===============================
+-------------------------------
 
 Measurement data pushed to Gnocchi is divided into sacks for better
 distribution. The number of partitions is controlled by the `sacks` option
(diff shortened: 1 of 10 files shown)


jd:
Pull-request updated, HEAD is now 34442dc3fe0d37af2326069644bc27eded71fa78

chungg:
i have a slight preference that we make users explicitly enable push rather than explicitly disable push but if everyone else prefers this, i'm ok as well.

jd:
Yeah I really prefer to have push per default. Most users are "complaining" about the delay, not about the eager processing. If you have no scale issue, disabling the notification system is removing a sugar.

jd:
Pull-request updated, HEAD is now 2d78ed3c7254a1f87b35d7ed84bb2442d1d05e52

chungg:
curious, why did you put "complaining" in quotes?

the reason i'd say the opposite is because this makes redis the only driver that behaves differently and it's only in the code that you would be able to figure out redis driver does something unique. it also seems strange to have a parameter to disable something that doesn't exist.

as a secondary, no one has benched this at scale, i'll be honest, the amount of messages this generates i'd be surprised if it actually works at scale.

jd:
> curious, why did you put "complaining" in quotes?

Because it's a more a feature request than a complaint (bug).

> the reason i'd say the opposite is because this makes redis the only driver that behaves differently and it's only in the code that you would be able to figure out redis driver does something unique. it also seems strange to have a parameter to disable something that doesn't exist.

Alright though that sounds more like a documentation issue than a code issue.

> as a secondary, no one has benched this at scale, i'll be honest, the amount of messages this generates i'd be surprised if it actually works at scale.

If your workers are all busy processing a giant sack, then it generates N messages per cycle at max (N = sack numbers), so by default 128 messages over 60 seconds = 2 notif/s. That does not sound like a scalability issue to me.
If you have a worst-case scenario that would show otherwise, please share though.

jd:
@chungg one thing that I can do, though, is renaming this option `respect_processing_delay=False` by `eager_measure_processing=True` (or something like that) if that makes it clearer.

chungg:
> Alright though that sounds more like a documentation issue than a code issue.

for sure. this is the minimum i'd request. we definitely need to highlight this functionality

> If your workers are all busy processing a giant sack, then it generates N messages per cycle at max (N = sack numbers), so by default 128 messages over 60 seconds = 2 notif/s. That does not sound like a scalability issue to me.
> If you have a worst-case scenario that would show otherwise, please share though.

hmm.. as i understand, every single write to redis will generate a MxN messages where M is the number of sacks written to and N is the number of workers. it just happens that the majority of the messages are discarded because it's not relevant to worker or because the worker already knows it needs to process it. so it's really increasing IO (memory) by some factor of MxN.

> `eager_measure_processing=True`

i like this option name... +documentation :)

jd:
> for sure. this is the minimum i'd request. we definitely need to highlight this functionality

Then you should already have opened a documentation issue rather than trying to kill this patch. :)

> hmm.. as i understand, every single write to redis will generate a MxN messages where M is the number of sacks written to and N is the number of workers. it just happens that the majority of the messages are discarded because it's not relevant to worker or because the worker already knows it needs to process it. so it's really increasing IO (memory) by some factor of MxN.

This is true **only if the sack is empty**. If the sack is not empty, not notifications is generated.
So unless your cluster is mostly/often idle, there are actually not that many notifications being generated.

> i like this option name... +documentation :)

Fair enough, I'll add that!

chungg:
uh... now that i think about it. do we want to disable the second write? i guess in theory, it doesn't change much... just whether we care that we are setting a bunch of stale keys constantly for no reason.

jd:
Pull-request updated, HEAD is now 0036fd2eb97655e942c88c6e57dd1604f29c2254

jd:
Pull-request updated, HEAD is now d11045324c521c79f3d4c0f9720898eab0baa558

PR #514, opened 2017-11-27
api: add a metric policy filter mechanism

jd:
Use the auth_helper plugin mechanism to filter metric listing rather than an
hardcoded rule.

The Keystone rule is identical than before and filters based on
created_by_project_id fake-attribute.

The basic/remoteuser rule is changed to mimic the one for resources, i.e., no
filtering is enforced.

### DIFF
--- a/gnocchi/tests/functional/gabbits/metric-list.yaml
+++ b/gnocchi/tests/functional/gabbits/metric-list.yaml
@@ -89,7 +89,7 @@ tests:
     - name: list metrics
       GET: /v1/metric
       response_json_paths:
-          $.`len`: 2
+          $.`len`: 4
 
     - name: list metrics by id
       GET: /v1/metric?id=$HISTORY['create metric 1'].$RESPONSE['id']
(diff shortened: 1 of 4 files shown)


gnocchi/tests/functional/gabbits/metric-list.yaml
```diff
@@ -89,7 +89,7 @@ tests:
     - name: list metrics
       GET: /v1/metric
       response_json_paths:
-          $.`len`: 2
+          $.`len`: 4
```
chungg: tbh, i don't really understand what the goal is. i don't think this makes sense that user foobar can list metrics from user foobaz and user jd.

we should probably add more tests to show what you're trying to achieve.

jd: I hoped the commit message was pretty explicit.

chungg: uh... so do we have the equivalent keystone tests to ensure that path still does what it should?

jd: Yes, it's in `test_rest`, that's the only place we test ACL.

PR #515, opened 2017-11-28
py3: pass a bytes type object to tooz coordinator

javacruft:
Ensure that a bytes type id is passed into tooz otherwise
metricd will fail to join the coordination group due to
an error concatenating the GROUP_ID with a str in tooz
when used with the memcached driver.

### DIFF
--- a/gnocchi/cli/metricd.py
+++ b/gnocchi/cli/metricd.py
@@ -132,7 +132,7 @@ class MetricReporting(MetricProcessBase):
 
 class MetricProcessor(MetricProcessBase):
     name = "processing"
-    GROUP_ID = "gnocchi-processing"
+    GROUP_ID = b"gnocchi-processing"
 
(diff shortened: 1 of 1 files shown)


gnocchi/cli/metricd.py
```diff
@@ -162,7 +162,8 @@ def _configure(self):
         try:
             self.partitioner = self.coord.join_partitioned_group(
                 self.GROUP_ID, partitions=200)
-            LOG.info('Joined coordination group: %s', self.GROUP_ID)
+            LOG.info('Joined coordination group: %s',
+                     self.GROUP_ID.decode('utf8'))
```
jd: Can you just use `decode()` without arg? It ought to work I think and allows to use whatever encoding is available rather than forcing an encoding we don't know about.

javacruft: Yes - that works fine - updated.

PR #522, opened 2017-11-29
redis: add documentation for the url

### LINKED ISSUE #520: Redis option are not documented
Options available through the URL are not documented.

https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/common/redis.py

sileht:
Closes #520

### DIFF
--- a/gnocchi/common/redis.py
+++ b/gnocchi/common/redis.py
@@ -16,6 +16,7 @@
 
 from __future__ import absolute_import
 
+from oslo_config import cfg
 from six.moves.urllib import parse
 
 try:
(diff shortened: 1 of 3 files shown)


gnocchi/common/redis.py
```diff
@@ -70,6 +64,47 @@
...
+OPTS = [
+    cfg.StrOpt('redis_url',
+               default='redis://localhost:6379/',
+               help="""Redis URL
+
+  For example::
+
+    redis://[:password]@localhost:6379?db=0
+
+  We proxy some options to the redis client (used to configure the redis client
```
jd: s/We proxy/Gnocchi passes/
s/redis/Redis/

sileht:
On Wed, Nov 29, 2017 at 02:56:57PM -0800, gord chung wrote:
>chungg commented on this pull request.
>
>i'm ok with this but just to raise discussion, is this the best place vs gnocchi.xyz?
>
>i personally feel like the generated conf file is already very verbose.

I think we should put the whole content of the config file into
gnocchi.xyz, I will do a separate patch to do that.

PR #524, opened 2017-11-30
indexer: fix upgrade when password contains a %

sileht:
oslo.config doesn't support %{} or %() variable interpolation while
ConfigParser does.

So when a password contains a '%', it's fine for oslo.config
But the Python ConfigParser will raise the exception.

Since alembic use ConfigParser and not oslo.config, we have to escape
the url ourself.

This is not a big deal since, we don't want alembic doing variable
interpolation.

### DIFF
--- a/gnocchi/tests/indexer/sqlalchemy/test_utils.py
+++ b/gnocchi/tests/indexer/sqlalchemy/test_utils.py
@@ -1,0 +1,25 @@
+# -*- encoding: utf-8 -*-
+#
+# Copyright © 2017 Red Hat
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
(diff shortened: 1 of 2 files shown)


sileht:
Pull-request updated, HEAD is now 9b3cf5dd22305fd52b9d362b50eea72d645f26bb

gnocchi/tests/indexer/sqlalchemy/test_utils.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2015 eNovance
```
jd: good one

sileht: done

sileht:
Pull-request updated, HEAD is now cb610f7c3934b39afa834654b0ef22519a37b93c

PR #526, opened 2017-12-01
Only install futures on Python 2

jd:
New futures release breaks entirely on Python 3:
 futures requires Python '>=2.6, <3' but the running Python is 3.5.2

Anyway it's a good idea to not install on Python 3.


jd:
@chungg http://logs.openstack.org/82/489682/1/gate/heat-functional-convg-mysql-lbaasv2-py35/d105fa6/logs/devstacklog.txt.gz#_2017-12-01_07_57_34_685

PR #527, opened 2017-12-01
rest: reduce number of index queries

bobuhiro11:
I've noticed that "/v1/batch/resources/metrics/measures" endpoint executes database queries many times. Since ceilometer-agent-notification daemons often post batch data for many resources which have a same metric name, it is a bottleneck.

### DIFF
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -1515,15 +1515,24 @@ class ResourcesMetricsMeasuresBatchController(rest.RestController):
         unknown_metrics = []
         unknown_resources = []
         body_by_rid = {}
+
+        attribute_filter = {"or": []}
+        for original_resource_id, resource_id in body:
+            names = list(body[(original_resource_id, resource_id)].keys())
+            attribute_filter["or"].append({"and": [
+                {"=": {"resource_id": resource_id}},
+                {"in": {"name": names}}]})
+
+        all_metrics = collections.defaultdict(list)
+        for metric in pecan.request.indexer.list_metrics(
+                attribute_filter=attribute_filter):
+            all_metrics[metric.resource_id].append(metric)
+
         for original_resource_id, resource_id in body:
             body_by_rid[resource_id] = body[(original_resource_id,
                                              resource_id)]
             names = list(body[(original_resource_id, resource_id)].keys())
-            metrics = pecan.request.indexer.list_metrics(
-                attribute_filter={"and": [
-                    {"=": {"resource_id": resource_id}},
-                    {"in": {"name": names}},
-                ]})
+            metrics = all_metrics[resource_id]
 
             known_names = [m.name for m in metrics]
             if strtobool("create_metrics", create_metrics):


jd:
@bobuhiro11 thank you! this is a great improvement!

gnocchi/rest/api.py
```diff
@@ -1515,15 +1515,24 @@ def post(self, create_metrics=False):
...
         body_by_rid = {}
+
+        attribute_filter = {"or": []}
+        for original_resource_id, resource_id in body:
+            names = list(body[(original_resource_id, resource_id)].keys())
+            attribute_filter["or"].append({"and": [
+                    {"=": {"resource_id": resource_id}},
+                    {"in": {"name": names}}]})
+
+        all_metrics = {}
```
chungg: should this be pre-populated with all resource_id? i imagine if a resource does not have any metrics, the list_metrics below will never a key for resource and cause it to throw an error below

gnocchi/rest/api.py
```diff
@@ -1515,15 +1515,24 @@ def post(self, create_metrics=False):
...
         for original_resource_id, resource_id in body:
             body_by_rid[resource_id] = body[(original_resource_id,
                                              resource_id)]
             names = list(body[(original_resource_id, resource_id)].keys())
-            metrics = pecan.request.indexer.list_metrics(
-                attribute_filter={"and": [
-                    {"=": {"resource_id": resource_id}},
-                    {"in": {"name": names}},
-                ]})
+            metrics = all_metrics[resource_id]
```
sileht: Agreed with @chungg , this line will fail is no metric have been found for a resource.

gnocchi/rest/api.py
sileht: all_metrics = collections.defaultdict(list), should fix that.

gnocchi/rest/api.py
```diff
@@ -1515,15 +1515,24 @@ def post(self, create_metrics=False):
...
+        for original_resource_id, resource_id in body:
+            names = list(body[(original_resource_id, resource_id)].keys())
+            attribute_filter["or"].append({"and": [
+                    {"=": {"resource_id": resource_id}},
+                    {"in": {"name": names}}]})
+
+        all_metrics = {}
+        for metric in pecan.request.indexer.list_metrics(
+                attribute_filter=attribute_filter):
+            all_metrics.setdefault(metric.resource_id, []).append(metric)
```
sileht: With defaultdict(), all_metrics[metrix.resource_id].append(metric) become enought.

gnocchi/rest/api.py
chungg: yeah, true. that works even before.

bobuhiro11:
Thank you for your review!
I've fixed it.

chungg:
looks good! could you squash your commits into a single commit please.

chungg:
:( pep8

bobuhiro11:
I've fixed for pep8.
Please approve it again.

PR #533, opened 2017-12-02
Only install futures on Python 2

jd:
New futures release breaks entirely on Python 3:
 futures requires Python '>=2.6, <3' but the running Python is 3.5.2

Anyway it's a good idea to not install on Python 3.

(cherry picked from commit 77d1f3b29af325c7f07ba0371ea789d0bccb2311)


sileht:
Unrelated but funny `Upgrading Gnocchi from 3.0.16 to 0.1.0.dev2217`, just due to the missing git history.

PR #536, opened 2017-12-02
rest/prometheus: use operation_timeout to stop retrying

jd:


jd:
Updated commit message

PR #539, opened 2017-12-02
indexer: do not return empty IN statement in QueryFilter

### LINKED ISSUE #530: metricd: IN predicate warnings
seems like we are doing a IN empty sequence comparison and it's probably not something we should do.

master

start metricd

2017-12-01 13:57:29,320 [19366] WARNING  py.warnings: /usr/lib64/python2.7/site-packages/sqlalchemy/sql/default_comparator.py:153: SAWarning: The IN-predicate on "metric.id" was invoked with an empty sequence. This results in a contradiction, which nonetheless can be expensive to evaluate.  Consider alternative strategies for improved performance.
  'strategies for improved performance.' % expr)

not that.

jd:
Fixes #530

### DIFF
--- a/gnocchi/indexer/sqlalchemy.py
+++ b/gnocchi/indexer/sqlalchemy.py
@@ -1131,6 +1131,13 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
         return sort_keys, sort_dirs
 
 
+def _operator_in(field_name, value):
+    # Do not generate empty IN comparison
+    # https://github.com/gnocchixyz/gnocchi/issues/530
+    if len(value):
+        return field_name.in_(value)
+
+
 class QueryTransformer(object):
 
     unary_operators = {
@@ -1160,7 +1167,7 @@ class QueryTransformer(object):
         u"≠": operator.ne,
         u"ne": operator.ne,
 
-        u"in": lambda field_name, values: field_name.in_(values),
+        u"in": _operator_in,
 
         u"like": lambda field, value: field.like(value),
     }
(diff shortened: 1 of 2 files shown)


sileht:
Should this forbidden with voluptuous on rest side ?

jd:
> Should this forbidden with voluptuous on rest side ?

Good question! I don't think so. It's not invalid per say, and it actually "works" with SQLAlchemy, it is just suboptimal and useless. No in Gnocchi's case, it'll optimize it out so we're good. I don't think we need to deny a request because somebody search with a criteria that might false anyway. :)

gnocchi/indexer/sqlalchemy.py
```diff
@@ -1131,6 +1131,13 @@ def _build_sort_keys(sorts, unique_keys):
         return sort_keys, sort_dirs
 
 
+def _operator_in(field_name, value):
+    # Do not generate empty IN comparison
+    # https://github.com/gnocchixyz/gnocchi/issues/530
+    if len(value):
+        return field_name.in_(value)
```
sileht: This looks wrong to me. If I do "{in: {id: []}}, I expect nothing to be returned. With this change the whole table will be returned, no ?

jd:
Added unit test

gnocchi/indexer/sqlalchemy.py
jd: This returns `None` which is equivalent to filter on `False`, i.e., nothing matches.
(that's my understanding since it works :-p)

PR #553, opened 2017-12-07
DNM: rest: port /v1/metric API to Falcon + Marshmallow + apispec (OpenAPI)

jd:
This is a PoC that I made in a few hours to port the metric API to Falcon. I
wanted to have a glimpse on how easy it would be – and it was very easy.

For now, I like how easier the routing is and how cleaner and easier to grasp
it is compared to Pecan. However, I don't see any killer feature yet.

This patch passes all the Gabbi test, and some of the test_rest tests. I don't
see a point yet in making test_rest pass, as it is good enough for what I was
trying to do.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ cover
 .coverage
 dist
 upgrade/
+doc/source/openapi.yml
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -29,8 +29,17 @@ import subprocess
 extensions = [
     'gnocchi.gendoc',
     'sphinxcontrib.httpdomain',
+    'sphinxcontrib.redoc',
     'sphinx.ext.autodoc',
     'reno.sphinxext',
+]
+
+redoc = [
+    {
+        'name': 'Gnocchi API',
+        'page': 'openapi',
+        'spec': 'openapi.yml',
+    },
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/gnocchi/gendoc.py
+++ b/gnocchi/gendoc.py
@@ -227,4 +227,32 @@ def setup(app):
         template = jinja2.Template(f.read())
     with open("doc/source/rest.rst", "w") as f:
         f.write(template.render(scenarios=scenarios))
+
+    # Generate API spec
+    from apispec.core import Path
+    from apispec.lazy_dict import LazyDict
+    from collections import OrderedDict
+
+    from gnocchi.rest import metric
+
+    def represent_dict(dumper, instance):
+        return dumper.represent_mapping('tag:yaml.org,2002:map',
+                                        instance.items())
+
+    if six.PY2:
+        def unicode_representer(dumper, uni):
+            return yaml.ScalarNode(tag=u'tag:yaml.org,2002:str', value=uni)
+
+        yaml.add_representer(unicode, unicode_representer)
+
+    # TODO(jd) Move that into a non-system wide hack in apispec itself
+    # https://github.com/marshmallow-code/apispec/issues/161
+    yaml.add_representer(OrderedDict, represent_dict)
+    yaml.add_representer(LazyDict, represent_dict)
+    yaml.add_representer(Path, represent_dict)
+    yaml.add_representer(Path, represent_dict)
+
+    with open("doc/source/openapi.yml", "w") as f:
+        f.write(yaml.dump(metric.spec.to_dict()))
+
     _RUN = True
--- a/gnocchi/rest/api-paste.ini
+++ b/gnocchi/rest/api-paste.ini
@@ -2,18 +2,21 @@
 use = egg:Paste#urlmap
 / = gnocchiversions_pipeline
 /v1 = gnocchiv1+noauth
+/v1/metric = gnocchiv1metric+noauth
 /healthcheck = healthcheck
 
 [composite:gnocchi+keystone]
 use = egg:Paste#urlmap
 / = gnocchiversions_pipeline
 /v1 = gnocchiv1+keystone
+/v1/metric = gnocchiv1metric+keystone
 /healthcheck = healthcheck
 
 [composite:gnocchi+remoteuser]
 use = egg:Paste#urlmap
 / = gnocchiversions_pipeline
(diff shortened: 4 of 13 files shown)


jd:
@gnocchixyz/devs publishing that for the sake of it and if you are curious.
@chungg if you have 5 minutes to do a benchmark, that could be neat. I tried but did not notice any big improvement, even if the Falcon website says its 10x faster than Pecan 😅

chungg:
Falcon is a cooler name than Pecan. that's a win from my pov. :P

sileht:
I don't really see any code style difference. That's cool if we migrate to this.

jd:
Well, the good thing about Falcon is that it does not do too much. It is way easier to write a REST API the way you want and do weird stuff.

I also took a look at Flask, and I'm not a fan of it for now for REST API. It uses a global object like pecan for requests objects and global variable. Kinda weird how they handle connection to databases. They are a ton of plugins though, but I'm also not sure we could use any because of feature and performance.

Next step is to look into marshmallow and apispec. :)

jd:
Pull-request updated, HEAD is now e2a9b15b70cf309e2ff98287478c2ef24a313f5e

jd:
@gnocchixyz/devs So this patch turned out to grow more than I thought! Here's what I did and what I need you to do.

I only played with `/v1/metrics`.

# Port from Pecan to Falcon
This is the first patch of this PR. It just shows how Falcon works. Falcon is very lightweight and the code base is quite small. Supposedly, it's one of the fastest libraries for doing HTTP. The routing is even compiled to Python AST (so I imagine it's fast :)).

Falcon quite popular, is used by Zaqar, has more than 4k starts on GitHub – Flask has 21k, and Pecan has 60… So I imagine there's a bigger user community than Pecan. We currently have a few limitations with Pecan and I don't think anyone ever like the object based routing.

I don't really see any downside to Falcon at this stage, but more flexibility.

An alternative would be Flask, which is really widespread and has a lot of extension and integration. The problem is that it is less flexible than Falcon, and my initial digging into it did not make me excited. It does too much for us I think.

I'd be in favor of porting the current API or starting a new one with Falcon

# Marshmallow

Marshmallow is basically voluptuous + `jsonify`. It allows to unserialize data, validate them and to serialize them too. It has fields definition, you can define custom fields, etc – see the second patch.

The reason I tested Marshmallow is because of `apispec`, as discussed in the next point: there's a plugin for `apispec` that supports, and there are even extension to integrate it with Flask.

I'm not a fan of Marshmallow currently. There's no gain compared to voluptuous, and I feel like it does too much. Voluptuous just validates the data in a Pythonic way and converts it as needed. The design is smoother than Marshmallow, when validation is separated from data type and where you need to define plenty of classes with custom pre/post methods.

The only thing I like is that you can define schema and use a `context` when using them. In our case, the context could be a config option (e.g. `max_limit`, `creator`), etc.

I don't really want to use Marshmallow I think. I really like how Voluptuous does things and its flexibility.

I might want to add the context support though :)

# apispec

apispec is a library designed by the Marshmallow team which produces OpenAPI schema. It has a plugin system that allows to plug it into schema library (e.g. marshmallow) or WSGI frameworks (e.g. Flask or Bottle).

I've added that support in the 3rd patch.

I've built a small integration with Falcon already in the 3rd patch – it's not that hard to do. Since I used Marshmallow, I used the plugin they provide to generate OpenAPI schema. The truth is that I'm pretty sure I could write an extension to convert Voluptuous to OpenAPI schema – I remember @joehakimrahme already did a protoype for JSONSchema a while back (https://github.com/joehakimrahme/luscious).

Other than APIspec looks promising. It allows to generate a YAML file with an OpenAPI specification. Using sphinx-openapi, we can generate a RST file, but even better, we can use ReDoc which is integrated with Sphinx thanks to sphinx-redoc, and you can have this:

screen shot 2017-12-11 at 18 28 41

You can try it at home, run `tox -e docs && cd doc/build/html && python -m SimpleHTTPServer` and check out http://localhost:8000/openapi.html

# Conclusion

That was a fun couple of days! I want your opinion @chungg and @sileht so when you can please take a coffee and a hour about what you think of all of that.

The goal here is really to generate a proper API spec and clean out the API of old cruft at some point.

jd:
Pull-request updated, HEAD is now a3dd220707b8e69f4e0fe1834c6c929aa7382f4b

chungg:
sorry, i missed this.

> Falcon quite popular, is used by Zaqar, has more than 4k starts on GitHub – Flask has 21k, and Pecan has 60…

^  was hilarious.

i just did a quick google on Falcon. i like how you can just add a route instead of jumping around objects to form a path.

i'm curious why we need a new api? is there something that will change? it seems compatible with uwsgi... is there some formatting change that will happen?

considering i don't think any of us want to support pecan, i think it makes sense to move to something that has a larger community

chungg:
regarding voluptuous v. marshmallow. i'm indifferent. it seems to allow us to validate the value as well so that's good. i think one thing that concerns me is that voluptuous seems to be abandoned as there seems to be requests for a release and nothing is released. is there a performance difference?

regarding apispec. i'm also indifferent. i guess it outputs better than what we have currently.

jd:
> i'm curious why we need a new api? is there something that will change? it seems compatible with uwsgi... is there some formatting change that will happen?

I think it might in the future. I don't know yet. I'm just exploring. I don't want to build a API v2 but you never know.

Falcon is WSGI like all Python frameworks so not a problem. We can replace part of our API endpoint per endpoint like I did with this patch.

> regarding voluptuous v. marshmallow. i'm indifferent. it seems to allow us to validate the value as well so that's good. i think one thing that concerns me is that voluptuous seems to be abandoned as there seems to be requests for a release and nothing is released. is there a performance difference?

voluptuous is not abandoned, I've just add a few PR merged there. I don't know about performances.

chungg:
> voluptuous is not abandoned, I've just add a few PR merged there. I don't know about performances.

oh. i thought it was because i saw https://github.com/alecthomas/voluptuous/issues/313 and no release/response.

jd:
Closing as I guess everyone had enough time to take a look. I'll try to steal some interesting from this in the future!

PR #562, opened 2017-12-09
rest: validate pagination options with voluptuous

jd:
This validate pagination options with voluptuous and remove the side effect of
modifying the query parametres. That avoids doing a copy of the dict each time.


jd:
Pull-request updated, HEAD is now 6d02a136d0b6da352fede7599092f7889a91e651

PR #564, opened 2017-12-09
rest: fail properly if granularity is not a timespan

jd:
The current code will fail with a 500 error.

(cherry picked from commit 63ea0aaa3a6b1c277f132afad157af5547b78b8b)


jd:
Pull-request updated, HEAD is now 1de2ef2b7c410a227fb50b80bf8b663f3dd3787d

PR #566, opened 2017-12-10
rest: do not validate paging keys in metric listing

jd:
The current code in metric listing tries to validate the paging keys with
voluptuous, but they are actually and correctly validated later by
get_pagination_options(). Let's stop trying to be smart here and ignore them.

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -125,7 +125,9 @@ Configuring authentication
 
 The API server supports different authentication methods:
 
-* `basic` (the default) which uses the standard HTTP `Authorization` header.
+* `basic` (the default) which uses the standard HTTP `Authorization` header. By
+  default, only the user `admin` has some special permissions (e.g. create
+  archive policies). The password of the authentication is not used.
 
 * `keystone` to use `OpenStack Keystone`_. If you successfully installed the
   `keystone` flavor using `pip` (see :ref:`installation`), you can set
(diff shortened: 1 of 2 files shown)


jd:
> It's a tiny API change, since we will not return a 400 on invalid option in the query string.

Yes. But I don't think we return 400 for any other URL if they send invalid params, so it seems logical to have the same behavior everywhere, right?

gnocchi/rest/api.py
```diff
@@ -636,15 +636,12 @@ def post(self):
...
-        "limit": six.text_type,
         "name": six.text_type,
         "id": six.text_type,
         "unit": six.text_type,
         "archive_policy_name": six.text_type,
         "status": voluptuous.Any("active", "delete"),
-        "sort": voluptuous.Any([six.text_type], six.text_type),
-        "marker": six.text_type,
-    })
+    }, extra=voluptuous.ALLOW_EXTRA)
```
chungg: this effectively lets user to filter on anything. do we catch all potential query errors? we have any other concerns here like security?

jd:
Better version :)

jd:
Pull-request updated, HEAD is now 2991b637c191e9e4f8dc753dc9b339370b360490

doc/source/install.rst
```diff
@@ -125,17 +125,19 @@ Configuring authentication
...
   `api.auth_mode` to `keystone` to enable Keystone authentication.
   You also need to configure the `keystone_authtoken` section in `gnocchi.conf`
   with the proper value so Gnocchi is able to validate tokens.
 
-* `remoteuser` Gnocchi will look at the HTTP server REMOTE_USER environment
-   variable to get the username. Then the permissions model is the same as the
-   `basic` mode.
+* `remoteuser` where Gnocchi will look at the HTTP server `REMOTE_USER`
+  environment variable to get the username. Then the permissions model is the
+  same as the `basic` mode.
```
chungg: what do these docs have to do with paging?

jd: Oh yeah my bad, I probably fucked up my commits.

PR #575, opened 2017-12-12
incoming: only know about metric id

jd:
The current code passes Metric objects to the incoming subsystem, but it
actually knows only about metric ids. That's why it only uses metric.id
everywhere.

Let's simplify that and make the incoming subsystem completely agnostic with
the representation model of the metric outside of it.


jd:
Pull-request updated, HEAD is now 9c99b4b0148fb96aa7b30e7d7b22b02eb5c883b7

PR #581, opened 2017-12-12
cleanup numpy+list usage

chungg:
- no real point creating a new list with tolist. zip can loop through
array
- to_timestamps requires a list so no point passing creating input
as generator
- numpy suggest to set 'count' when using fromiter. timeit shows
no real diff on small sizes


chungg:
lol, finding optimisations are harder to find now.

Issue #548, opened 2017-12-06
Indexer-less measure processing in metricd


jd:
Currently, metricd `MetricProcessor` needs an indexer connection for the only reason of accessing archive policy information for each metric.

It seems that metricd could not use an indexer connection at all if the archive policy content was duplicated to the metric in the aggregate storage.

This has other implication, especially for metrics using archive policies.
However, in a world where archive policies would be gone (check https://github.com/gnocchixyz/gnocchi/issues/547), that might be feasible.

chungg:
this seems like a potentially useful performance optimisation or just fixing list_metrics in general.

when writing to local disk, list_metrics in process_new_measures can take up 1/3 of the runtime. this is probably less significant if storage is remote but 1/3 of the time seems strange since it's just listing a single metric... granted we're talking single digit millisecond.

attached is a profile pushing 300 datapoints to metric with low policy
[profile-process-new.zip](https://github.com/gnocchixyz/gnocchi/files/1607527/profile-process-new.zip)

chungg:
did you have a plan for this?

- add the archive policy name to each measures obj and cache the definition in each worker?
- add the definition to the measures obj itself?

i guess in both cases it gets kinda funky when you start updating timespan of policy

jd:
My think is rather to dump archive policy requirement before doing this, i.e. #547.

Once this one is done, each metric needs to store its own aggregation definition and the question is where to store it. If you store it in the storage engine, then you solve this issue.

I don't have more details for now so there might be blockers (?).

Issue #175, opened 2017-07-03
Allow to not use archive policies


jd:
Archive policies are very useful and handy. However they also limit what a user can do with the metrics. If an application wants to create a metric and really knows what kind of aggregation methods it needs, there's no way for it to bypass the archive policy mechanism.

Historically, archive policies were implement as a mechanism to limit what a user can do and as a template, but it seems that to me that this limitation does not make much sense anymore and is actually preventing smarter definition of metrics.

chungg:
you know how you'd implement this? seems like we'd need a new storage format. if we capture nanosecond raw data, we can't create 3600point blocks anymore or else that's 3.6ms range per obj?

jd:
Not sure I understand your concern, but this is just an API change mainly. It's just to allow metric to be created with:

```
POST /v1/metric

{
  archive_policy: {
    aggregation_methods: [mean, min, max],
    definition: {granularity: 5 points: 100}
  }
}
```

Instead of only supporting the current `archive_policy_name` parameter.

chungg:
oh. so like using a policy that is not necessarily formally created in gnocchi.

i thought this meant, we had no aggregation and just storing raw data (which isn't really possibly... we can only store to second precision currently)

jd:
Indeed that's not what I meant. Storing raw data could be another issue to open, but since I don't see any scalable solution to that currently, I'm not talking about i. :)

PR #583, opened 2017-12-13
Pull request for no-ap-in-storage

jd:
storage: merge GranularityDoesNotExist and AggregationDoesNotExist
storage: remove double check of granularity

### DIFF
--- a/gnocchi/storage/__init__.py
+++ b/gnocchi/storage/__init__.py
@@ -59,23 +59,25 @@ class MetricDoesNotExist(StorageError):
 class AggregationDoesNotExist(StorageError):
     """Error raised when the aggregation method doesn't exists for a metric."""
 
-    def __init__(self, metric, method):
+    def __init__(self, metric, method, granularity):
         self.metric = metric
         self.method = method
-        super(AggregationDoesNotExist, self).__init__(
-            "Aggregation method '%s' for metric %s does not exist" %
-            (method, metric))
-
-
-class GranularityDoesNotExist(StorageError):
-    """Error raised when the granularity doesn't exist for a metric."""
-
-    def __init__(self, metric, granularity):
-        self.metric = metric
         self.granularity = granularity
-        super(GranularityDoesNotExist, self).__init__(
-            "Granularity '%s' for metric %s does not exist" %
-            (utils.timespan_total_seconds(granularity), metric))
+        super(AggregationDoesNotExist, self).__init__(
+            "Aggregation method '%s' at granularity '%s' "
+            "for metric %s does not exist" %
+            (method, utils.timespan_total_seconds(granularity), metric))
+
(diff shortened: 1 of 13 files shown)


chungg:
i don't really understand the second commit. why don't we have enough info to detect whether granularity or aggregation_method exists?

jd:
> i don't really understand the second commit. why don't we have enough info to detect whether granularity or aggregation_method exists?

This is how most storage driver works: for example the file driver tries to open the file for "mean/300s" and fails. It does not know that's because the metric does not exist or because last is not in the metric AP or because 300 is not in the metric AP. Therefore it checks for the directory of the metric to check if the metric exists, and then raises `AggregationDoesNotExist` whereas it might be `GranularityDoesNotExist`.

jd:
Pull-request updated, HEAD is now 2ec6f26a8dc84f9904335ec66dee30fd6ab687d1

gnocchi/storage/__init__.py
```diff
@@ -198,7 +200,9 @@ def get_measures(self, metric, from_timestamp=None, to_timestamp=None,
         :param resample: The granularity to resample to.
         """
         if aggregation not in metric.archive_policy.aggregation_methods:
-            raise AggregationDoesNotExist(metric, aggregation)
+            if granularity is None:
+                granularity = metric.archive_policy.definition[0].granularity
+            raise AggregationDoesNotExist(metric, aggregation, granularity)
```
chungg: ok. but it literally validates this right here and knows exactly if it's the granularity or the aggregate that is wrong... imo, if a storage driver does not find the actual corresponding file/object because it's yet to receive data, it should not return an error but return an empty timeseries becaues that's exactly what it is.

that said, i imagine this is because you are trying to address #548? is this the only way to handle it?

jd: No, I'm actually targeting #175. For that I need to remove archive policy usage from the storage driver and move the actual detection of invalid requests to the API layer.

chungg: i see. so how do you envision 175 working? there is no validation and the user can submit any combination of granularity+aggregate? because if there's validation at some point prior to this, i'm thinking it's still better to just return empty ts rather than error.

jd: >  the user can submit any combination of granularity+aggregate?

Exactly.

> i'm thinking it's still better to just return empty ts rather than error.

Actually, it's better to raise an error in storage and leave the layer above (e.g. REST API) decides what behaviour it wants to implement

That being said, this PR do not change the current behaviour of the API or even of storage – it merges 2 different exception into one. I feel like we're talking about something that is not there (yet).

chungg: that's fair. i guess as long as we eventually move this validation stuff from storage then.

PR #587, opened 2017-12-14
storage: rework search_value

jd:
This cleans a bit the old search_value/find_measure storage API to do most of
the API related work in the API, and only do the minimum work in the storage
part.

### DIFF
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -1467,8 +1467,6 @@ class SearchMetricController(rest.RestController):
     @pecan.expose('json')
     def post(self, metric_id, start=None, stop=None, aggregation='mean',
              granularity=None):
-        granularity = [utils.to_timespan(g)
-                       for g in arg_to_list(granularity or [])]
         metrics = pecan.request.indexer.list_metrics(
             attribute_filter={"in": {"id": arg_to_list(metric_id)}})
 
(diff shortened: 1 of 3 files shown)


gnocchi/rest/api.py
```diff
@@ -1484,20 +1482,32 @@ def post(self, metric_id, start=None, stop=None, aggregation='mean',
...
+            for metric in metrics:
+                if granularity is None:
+                    granularity = [d.granularity
+                                   for d in metric.archive_policy.definition]
+                results[str(metric.id)] = []
+                for g in sorted(granularity):
+                    results[str(metric.id)].extend(
+                        pecan.request.storage.find_measure(
+                            metric, predicate, g, aggregation,
+                            start, stop))
```
chungg: you lost concurrency in this patch. what's up?

jd: Yeah, I did not bother re-implementing it as it complicates the code.

I'm pretty sure nobody ever used that feature anyway. If you really want I can add a few lines to parallelize stuff again.

chungg: well it's on by default...

i'd prefer we keep it. personal experience, threading in python works well as long as you know it's I/O.

jd: :(

chungg: well, i mean if you can argue that there isn't much to make concurrent than that's valid but right your technically making this slower and i'm not sure 'no one uses it' is a valid reason.

jd:
Parallel readded

gnocchi/rest/api.py
```diff
@@ -1484,20 +1482,33 @@ def post(self, metric_id, start=None, stop=None, aggregation='mean',
...
+            for metric in metrics:
+                if granularity is None:
+                    granularity = [d.granularity
+                                   for d in metric.archive_policy.definition]
+                results[str(metric.id)] = []
+                for r in utils.parallel_map(
+                        pecan.request.storage.find_measure,
+                        ((metric, predicate, g, aggregation,
+                          start, stop)
+                         for g in sorted(granularity))):
```
chungg: we're needlessly sorting repeatedly if granularity is provided. and it's done in multithread so it's not guaranteed to do return sorted.

also in original code it's sorting in descending order.

jd: > and it's done in multithread so it's not guaranteed to do return sorted

It's not guaranteed to be executed in order, but it's returned ordered.

jd:
Pull-request updated, HEAD is now fd41974ffc5bd2df8c9bfed62b4ab46cdf4aea59

PR #589, opened 2017-12-15
rest: narrow down the response attributes

### LINKED ISSUE #545: rest: narrow down the response attributes
master branch (commit 7892684dde395dd3ccec67a9cb172659975d74a8)

```bash
$ cd gnocchi-docker
$ source devel.env
$ docker-compose up
```
Templating of gnocchi's grafana-datasource converts a query option (ex. `resources(instance, host, {})`) to a query(ex. `{}`) and post it to `/v1/search/resource/instance` endpoint. The response size of this endpoint will be too large (ex. dozens of MB), if the number of resources increases (ex. 1,000 instances). So, I'd like to modify gnocchi-api to response only specified attributes such as `host` and `flavor_id` as below:

```bash
# Request
POST /v1/search/resource/instance?details=true&attributes=host,id,flavor_id HTTP/1.1
Content-Length: XX

{
  "=": {
    "user_id": "BD3A1E52-1C62-44CB-BF04-660BD88CD74D"
  }
}

# Response
HTTP/1.1 200 OK
Content-Length: XXXX

[
  {
    "flavor_id": "2",
    "host": "compute1",
    "id": "6868da77-fa82-4e67-aba9-270c5ae8cbca"
  }
]
```
I would like to get your feedback.

nmi:
To reduce the response size of search/list queries for gnocchi's grafana-datasource, I've added an `attrs` parameter.
```
GET /v1/resource/generic?attrs=id,type,user_id HTTP/1.1
POST /v1/search/resource/generic?attrs=id,type,user_id HTTP/1.1
```

Fixes #545

### DIFF
--- a/gnocchi/indexer/sqlalchemy_base.py
+++ b/gnocchi/indexer/sqlalchemy_base.py
@@ -198,7 +198,7 @@ class ResourceType(Base, GnocchiBase, resource_type.ResourceType):
 
 
 class ResourceJsonifier(indexer.Resource):
-    def jsonify(self):
+    def jsonify(self, attrs=None):
         d = dict(self)
         del d['revision']
         if 'metrics' not in sqlalchemy.inspect(self).unloaded:
@@ -212,7 +212,10 @@ class ResourceJsonifier(indexer.Resource):
                 self.creator.partition(":")
             )
 
-        return d
+        if attrs:
+            return {key: val for key, val in d.items() if key in attrs}
+        else:
+            return d
 
 
 class ResourceMixin(ResourceJsonifier):
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -188,12 +188,30 @@ def get_header_option(name, params):
                      options.get(name, params.get(name, 'false')))
 
 
+def get_header_option_array(name, params):
+    type, options = werkzeug.http.parse_options_header(
+        pecan.request.headers.get('Accept'))
+    header_option = options.get(name, None)
+    post_option = params.get(name, None)
+
+    if post_option:
+        return arg_to_list(post_option)
+    elif header_option:
+        return header_option.split('+')
+    else:
+        return None
+
+
 def get_history(params):
     return get_header_option('history', params)
 
 
 def get_details(params):
     return get_header_option('details', params)
+
+
+def get_json_attrs(params):
+    return get_header_option_array('attrs', params)
 
 
 def strtobool(varname, v):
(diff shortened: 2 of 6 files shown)


gnocchi/tests/test_rest.py
```diff
@@ -1591,6 +1591,33 @@ def test_search_resources_with_details_via_accept(self):
                 "/v1/search/resource/generic",
                 headers={"Accept": "application/json; details=true"}))
 
+    def _do_test_list_resources_with_attrs(self, request):
```
chungg: this would be better as gabbi test but meh.

can you also verify: ?attrs=id&attrs=user_id works. i'm more accustomed to this formatting and i think most frameworks handle this natively.

also, you should add negative tests to tests:
- if you give an attribute that does not exist.
- if you give no attrs, you get all attributes

jd: yes please gabbi :(

nmi:
@chungg Thank you for your reply!
I added some tests using gabbi.

gnocchi/rest/api.py
```diff
@@ -1380,6 +1404,10 @@ def _search(self, **kwargs):
             details=details,
             history=history,
             **pagination_opts)
+
+        for r in resources:
+            r.set_json_attrs(json_attrs)
```
jd: So I put my functional hat and I don't like that. :(

Can we just `return [r.jsonify(json_attrs) for resources]` instead?

nmi:
@sileht Thanks! This is an adequate improvement for my workload.
@jd Thanks to your view, this patch became simple.

gnocchi/indexer/sqlalchemy_base.py
```diff
@@ -198,7 +198,7 @@ def to_baseclass(self):
 
 
 class ResourceJsonifier(indexer.Resource):
-    def jsonify(self):
+    def jsonify(self, json_attrs=None):
```
jd: The name of the method is `jsonify` so calling the argument `attrs` should be enough. But not a big deal. :)

gnocchi/rest/api.py
```diff
@@ -1392,7 +1411,8 @@ def _search(self, **kwargs):
     @pecan.expose('json')
     def post(self, **kwargs):
         try:
-            return self._search(**kwargs)
+            json_attrs = get_json_attrs(kwargs)
```
jd: I'd probably put that outside of the `try/except` block for clarity. Otherwise one could think that this could raise an e.g. `IndexerException`, which should not be the case.

gnocchi/rest/api.py
```diff
@@ -188,6 +188,20 @@ def get_header_option(name, params):
                      options.get(name, params.get(name, 'false')))
 
 
+def get_header_option_array(name, params):
+    type, options = werkzeug.http.parse_options_header(
+        pecan.request.headers.get('Accept'))
```
jd: I don't see a Gabbi test for the `Accept` case – i'm not even sure I understand how it works. Can you add one?

nmi: I've added a test for `Accept` case.

nmi:
I've rebase this branch on master branch.

Issue #585, opened 2017-12-14
Gnocchi wsgi app should always success to start


sileht:
### How to reproduce your problem

make coordinator, or indexer, or incoming or storage unreachable and gnocchi wsgi app can't be loaded

### What is result that you expected

I want gnocchi wsgi app always be able to be load if a backend is unreachable. And API returning a 500 or 503 on API call until everything is back. Exactly like when backend service are unreachable after the wsgi app have started.

PR #591, opened 2017-12-18
rest: lazy load backends

sileht:
When of the backend are unreachable, the error is differ depending
of if the issue occurs on startup or at runtime.

This change ensures any error are raised at runtime.

Closes-bug: #585

### DIFF
--- a/gnocchi/rest/app.py
+++ b/gnocchi/rest/app.py
@@ -15,6 +15,7 @@
 # under the License.
 import os
 import pkg_resources
+import threading
 import uuid
 
 import daiquiri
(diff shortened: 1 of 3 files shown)


sileht:
Pull-request updated, HEAD is now a7d08d3f99cb27530b3860236dec77b1b8c41156

sileht:
Pull-request updated, HEAD is now afcaabba39fe44e78d51b011f7aa9fb63afca88a

gnocchi/rest/app.py
```diff
@@ -42,26 +43,49 @@
...
+        'indexer': (threading.Lock(), gnocchi_indexer.get_driver),
+    }
+
+    def _lazy_load(self, name):
+        # NOTE(sileht): We don't care about raise error here, if something
+        # fail, this will just raise a 500, until the backend is ready.
+        if name not in self.backends:
+            lock, loader = self.BACKEND_LOADERS[name]
+            with lock:
+                # Recheck, maybe it have been created in the meantime.
```
jd: has

PR #595, opened 2017-12-18
optimise aggregations

chungg:
- vectorize std. ~35x faster
- use bincount+weights to compute sum. ~1.7x faster
- use sum to get mean. ~1.6x faster
- insert creates a copy. just flip array to get first. ~1.1x faster

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -124,10 +124,14 @@ class GroupedTimeSeries(object):
                                                  return_counts=True)
 
     def mean(self):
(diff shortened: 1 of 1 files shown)


sileht:
When I have initially moved to numpy/scipy I have added at one test per aggregation, to ensure I generate the same output as pandas: https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/tests/test_carbonara.py#L279-L319

It's light but it's there.

chungg:
yeah, i was basing it off those tests... there might be some edge case (like empty series) that aren't tested.

PR #601, opened 2017-12-27
tests: Fix resource type with optional uuid

jd:
tests: Fix resource type with optional uuid

This changes adds all missing tests for PATCH an optional attributes of
a resource type (Only string type was having tests).

And fix the bug where False is returned instead of None for uuid.

Closes-bug: #616
(cherry picked from commit 13a0123)


jd:
Pull-request updated, HEAD is now 19441295b855b2f0956430e2aebd44daddbfe224

jd:
Remove second commit

jd:
Pull-request updated, HEAD is now 58aa38c6a5a7483e65f11fcf3ad915e6e735d828

PR #602, opened 2018-01-02
speed up median aggregate

chungg:
- find the position of max value of each range
- calculate distance to middle based on range counts
- get floor/ceiling of middle (same if odd count), add, and divide.
- ~1.9x faster in random case

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -148,7 +148,18 @@ class GroupedTimeSeries(object):
         return make_timeseries(self.tstamps, values)
 
     def median(self):
-        return self._scipy_aggregate(ndimage.median)
+        ordered = self._ts['values'].argsort()
+        uniq_inv = numpy.repeat(numpy.arange(self.counts.size), self.counts)
+        max_pos = numpy.zeros(self.tstamps.size, dtype=numpy.int)
+        max_pos[uniq_inv[ordered]] = numpy.arange(self._ts.size)
+        # TODO(gordc): can use np.divmod when centos supports numpy 1.13
+        mid_diff = numpy.floor_divide(self.counts, 2)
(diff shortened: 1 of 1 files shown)


gnocchi/carbonara.py
```diff
@@ -148,7 +148,16 @@ def max(self):
         return make_timeseries(self.tstamps, values)
 
     def median(self):
-        return self._scipy_aggregate(ndimage.median)
+        ordered = self._ts['values'].argsort()
+        uniq_inv = numpy.repeat(numpy.arange(self.counts.size), self.counts)
+        max_pos = numpy.zeros(self.tstamps.size, dtype=numpy.int)
+        max_pos[uniq_inv[ordered]] = numpy.arange(self._ts.size)
+        mid_diff, odd = numpy.divmod(self.counts, 2)
```
sileht: Can we use "floor_divide(self.count, 2), reminder(self.count, 2)" for numpy < 1.13 and avoid to raise the requirement ? Centos/RHEL RDO ships only numpy 1.11 for queens

chungg: kk. i'll add it as a todo.

jd: hm we can't add this as a todo if we don't want to depends on 1.9.0 and not 1.13.0, no?

chungg: NOTE then? i'm not sure what you mean to be honest.

jd: @chungg Sorry I misunderstood you. I thought you were only gonna add a TODO and not change the code at all lol. That's all good, thanks!

chungg: ah. i see :) i am that lazy so it was a good assumption.lol

PR #611, opened 2018-01-08
optimise percentile computation with pure numpy

chungg:
i believe i figured out how percentile is computed. this does
something very similar to median (for obvious reasons). at highlevel:
- get the index the percentile would land on and assume it falls
between indices
- take the values of the indices it falls between and figure out
weight of each value.
- handle the percentiles that fall on an exact index.

this is ~30x better. it's actually a bit slower (3%-5%?) than the
median computation so it's debatable if we want to remove median code.

drop scipy. also remove numpy.lib.recfunctions as it doesn't seem to
used anywhere.

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -26,8 +26,6 @@ import time
 
 import lz4.block
 import numpy
-import numpy.lib.recfunctions
-from scipy import ndimage
 import six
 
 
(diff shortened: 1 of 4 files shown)


chungg:
hmm... i need to test this. this might mean my median patch is wrong too. :/

chungg:
updated but requires median patch to leverage test.

PR #615, opened 2018-01-08
service: fix utils.parallel_map number of workers setting

### LINKED ISSUE #605: parallel_operations doesn't honour the provided value
master

set parallel_operations or aggregation_workers_number to any number other than the cpu core count

https://github.com/gnocchixyz/gnocchi/blob/2dbb4bcb69ceb23e1d79baed28699da6b596e9f9/gnocchi/utils.py#L305 is always the cpu core count

the value i set in conf. it does say the correct value when service is launched and all the option values are printed out in debug. it's just that the value there is not what is eventually used.

jd:
Naming variable is hard

Fixes #605


jd:
Enhanced tests

chungg:
thanks!

PR #622, opened 2018-01-11
doc: disable build of stable/3.?

### LINKED ISSUE #614: Doc for Gnocchi 3.0 and 3.1 is not displayed on gnocchi.xyz
This supposedly works.

Build log https://gnocchi.xyz/build-logs/gnocchi.log

jd:
Those builds are currently broken due to incompatibility between the lz4
dependencies of master and stable/3.?.

While it would be boring but fixable, I imagine it's now better to just ignore
those doc. 3.1 is now more than 1 year old, we can expect most people to
upgrade by now.

Fixes #614

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -182,7 +182,7 @@ scv_sort = ('semver',)
 scv_show_banner = True
 scv_banner_main_ref = 'stable/4.1'
 scv_priority = 'branches'
-scv_whitelist_branches = ('master', '^stable/([3-9]\.)')
+scv_whitelist_branches = ('master', '^stable/([4-9]\.)')
 scv_whitelist_tags = ("^$",)
 
 here = os.path.dirname(os.path.realpath(__file__))


chungg:
did we have v3 specific configuration info? like how to tweak settings to fix scheduling issue?

chungg:
i'm hesitant because we really only released v4 7 months ago (https://github.com/gnocchixyz/gnocchi/releases/tag/4.0.0). but i also don't know how to fix this.lol could we just cap the lz4 requirement in doc build?

jd:
> did we have v3 specific configuration info? like how to tweak settings to fix scheduling issue?

No, though latest patches mitigated those.

> could we just cap the lz4 requirement in doc build?

Well we use the same venv to build the doc for all versions and we have lz4<9 and lz4>=9. The last one is used, so `stable/3.?` is failing. If we pick lz4<9, then master will fail.

I don't think there's a way to fix that with the Sphinx extension we use. The only fix I see is to make stable/3.? work with lz4 <9 AND >=9. Which I don't really want to do…

chungg:
> Well we use the same venv to build the doc for all versions and we have lz4<9 and lz4>=9. The last one is used, so stable/3.? is failing. If we pick lz4<9, then master will fail.

well, there was a version that supported both. it had both lz4.block.* and lz4.loads/dumps... if we just locked the doc builds to that for now...

because right now, iiuc, ubuntu is currently on 3.1 for pike (https://launchpad.net/ubuntu/+source/gnocchi) and

jd:
Merged #625 instead

PR #623, opened 2018-01-11
Pull request for expose-workers-api

jd:
metricd: use human readable member id
rest: export the available metricd processors in status
rest: store coordinator in pecan.request

### DIFF
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -27,10 +27,12 @@ import pyparsing
 import six
 from six.moves.urllib import parse as urllib_parse
 import tenacity
+import tooz
 import voluptuous
 import werkzeug.http
 
 from gnocchi import archive_policy
+from gnocchi.cli import metricd
 from gnocchi import incoming
 from gnocchi import indexer
 from gnocchi import json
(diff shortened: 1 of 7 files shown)


gnocchi/tests/functional/gabbits/base.yaml
```diff
@@ -132,6 +132,7 @@ tests:
     authorization: "basic YWRtaW46"
   response_json_paths:
     $.storage.`len`: 2
+    $.metricd.`len`: 1
```
chungg: should this be metricd.processors.`len`?

gnocchi/rest/api.py
```diff
@@ -1967,6 +1974,11 @@ def get(details=True):
         report_dict = {"storage": {"summary": report['summary']}}
         if 'details' in report:
             report_dict["storage"]["measures_to_process"] = report['details']
+        report_dict['metricd'] = {}
+        if members_req:
+            report_dict['metricd']['processors'] = members_req.get()
```
chungg: tbh, i thought `processors` meant total core count. is there a more standard name we've been using? agent? daemon? or are we using `processor`  name going forward?

jd:
```class MetricProcessor(MetricProcessBase):
    name = "processing"
    GROUP_ID = b"gnocchi-processing"
```

I'd suggest we change name to processor in gnocchi.cli.metricd too then.

gnocchi/tests/functional/gabbits/base.yaml
jd: No, because the len of `metricd.processors` depend on the number of workers and in this case depends on the coordination driver, which is likely to be e.g. SQL to it'd be NULL.

I"m just checking the key `processors` is here, like we do L134.

PR #624, opened 2018-01-11
Introduce Aggregation object

jd:
This introduces a new Aggregation objects (a named tuple) that stores the
information needed to retrieve any aggregation from the storage driver.

The goal is make most methods accept this Aggregation object as argument
instead of a list of kwargs and to replace the archive policy usage in the
storage driver.


jd:
Fix typo

chungg:
i'm sort of curious how you plan on expanding this? is this suppose to be a step to removing indexer? i'm not really sure where this is going so i'd like to learn more :)

gnocchi/storage/__init__.py
```diff
@@ -589,9 +586,12 @@ def _map_add_measures(bound_timeserie):
 
     def find_measure(self, metric, predicate, granularity, aggregation="mean",
                      from_timestamp=None, to_timestamp=None):
+        agg = metric.archive_policy.get_aggregation(aggregation, granularity)
+        if agg is None:
+            raise AggregationDoesNotExist(metric, aggregation, granularity)
```
chungg: should get_aggregation raise this?

chungg:
> Now, this is part of my branch for removing archive policy knowledge from storage. That'll allow to have metrics created without using an archive policy, but just a list of Aggregation to use.

i'll ask this now, but part of the processing code leverages the fact that the model currently groups all methods associated to a granularity in one so we don't have to regroup the same stuff[1]. if we have it separated, do you see any issues?

[1] https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/storage/__init__.py#L559-L562

jd:
> if we have it separated, do you see any issues?

The common key here is granularity, so it's easy to groupby `Aggregation` objects even if there's no AP in the mix.

Issue #616, opened 2018-01-08
PATCH resource type error 500


sileht:
### Which version of Gnocchi are you using

-e git+https://github.com/gnocchixyz/gnocchi@cb2a4dfb1a1e2cab1a69d4f29b97a19ed4b280a0#egg=gnocchi

### How to reproduce your problem

Jan 08 14:13:39.860542 ubuntu-xenial-rax-iad-0001732328 devstack@gnocchi-api.service[13866]: [pid: 13869|app: 0|req: 24/48] 104.239.168.68 () {60 vars in 1263 bytes} [Mon Jan  8 14:13:39 2018] PATCH /metric/v1/resource_type/volume => generated 0 bytes in 49 msecs (HTTP/1.1 500) 0 headers in 0 bytes (0 switches on core 0)

Jan 08 14:13:39.851908 ubuntu-xenial-rax-iad-0001732328 devstack@gnocchi-api.service[13866]: 2018-01-08 14:13:39,851 [13869] CRITICAL root: Traceback (most recent call last):
Jan 08 14:13:39.852200 ubuntu-xenial-rax-iad-0001732328 devstack@gnocchi-api.service[13866]:   File "/usr/local/lib/python2.7/dist-packages/webob/dec.py", line 131, in __call__
Jan 08 14:13:39.852413 ubuntu-xenial-rax-iad-0001732328 devstack@gnocchi-api.service[13866]:     resp = self.call_func(req, *args, **self.kwargs)
Jan 08 14:13:39.852641 ubuntu-xenial-rax-iad-0001732328 devstack@gnocchi-api.service[13866]:   File "/usr/local/lib/python2.7/dist-packages/webob/dec.py", line 196, in call_func
Jan 08 14:13:39.852844 ubuntu-xenial-rax-iad-0001732328 devstack@gnocchi-api.service[13866]:     return self.func(req, *args, **kwargs)
Jan 08 14:13:39.853045 ubuntu-xenial-rax-iad-0001732328 devstack@gnocchi-api.service[13866]:   File "/usr/local/lib/python2.7/dist-packages/oslo_middleware/base.py", line 131, in __call__
Jan 08 14:13:39.853245 ubuntu-xenial-rax-iad-0001732328 devstack@gnocchi-api.service[13866]:     response = req.get_response(self.application)
[...]

PR #631, opened 2018-01-12
tests: Fix resource type with optional uuid

jd:
This changes adds all missing tests for PATCH an optional attributes of
a resource type (Only string type was having tests).

And fix the bug where False is returned instead of None for uuid.

Closes-bug: #616
(cherry picked from commit 13a0123fbc5087f9ed8c4c0ebb6348a85ab71a67)
(cherry picked from commit d2ed3846b7e99b59de96827afc71a27b031225c6)


jd:
Pull-request updated, HEAD is now 5ac4d843f7afc6ef2b29c52085a0fa94599a4108

PR #654, opened 2018-01-16
 cleanup only when new object

### LINKED ISSUE #608: avoid/minimise object scanning on write
when storing measures, we run _list_split_keys_for_metric, which does a scan of objects, for each single granularity and aggregate to figure out what to delete/compress. this sucks.

ideally, we shouldn't call _list_split_keys_for_metric at all as we can easily just compute the keys and delete/compress until the object fails to delete/compress. the problem is that certain drivers s3 do not fail if you attempt to delete something that doesn't exist. i'm not sure how ceph and s3 behave. maybe @sileht would know.

that said, if we are forced to do a _list_split_keys_for_metric scan, we should do it once and rather than for ever single granularity x aggregation permutation as _list_split_keys_for_metric already is capable of returning everything, we're just filtering in post (in most cases)

chungg:
currently, we scan for objects to cleanup or to rewrite. this is
unnecessary i/o as the policies requires, at least, 1hrs worth of
data (at high policy), and at lower policies requires days worth of
data. it does not make sense for us to check for objects to clean
for every measure.

this leverages the fact we can compute when we need to rewrite
objects without scanning by computing new split vs old split. using
this knowledge, we scan only if we know a new object is created.
only in this case will we check to see what objects need to be
deleted or rewritten.

this will still end up scanning sometimes even though no work is
needed for non-append drivers (see: not ceph) but this should lower
the false scans significantly.

Fixes #608


chungg:
reworded commit message. it didn't make sense.

chungg:
hmm.. testing in devstack. i think the splitkey comparison might be broken :/

PR #655, opened 2018-01-17
batch: allow to pass archive policy

sileht:
This change allow to pass the archive policy when we batch
measures for metrics.

### DIFF
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -148,18 +148,24 @@ def deserialize(expected_content_types=None):
     return params
 
 
-def validate(schema, data, required=True):
+def validate(schema, data, required=True, detailed_exc=False):
     try:
         return voluptuous.Schema(schema, required=required)(data)
     except voluptuous.Error as e:
-        abort(400, "Invalid input: %s" % e)
+        if detailed_exc:
+            abort(400, {"cause": "Invalid input",
+                        "detail": str(e)})
+        else:
+            abort(400, "Invalid input: %s" % e)
 
 
 def deserialize_and_validate(schema, required=True,
-                             expected_content_types=None):
+                             expected_content_types=None,
+                             detailed_exc=False):
     return validate(schema,
                     deserialize(expected_content_types=expected_content_types),
-                    required)
+                    required,
+                    detailed_exc)
 
 
 def Timespan(value):
@@ -434,14 +440,14 @@ def MeasuresListSchema(measures):
     try:
         times = utils.to_timestamps([m['timestamp'] for m in measures])
     except TypeError:
-        abort(400, "Invalid format for measures")
+        raise voluptuous.Invalid("unexpected measures format")
     except ValueError as e:
-        abort(400, "Invalid input for timestamp: %s" % e)
+        raise voluptuous.Invalid("unexpected timestamp '%s'" % e)
 
     try:
         values = [float(i['value']) for i in measures]
     except Exception:
-        abort(400, "Invalid input for a value")
+        raise voluptuous.Invalid("unexpected measures value")
 
     return (incoming.Measure(t, v) for t, v in six.moves.zip(times, values))
 
@@ -462,15 +468,13 @@ class MetricController(rest.RestController):
         self.enforce_metric("get metric")
         return self.metric
 
-    @pecan.expose()
+    @pecan.expose('json')
     def post_measures(self):
         self.enforce_metric("post measures")
-        params = deserialize()
-        if not isinstance(params, list):
-            abort(400, "Invalid input for measures")
-        if params:
-            pecan.request.incoming.add_measures(
-                self.metric.id, MeasuresListSchema(params))
+        measures = deserialize_and_validate(MeasuresListSchema,
+                                            detailed_exc=True)
+        if measures:
+            pecan.request.incoming.add_measures(self.metric.id, measures)
         pecan.response.status = 202
 
     @pecan.expose('json')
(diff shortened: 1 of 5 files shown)


sileht:
Pull-request updated, HEAD is now 3192bd07a90df570641c8eaec61bac4b90954f24

sileht:
Pull-request updated, HEAD is now 17b94ac668c080de0c9647262e4a4df9d5973cc8

sileht:
Pull-request updated, HEAD is now 9a4efb495405317349a12f5b216ae1715bc3b1e5

sileht:
Pull-request updated, HEAD is now 74200008ce84eaf1447b3120f31c21a3913024fe

sileht:
Pull-request updated, HEAD is now fd5fa4010de8b630e56c1bb1c99c38033c69571e

sileht:
Pull-request updated, HEAD is now 708aa13536ccc5cf9206901419ba64edae6f3ab6

sileht:
Pull-request updated, HEAD is now 459736f118c15e4d03dcc482c27999f08c0721b3

sileht:
Pull-request updated, HEAD is now 875b8919b2a85edab1ead7b12a85095655627984

sileht:
Pull-request updated, HEAD is now 0871564844b2eab54e220324f1e2dc80634d2d91

gnocchi/rest/api.py
```diff
@@ -466,11 +466,12 @@ def get_all(self):
...
         params = deserialize()
-        if not isinstance(params, list):
-            abort(400, "Invalid input for measures")
         if params:
-            pecan.request.incoming.add_measures(
-                self.metric.id, MeasuresListSchema(params))
+            try:
+                measures = MeasuresListSchema(params)
+            except voluptuous.Error as e:
+                abort(400, "Invalid input: %s" % e)
```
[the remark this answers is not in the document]
sileht: No their are not used. I changed that.

gnocchi/rest/api.py
```diff
@@ -148,18 +153,23 @@ def deserialize(expected_content_types=None):
...
 
 
-def validate(schema, data, required=True):
+def validate(schema, data, required=True, detailed_exc=False):
     try:
         return voluptuous.Schema(schema, required=required)(data)
     except voluptuous.Error as e:
-        abort(400, "Invalid input: %s" % e)
+        if detailed_exc:
+            abort(400, e)
```
jd: But this is gonna return a string right, because `voluptuous.Error` has no jsonify()? This does not seem like the final goal we want to achieve.

sileht: abort() method build the json for any voluptuous error now.

sileht: I see, I will move the json build here, it make more sense

sileht:
Pull-request updated, HEAD is now de28beca957f2d0de520feaa3c17f6bd52f495b5

gnocchi/rest/api.py
```diff
@@ -1542,16 +1546,32 @@ def post(self, metric_id, start=None, stop=None, aggregation='mean',
...
 
 class ResourcesMetricsMeasuresBatchController(rest.RestController):
+
+    @staticmethod
+    def BackwardCompatibleMeasuresList(v):
+        v = voluptuous.Schema(
+            voluptuous.Any(MeasuresListSchema,
+                           {voluptuous.Optional("archive_policy_name"):
+                            six.text_type,
+                            voluptuous.Optional("unit"):
```
chungg: is unit here because in case it needs to be created?

i'll be honest, i don't really like the unit attribute... to me it's equivalent to having a description attribute... but oh well.

gnocchi/rest/api.py
```diff
@@ -148,18 +148,24 @@ def deserialize(expected_content_types=None):
     return params
 
 
-def validate(schema, data, required=True):
+def validate(schema, data, required=True, detailed_exc=False):
     try:
         return voluptuous.Schema(schema, required=required)(data)
     except voluptuous.Error as e:
```
jd: This should only catch `voluptuous.Invalid` so then…

gnocchi/rest/api.py
```diff
@@ -148,18 +148,24 @@ def deserialize(expected_content_types=None):
...
-        abort(400, "Invalid input: %s" % e)
+        if detailed_exc:
+            abort(400, {"cause": "Invalid input",
+                        "detail": str(e)})
```
jd: … here you can use `e.path` to have details being not a string but a computer understandable path to the key that is wrong.

A string could be constructed for the "reason" field.

Also `cause` has been set usually to "Attribute value error"

gnocchi/rest/api.py
sileht: I don't really get the use case behind unit too. I just need to be able to fill any metric attributes here for Ceilometer and unit is just one of them.

sileht:
Pull-request updated, HEAD is now 4fe78cbfe06146e88d0bb50a8843485056d212bd

gnocchi/rest/api.py
sileht: done

gnocchi/rest/api.py
sileht: done

gnocchi/tests/functional/gabbits/batch-measures.yaml
```diff
@@ -188,6 +216,25 @@ tests:
       response_json_paths:
         $:
           - ["2015-03-06T14:34:12+00:00", 1.0, 12.0]
+          - ["2015-03-06T14:35:12+00:00", 1.0, 24.0]
+
+    - name: get created metric to check creation with newformat
+      GET: /v1/resource/generic/46c9418d-d63b-4cdd-be89-8f57ffc5952e/metric/newformat
```
chungg: should you check that this newformat uses the defined policy and not a default policy.

sileht: This is checked line 228

chungg: word :)

Issue #658, opened 2018-01-17
SplitKey equivalence comparison is broken


chungg:
```python
>>> skey = carbonara.SplitKey.from_timestamp_and_sampling(dt, sampling)
>>> skey == skey
True
>>> skey != skey
False
>>> skey2 = carbonara.SplitKey.from_timestamp_and_sampling(dt, sampling)
>>> skey != skey2
True
>>> skey == skey2
True
```

skey is both equal and not equal to skey2 even though it's the same datetime and sampling. what's strange is when comparing the key, it's right:
```python
>>> skey.key != skey2.key
False
>>> skey.key == skey2.key
True
```

chungg:
uh. so this could be really, really, really bad or not a big deal.

i know __ne__ is wrong but it seems to be only used in one place (https://github.com/gnocchixyz/gnocchi/blob/stable/3.0/gnocchi/storage/_carbonara.py#L321). so as long as __le__, __gt__, and __ge__ work fine, then it's good, it's just doing some needless work. that said, i have no idea how __le__, __gt__, and __ge__ work. they seem to be fine in test but i don't know if it's just because the values i'm using or because it's fine.

chungg:
see #659. long story short, we use total_ordering so only ne was missing incorrect... and only for python2.

PR #659, opened 2018-01-17
 set __ne__ method

chungg:
neither total_ordering nor python2 sets a default __ne__ value
behaviuor opposite to __eq__. this only happens in python3[1] so we
need to set this.

[1] https://docs.python.org/3/whatsnew/3.0.html#operators-and-special-methods

Fixes: #658

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -19,6 +19,7 @@
 import functools
 import itertools
 import math
+import operator
 import random
 import re
 import struct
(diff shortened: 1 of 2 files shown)


jd:
@chungg You missed https://github.com/gnocchixyz/gnocchi/pull/659/files#diff-d99450bd1e2fa5016cec0a79f33d77afR445

https://docs.python.org/2/library/functools.html#functools.total_ordering

chungg:
lol! google even told me about total_ordering... I should've searched our code before trying to debug how it was doing undefined comparisons.

I'll clean up code to add just ne

PR #662, opened 2018-01-18
set __ne__ method

chungg:
neither total_ordering nor python2 sets a default __ne__ value
behaviuor opposite to __eq__. this only happens in python3[1] so we
need to set this.

[1] https://docs.python.org/3/whatsnew/3.0.html#operators-and-special-methods

Fixes: #658

(cherry picked from commit 70cc23c4289b97c8fb3e0f12d5dadc00549516f3)

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -22,6 +22,7 @@ import itertools
 import logging
 import math
 import numbers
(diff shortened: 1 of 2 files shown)


chungg:
:( i need new tests.

PR #667, opened 2018-01-18
remove optional datetime resourcetype patch

chungg:
this was originally incorrectly backported in d2ed3846b7e99b59de96827afc71a27b031225c6.
it fails as expected now. why? i don't know.


jd:
Because pifpaf 2.0.0 was bugged.

PR #669, opened 2018-01-19
storage: pass Aggregation object to storage get_measures()

jd:
This frees a bit more the storage engine from the archive policy system.

### DIFF
--- a/gnocchi/archive_policy.py
+++ b/gnocchi/archive_policy.py
@@ -27,6 +27,9 @@ from gnocchi import aggregation
 from gnocchi import utils
 
 
+ATTRGETTER_GRANULARITY = operator.attrgetter("granularity")
+
+
 class ArchivePolicy(object):
 
     DEFAULT_AGGREGATION_METHODS = ()
@@ -93,6 +96,17 @@ class ArchivePolicy(object):
             if d.granularity == granularity:
                 return aggregation.Aggregation(
                     method, d.granularity, d.timespan)
+
+    def get_aggregations_for_method(self, method):
+        """Return a list of aggregation for a method.
+
+        List is sorted by granularity, desc.
+
+        :param method: Aggregation method.
+        """
+        return [aggregation.Aggregation(method, d.granularity, d.timespan)
+                for d in sorted(self.definition,
+                                key=ATTRGETTER_GRANULARITY, reverse=True)]
 
     @property
     def aggregations(self):
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -17,6 +17,7 @@
 import collections
 import functools
 import itertools
+import operator
 import uuid
 
 import jsonpatch
@@ -48,6 +49,9 @@ try:
     PROMETHEUS_SUPPORTED = True
 except ImportError:
     PROMETHEUS_SUPPORTED = False
+
+
+ATTRGETTER_GRANULARITY = operator.attrgetter("granularity")
 
 
 def arg_to_list(value):
@@ -530,6 +534,17 @@ class MetricController(rest.RestController):
                 },
             })
 
+        aggregations = []
+        for g in sorted(granularity, reverse=True):
+            agg = self.metric.archive_policy.get_aggregation(
+                aggregation, g)
+            if agg is None:
+                abort(404, six.text_type(
+                    storage.AggregationDoesNotExist(
+                        self.metric, aggregation, g)
+                ))
+            aggregations.append(agg)
+
         if (strtobool("refresh", refresh) and
                 pecan.request.incoming.has_unprocessed(self.metric.id)):
             try:
@@ -540,8 +555,7 @@ class MetricController(rest.RestController):
                 abort(503, six.text_type(e))
         try:
             return pecan.request.storage.get_measures(
-                self.metric, granularity, start, stop, aggregation,
-                resample)
+                self.metric, aggregations, start, stop, resample)[aggregation]
         except (storage.MetricDoesNotExist,
                 storage.AggregationDoesNotExist) as e:
             abort(404, six.text_type(e))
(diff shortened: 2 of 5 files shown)


jd:
Pull-request updated, HEAD is now 463328798a24e4f991b177afd0b8d200bf4be029

gnocchi/archive_policy.py
```diff
@@ -94,6 +94,17 @@ def get_aggregation(self, method, granularity):
...
+    def get_aggregations_for_method(self, method):
+        """Return a list of aggregation for a method.
+
+        List is sorted by granularity, desc.
+
+        :param method: Aggregation method.
+        """
+        return [aggregation.Aggregation(method, d.granularity, d.timespan)
+                for d in sorted(self.definition,
+                                key=lambda d: d.granularity, reverse=True)]
```
chungg: use operator.attrgetter('granularity')

gnocchi/rest/api.py
```diff
@@ -1858,6 +1868,18 @@ def get_cross_metric_measures_from_objs(metrics, start=None, stop=None,
...
+        for metric in metrics:
+            for g in granularity:
+                agg = metric.archive_policy.get_aggregation(
+                    aggregation, g)
+                if agg is None:
+                    abort(404, six.text_type(
+                        storage.AggregationDoesNotExist(metric, aggregation, g)
+                    ))
+                aggregations.add(agg)
+        aggregations = sorted(aggregations, reverse=True)
```
chungg: shouldn't this be sorted on a key?

gnocchi/rest/api.py
```diff
@@ -1858,6 +1868,18 @@ def get_cross_metric_measures_from_objs(metrics, start=None, stop=None,
                          for metric in metrics),
                     'No granularity match'))
 
+        aggregations = set()
+        for metric in metrics:
+            for g in granularity:
```
chungg: this duplicates what is happening on new line 1912. probably remove it from there?

gnocchi/storage/__init__.py
```diff
@@ -226,8 +218,13 @@ def get_measures(self, metric, granularities,
             agg_timeseries = list(map(lambda agg: agg.resample(resample),
                                       agg_timeseries))
 
-        return list(itertools.chain(*[ts.fetch(from_timestamp, to_timestamp)
-                                      for ts in agg_timeseries]))
+        return {
```
chungg: do we really need to change the output? there isn't really a scenario which allows us to get multiple aggregation_methods at once.

gnocchi/rest/api.py
```diff
@@ -525,6 +525,17 @@ def get_measures(self, start=None, stop=None, aggregation='mean',
                 },
             })
 
+        aggregations = []
+        for g in sorted(granularity, reverse=True):
+            agg = self.metric.archive_policy.get_aggregation(
+                aggregation, g)
+            if agg is None:
```
chungg: this duplicates the check above?

gnocchi/archive_policy.py
jd: Done.

gnocchi/rest/api.py
jd: No, that only check granularity, not the method.

gnocchi/rest/api.py
jd: That only checks the granularity, not the method.

gnocchi/rest/api.py
jd: It's sorted on granularity because it's the second field and the first one is the same (the method), but I guess it's not explicit. :) fixed

gnocchi/storage/__init__.py
jd: …yet :)

jd:
Fixed @chungg suggestions

jd:
@chungg We're actually validating archive policy related stuff in the API only. My plan is to make the storage AP agnostic. There might be some refactoring for validation in the API: a lot of the query parameters are the same. I've postponed that for now to keep things simple and also to be sure I don't break the error return code of the API.

The new `Aggregation` object allows to do fancy stuff like getting the `mean of 5 minutes` and `max of 10 minutes`. If a driver can batch `get` request we'll could get that in a snap. That could be also useful for the dynamic aggregation endpoint.

I'm really doing this patch per patch so it's easy to review :)

chungg:
> The new Aggregation object allows to do fancy stuff like getting the mean of 5 minutes and max of 10 minutes. If a driver can batch get request we'll could get that in a snap. That could be also useful for the dynamic aggregation endpoint.

is this a new api endpoint?

gnocchi/rest/api.py
chungg:
```
         if aggregation not in self.metric.archive_policy.aggregation_methods:
             abort(404, {
                 "cause": "Aggregation method does not exist for this metric",
                 "detail": {
                     "metric": self.metric.id,
                     "aggregation_method": aggregation,
                 },
             })
```
this?

gnocchi/rest/api.py
chungg:
```
                 if (aggregation
                    not in metric.archive_policy.aggregation_methods):
                     abort(404, {
                         "cause":
                         "Aggregation method does not exist for this metric",
                         "detail": {
                             "metric": str(metric.id),
                             "aggregation_method": aggregation,
                         },
                     })
```

that really seems to be checking method :P

gnocchi/storage/__init__.py
chungg: mysterious...

jd:
> is this a new api endpoint?

That or we need to evolve the current one to get several `aggregation_methods` parameters. Currently it only accept one and one `granularity`.

gnocchi/rest/api.py
jd: Yes this checks the method; so we need both.

Not saying that it might be factorizable in a next patch

gnocchi/rest/api.py
jd: Not sure I understand. What I mean is that the piece of code I added only returns 404 if the granularity does not exist. But if the method does not exist it won't return 404.

The piece of code you pasted do this.

We need both.

chungg: the code i pasted was what i was suggesting is being duplicated but i guess we can refactor in next.

jd:
Rebased

gnocchi/storage/redis.py
```diff
@@ -94,7 +94,7 @@ def _delete_metric(self, metric):
 
     def _get_measures(self, metric, keys, aggregation, version=3):
         if not keys:
-            return []
+            return {}
```
chungg: is there a reason you change this to return empty dict? this is just serialized data so i don't think it's an dict... or a list either

jd: It's actually a list that is returned in the canonical implementation in `gnocchi.storage`. I'll fix it.

jd:
Pull-request updated, HEAD is now 72fdb8701083650b90cbc70f1a3dfc85f596b125

PR #674, opened 2018-01-23
Pull request for rejoin-on-fail

jd:
metricd: restart scheduler on group membership issue
cli: fix error message string formatting

### DIFF
--- a/gnocchi/cli.py
+++ b/gnocchi/cli.py
@@ -174,7 +174,7 @@ class MetricScheduler(MetricProcessBase):
                      'with up to %s metrics', self.block_index,
                      self.block_size)
         except Exception:
-            LOG.error('Error getting block to work on (%s), '
+            LOG.error('Error getting block to work on, '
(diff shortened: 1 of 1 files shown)


jd:
@sileht I need your review on this because I'm not 100% sure the `SystemExit` is correctly used in Cotyledon (1.2 branch here).

gnocchi/cli.py
```diff
@@ -162,7 +162,11 @@ def set_block(self, event):
         get_members_req = self._coord.get_members(self.GROUP_ID)
         try:
             members = sorted(get_members_req.get())
-            self.block_index = members.index(self._my_id)
+            try:
+                self.block_index = members.index(self._my_id)
+            except ValueError:
+                LOG.error('Unable to find myself in group, restarting')
```
sileht: Maybe you should call self.terminate() to cleanly shutdown, or at least self._coord.stop() to disconnect from redis before raising SystemExit().

Issue #464, opened 2017-11-02
Regular non-admin user can't get measures of his metrics. Gets the 403 response instead.


ozerovandrei:
Hello.
I've performed an upgrade from Gnocchi 4.0 to 4.1 in my testing environment.
I have some issue related to the Gnocchi REST API.

When I'm authorized with the non-admin Keystone token I can't get measures for metrics that have resource.project_id value of my project_id.
I don't have that issue in 4.0 release.

I tested that with the following request:
`curl -XGET -H 'Content-Type: application/json' -H "X-Auth-Token: $(openstack token issue -c id -f value)" 'http://gnocchi_ip:gnocchi_port/v1/resource/SOME_RESOURCE_TYPE/RESOURCE_UUID/metric/SOME_METRIC_NAME/measures'`

I've got the 403 reply to that request, but I can successfully perform the following requests:
`GEl -XGET -H 'Content-Type: application/json' -H "X-Auth-Token: $(openstack token issue -c id -f value)" 'http://gnocchi_ip:gnocchi_port/v1/resource/SOME_RESOURCE_TYPE/RESOURCE_UUID'`

` GEl -XGET -H 'Content-Type: application/json' -H "X-Auth-Token: $(openstack token issue -c id -f value)" 'http://gnocchi_ip:gnocchi_port/v1/resource/SOME_RESOURCE_TYPE/RESOURCE_UUID/metric/SOME_METRIC_NAME'`

As far as I understand, that 403 responce is coming from there: 
[self.enforce_metric("get measures")](https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L466)

I've done some debbugging of the [MetricController](https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L432.)
What I've got, is that when I'm doing the request to a metric (_curl ...metric/SOME_METRIC_NAME_), I get the
**json.to_primitive(self.metric)** result in the [enforce()](https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L444) with the
full info about the metric with the resource info inside (**resource.created_by_user_id**, **resource.id**, **resource.revision_start**, **resource.project_id**, etc.).

But when I make the request to a measures (_curl ...metric/SOME_METRIC_NAME/measures_), I get the tiny result of the **json.to_primitive(self.metric)** without
fields of resource info (just the **resource_id** field).

[...]

PR #675, opened 2018-01-23
rest: load metric details for ACL

jd:
Fixes: #464


chungg:
works as good as it looks. :)

PR #679, opened 2018-01-24
metricd: stop coordinator on membership issue

sileht:
This change stop the coordinator before existing on membership issue.


chungg:
is this a backport or just something 3.0 specific?

sileht:
Yes, and it's to follow this @jd change https://github.com/gnocchixyz/gnocchi/pull/674

jd:
I'll check if we need this for 3.1

PR #700, opened 2018-01-29
update documentation regarding resource ids

### LINKED ISSUE #698: Document support for arbitrary resource ids
I'm using gnocchi `4.1.3`.

I've discovered that it is possible to create resources with an arbitrary resource id.  For example,

```
$ curl -H 'content-type: application/json' -u admin: localhost:8041/v1/resource/generic -d @- <<EOF
{
  "id": "some-arbitrary-string",
}
EOF
```

Gnocchi moves that over to the `original_resource_id` field and assigns a unique UUID, and that original id can be used in queries where normally one would put the UUID, as in:

```
$ curl -s -u admin: localhost:8041/v1/resource/generic/some-arbitrary-string
{
  "user_id": null,
  "revision_end": null,
  "original_resource_id": "some-arbitrary-string",
  "created_by_user_id": "admin",
  "revision_start": "2018-01-28T17:24:11.509758+00:00",
  "project_id": null,
  "id": "6fba60c9-8174-5822-8730-9d4bf6a7c1b9",
  "metrics": {},
  "ended_at": null,
  "creator": "admin",
  "started_at": "2018-01-28T17:24:11.509746+00:00",
  "created_by_project_id": "",
  "type": "generic"
}
```

I don't see this behavior documented anywhere. Can I rely on it? It's extremely interesting for an IoT use case, where a device may already have a unique id and being able to use that avoids some round-trips to the gnocchi api.

larsks:
Add information about the special handling of resource ids.

Closes #698

### DIFF
--- a/doc/source/rest.j2
+++ b/doc/source/rest.j2
@@ -387,12 +387,19 @@ To create a generic |resource|:
 
 {{ scenarios['create-resource-generic']['doc'] }}
 
-The *id*, *user_id* and *project_id* attributes must be an UUID. The timestamp
+The *user_id* and *project_id* attributes must be UUIDs. The timestamp
 describing the lifespan of the |resource| are optional, and *started_at* is by
 default set to the current timestamp.
 
-It's possible to retrieve the |resource| by the URL provided in the `Location`
-header.
+The *id* attribute may be a UUID or some other arbitrary string.  If it is
+not a UUID, the original value will be stored in the *original_resource_id*
+attribute and Gnocchi will generate a new UUID that is unique for the user.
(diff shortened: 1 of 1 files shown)


doc/source/rest.j2
```diff
@@ -387,12 +387,19 @@ To create a generic |resource|:
 
 {{ scenarios['create-resource-generic']['doc'] }}
 
-The *id*, *user_id* and *project_id* attributes must be an UUID. The timestamp
+The *user_id* and *project_id* attributes must be UUIDs. The timestamp
```
jd: This is actually wrong now. We should fix it too. They are just string.

PR #704, opened 2018-01-30
metricd: remove janitor process

### LINKED ISSUE #652: janitor is idle most of the time
Before reporting an issue on Gnocchi, please be sure to provide all necessary
information.

master

just run it 

janitor process on each metricd

do we really need a janitor process on each metricd? majority of the time it's doing nothing. when there is work, it's a race between all the other janitors in metricd agents to cleanup a deleted metric.

this connects back to #608... in my mind, it's a bit strange that part of the cleanup is done by janitor and part of it's done by processors. it'd be good to make compress/delete separate from standard processing but we need to leverage the knowledge of processors to avoid pointless checking in janitor

maybe the processor can put on to a (internal/external) queue that metricA is processed and has a new last timestamp and first mutable timestamp. it's then the janitor's job to deal with cleanup.

this isn't necessarily to speed up processing time but to make better use of janitor.

jd:
This merge the janitor process in the worker 0 of the metric processor. This
saves some megabytes of memory.

Closes #652

### DIFF
--- a/gnocchi/opts.py
+++ b/gnocchi/opts.py
@@ -153,7 +153,7 @@ def list_opts():
                        deprecated_group='storage',
                        default=300,
                        required=True,
-                       help="How many seconds to wait between "
+                       help="How many seconds to wait (at least) between "
                        "cleaning of expired data"),
             cfg.IntOpt('processing_replicas',
                        default=3,
--- a/gnocchi/cli/metricd.py
+++ b/gnocchi/cli/metricd.py
@@ -148,6 +148,9 @@ class MetricProcessor(MetricProcessBase):
         self._get_sacks_to_process = cachetools.func.ttl_cache(
             ttl=conf.metricd.metric_processing_delay
         )(self._get_sacks_to_process)
+        if worker_id == 0:
+            self.name = "processing+janitor"
+            self._last_expunge = utils.StopWatch().start()
 
     @tenacity.retry(
         wait=utils.wait_exponential,
@@ -217,6 +220,12 @@ class MetricProcessor(MetricProcessBase):
             return self._tasks or self.fallback_tasks
 
     def _run_job(self):
+        if (self.worker_id == 0 and
+           self._last_expunge.elapsed()
+           >= self.conf.metric.metric_cleanup_delay):
+            self.store.expunge_metrics(self.incoming, self.index)
+            self._last_expunge.reset()
+
         m_count = 0
         s_count = 0
         # We are going to process the sacks we got notified for, and if we got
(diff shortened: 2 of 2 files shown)


gnocchi/opts.py
```diff
@@ -149,12 +149,6 @@ def list_opts():
                        help="How many seconds to wait between "
                        "metric ingestion reporting. Set value to -1 to "
                        "disable reporting"),
-            cfg.IntOpt('metric_cleanup_delay',
```
chungg: we should probably keep this? part of the issue is that it does nothing on cleanup also as there's little to cleanup. if we remove this, it's checking to do nothing even more often.

jd: Not sure I'm smart enough to do that. I'll keep a look :)

jd:
I actually implemented that on the wrong class.

jd:
Pull-request updated, HEAD is now 00ff6879f98efdd9d52564f8c6d5210020fbc4ec

jd:
@chungg I'll do the same for the reporting process is that works for you.

chungg:
you'll need to set a constraint that reporting interval > processing interval

i'm still slightly concerned that this locks us to indexer. if this is only to lower memory, we will probably save more not loading oslo_db in n*workers vs merging janitor+processing.

jd:
> you'll need to set a constraint that reporting interval > processing interval

Same goes for cleanup already. I'll add some code to make that ok.

> i'm still slightly concerned that this locks us to indexer. if this is only to lower memory, we will probably save more not loading oslo_db in n*workers vs merging janitor+processing.

I don't understand what you mean here by "locks us to indexer".

chungg:
> I don't understand what you mean here by "locks us to indexer".

it's my understanding that janitor will always require indexer. if we merge processing and janitor, then the processing worker will always have indexer so i don't know how you'd achieve #548.

i thought about #548 a little bit last night and i'm not entirely sure how that'd be achieved regardless.

sileht:
As you want for this, but my preference is for isolated process than saving memory.

jd:
Yeah. The real question is how much memory a process like janitor takes @chungg?

chungg:
> Yeah. The real question is how much memory a process like janitor takes @chungg?

the janitor process is 40% of total memory and it decreases as a percentage significantly the more workers/children an agent has. you'll save significantly more memory if you can somehow remove the indexer from processing agents.

as i mentioned, this isn't a big issue. i just wanted to point out that the janitor is mostly idle and even in very large environments, i don't think it has much work... was looking to see if there was an easy hack but it's not the end of world if we have an idle janitor per agent.

chungg:
so my equivalent comparison would be comparing this to ceilometer-expirer which really only needed to be run every few days... except the janitor has signficantly less work to do than ceilometer-expirer.

jd:
40% of total memory of all metricd workers? that depends on the number of workers?
do you have a size in MB?

chungg:
if there is only one processing worker... the janitor and reporting workers are always there (you can disable reporting process technically). reporting worker is roughly 1/2 memory of processing workers... janitor worker is roughly same as processing worker.

for example, if you have one processing worker in a metricd, memory percentage used by janitor is n/(n*2.5) (where n is memory of a single processing agent. ~80MB?). if you have 10 processing workers in metricd, it's n/(n*9.5) which is significantly less.

but yeah, like i said, it's not GB of memory and if you have a large backwindow, the memory used to actually load data would dwarf anything the janitor uses at idle.

jd:
Ok… Well I'll just close this for the time being. I think it's good we discussed it, but process isolation strikes me as more important for now.

Issue #702, opened 2018-01-30
Cannot create named metric after removing it due to duplication


SuderPawel:
### Which version of Gnocchi are you using

Latest 02936ff4

```
commit 02936ff4a950c7c3151dc7f70793f91e2225d926
Author: Lars Kellogg-Stedman <lars@redhat.com>
Date:   Mon Jan 29 08:40:25 2018 -0500

    update documentation regarding resource ids

    Add information about the special handling of resource ids.

    Closes #698
```

### How to reproduce your problem

* create a resource
* create a metric named assigned to resource
* remove created metric
* create a metric named assigned to the same resource - name needs to be the same

### What is the result that you get

```
stack@pasuder-devstack-upstream-all:~/gnocchi$ gnocchi resource list
+--------------------------------------+---------+------------+---------+--------------------------------------+----------------------------------+----------+----------------------------------+--------------+---------+
| id                                   | type    | project_id | user_id | original_resource_id                 | started_at                       | ended_at | revision_start                   | revision_end | creator |
+--------------------------------------+---------+------------+---------+--------------------------------------+----------------------------------+----------+----------------------------------+--------------+---------+
| 803920d3-03a4-42f6-8f64-ae5e0f3a6531 | generic | None       | None    | 803920d3-03a4-42f6-8f64-ae5e0f3a6531 | 2017-12-19T09:53:35.630439+00:00 | None     | 2017-12-19T09:53:35.630452+00:00 | None         | admin   |
+--------------------------------------+---------+------------+---------+--------------------------------------+----------------------------------+----------+----------------------------------+--------------+---------+
stack@pasuder-devstack-upstream-all:~/gnocchi$ gnocchi metric list

stack@pasuder-devstack-upstream-all:~/gnocchi$ gnocchi metric create --resource-id 803920d3-03a4-42f6-8f64-ae5e0f3a6531 metric_name
+------------------------------------+------------------------------------------------------------------+
| Field                              | Value                                                            |
+------------------------------------+------------------------------------------------------------------+
| archive_policy/aggregation_methods | std, count, min, max, sum, mean                                  |
| archive_policy/back_window         | 0                                                                |
| archive_policy/definition          | - points: 8640, granularity: 0:05:00, timespan: 30 days, 0:00:00 |
| archive_policy/name                | low                                                              |
| created_by_project_id              |                                                                  |
| created_by_user_id                 | admin                                                            |
| creator                            | admin                                                            |
[...]

PR #708, opened 2018-01-31
indexer: detach metric from resource when marked as 'deleted'

jd:
This allows to re associate another metric using the same name as used
previously, even if the metric is not expunged yet.

Closes: #702
(cherry picked from commit d3a93519e55fa10bb1de3eac8821e26b2c6581e3)


jd:
Fix create_metric args

PR #709, opened 2018-01-31
redis: concatenate bytes before unserializing

jd:
This is actually faster than concatenating numpy arrays.

### DIFF
--- a/gnocchi/storage/__init__.py
+++ b/gnocchi/storage/__init__.py
@@ -514,7 +514,7 @@ class StorageDriver(object):
             LOG.debug("Skipping %s (already processed)", metric)
             return
 
-        measures.sort(order='timestamps')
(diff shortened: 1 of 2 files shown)


gnocchi/storage/__init__.py
```diff
@@ -514,7 +514,7 @@ def _compute_and_store_timeseries(self, metric, measures):
             LOG.debug("Skipping %s (already processed)", metric)
             return
 
-        measures.sort(order='timestamps')
+        measures = numpy.sort(measures, order='timestamps')
```
chungg: why'd you change this to create new array vs inplace?

jd: Because the array is readonly.

chungg: ok

PR #713, opened 2018-02-02
Pull request for redis-lua

jd:
redis: use a Lua script to get measures
redis: create metrics using a single hsetnx() operation
redis: batch new measures retrieval with a Lua script
redis: use a Lua script to list split keys

### DIFF
--- a/gnocchi/storage/redis.py
+++ b/gnocchi/storage/redis.py
@@ -27,9 +27,36 @@ class RedisStorage(storage.StorageDriver):
     STORAGE_PREFIX = b"timeseries"
     FIELD_SEP = '_'
 
+    _SCRIPTS = {
+        "list_split_keys": """
+local metric_key = KEYS[1]
+local ids = {}
+local cursor = 0
+local substring = "([^" .. ARGV[2] .. "]*)"
+repeat
+    local result = redis.call("HSCAN", metric_key, cursor, "MATCH", ARGV[1])
+    cursor = tonumber(result[1])
+    for i, v in ipairs(result[2]) do
+        -- Only return keys, not values
+        if i % 2 ~= 0 then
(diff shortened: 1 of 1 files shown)


chungg:
is there a reason for this? trying to figure out if this fixes something that is worth making the code less readable.

jd:
Performance improvement :)

chungg:
got some benchmarks? :)

jd:
It improves latency and the number of calls to `sendall()/recv()` on the socket is divided by 2, which is a small win. :)

gnocchi/storage/redis.py
```diff
@@ -27,9 +27,36 @@ class RedisStorage(storage.StorageDriver):
...
+local metric_key = KEYS[1]
+local ids = {}
+local cursor = 0
+local substring = "([^" .. ARGV[2] .. "]*)"
+repeat
+    local result = redis.call("HSCAN", metric_key, cursor, "MATCH", ARGV[1])
+    cursor = tonumber(result[1])
+    for i, v in ipairs(result[2]) do
+        -- Only return keys, not values
+        if i % 2 ~= 0 then
```
chungg: i've no idea what this is for but ok?

jd: HSCAN returns a table in format `{k1, v1, k2, v2, k3, v3…}` so using this condition allows to ignore `k1, k2, k3` and just push `v1, v2, v3` in `ids`

chungg: you mean ignore v1, v2, v3...

jd: Sorry yeah, we want keys here, not values. :)

chungg: cool cool. yeah i google'd most of the code. seems legit to someone that knows zero lua.

jd:
Rebased.

PR #719, opened 2018-02-06
storage: stop requiring a coordinator

jd:
The storage engine does not need a coordinator, except for locking sacks.
However sacks do not belong to the storage driver itself, so it makes no sense
to requires one.


jd:
Fix gabbi tests

PR #720, opened 2018-02-06
tests: test upgrade from Gnocchi 4.2

jd:


jd:
Add missing Travis entries

PR #721, opened 2018-02-06
Stop testing upgrade from Gnocchi 3.1

jd:


jd:
I think the idea is better be safe than sorry in this case. Since master is gonna be 5.x I think, I just assumed we could remove this.

chungg:
what's in fancy 5.x? :)

jd:
@chungg no idea yet. Do you want to wait for the release number to be sure before removing that upgrade test?

PR #724, opened 2018-02-07
tox: don't install gnocchi twice

sileht:
We current install Gnocchi twice, once with no extra and develop mode,
and a second time with extra but without develop mode.

This change install gnocchi only once with extra and develop mode.


sileht:
Pull-request updated, HEAD is now f6e8018a703ce19fc04faf0e5f26b5e7052acd36

PR #727, opened 2018-02-08
redis: batch new measures retrieval with a Lua script

jd:


gnocchi/incoming/redis.py
```diff
@@ -106,12 +115,11 @@ def has_unprocessed(self, metric_id):
...
         key = self._build_measure_path(metric_id)
-        item_len = self._client.llen(key)
-        # lrange is inclusive on both ends, decrease to grab exactly n items
-        item_len = item_len - 1 if item_len else item_len
+        item_len, data = self._scripts['process_measure_for_metric'](
+            keys=[key],
+        )
 
-        yield self._unserialize_measures(metric_id, b"".join(
-            self._client.lrange(key, 0, item_len)))
```
chungg: isn't this already a batch/single operation?  or is this just saving the initial client.llen call?

jd: This is 2 operations, e.g. 2 round trips: `LLEN` and then `LRANGE`. So yeah this compacts it to one round trip.

chungg: i guess the lua code is simple enough

PR #730, opened 2018-02-08
redis: use a Lua script to get measures

jd:


gnocchi/storage/redis.py
```diff
@@ -114,16 +128,18 @@ def _delete_metric(self, metric):
...
+        code, result = self._scripts['get_measures'](
+            keys=[self._metric_key(metric)],
+            args=fields,
+        )
+        if code == -1:
+            sampling = utils.to_timespan(result.split(self.FIELD_SEP_B)[2])
+            raise storage.AggregationDoesNotExist(
+                metric, aggregation, sampling)
+        if code == -2:
+            raise storage.MetricDoesNotExist(metric)
```
chungg: i don't think we have tests for this as i deleted the following and it still passes.

```
+        if code == -1:
+            sampling = utils.to_timespan(result.split(self.FIELD_SEP_B)[2])
+            raise storage.AggregationDoesNotExist(
+                metric, aggregation, sampling)
+        if code == -2:
+            raise storage.MetricDoesNotExist(metric)
```

tbh, i'm not sure it's possible to get this far as we do validate the aggregation beforehand.

jd: I just kept the same structure, so IIUC, this is not a new problem (if it is one at all)?

chungg: well it's a problem in that the errors raise are never validated so in theory, you could've broken it :P

jd:
@chungg fair enough https://github.com/gnocchixyz/gnocchi/issues/732

PR #731, opened 2018-02-08
incoming: batch incoming measures retrieval

jd:
This patches rename the context manager process_measure_for_metric to
process_measure_for_metrics as it now yields a dict of several metrics at the
same time.

### DIFF
--- a/gnocchi/incoming/ceph.py
+++ b/gnocchi/incoming/ceph.py
@@ -190,40 +190,44 @@ class CephStorage(incoming.IncomingDriver):
         return bool(self._list_keys_to_process(sack, object_prefix))
 
     @contextlib.contextmanager
-    def process_measure_for_metric(self, metric_id):
-        sack = self.sack_for_metric(metric_id)
-        key_prefix = self.MEASURE_PREFIX + "_" + str(metric_id)
-
-        processed_keys = []
+    def process_measure_for_metrics(self, metric_ids):
+        measures = {}
+        processed_keys = defaultdict(list)
         with rados.ReadOpCtx() as op:
-            omaps, ret = self.ioctx.get_omap_vals(op, "", key_prefix, -1)
-            self.ioctx.operate_read_op(op, self.get_sack_name(sack),
-                                       flag=self.OMAP_READ_FLAGS)
-            # NOTE(sileht): after reading the libradospy, I'm
-            # not sure that ret will have the correct value
-            # get_omap_vals transforms the C int to python int
-            # before operate_read_op is called, I dunno if the int
-            # content is copied during this transformation or if
-            # this is a pointer to the C int, I think it's copied...
-            try:
-                ceph.errno_to_exception(ret)
-            except rados.ObjectNotFound:
-                # Object has been deleted, so this is just a stalled entry
-                # in the OMAP listing, ignore
-                return
+            for metric_id in metric_ids:
+                sack = self.sack_for_metric(metric_id)
+                key_prefix = self.MEASURE_PREFIX + "_" + str(metric_id)
+                omaps, ret = self.ioctx.get_omap_vals(op, "", key_prefix, -1)
+                self.ioctx.operate_read_op(op, self.get_sack_name(sack),
+                                           flag=self.OMAP_READ_FLAGS)
+                # NOTE(sileht): after reading the libradospy, I'm
+                # not sure that ret will have the correct value
+                # get_omap_vals transforms the C int to python int
+                # before operate_read_op is called, I dunno if the int
+                # content is copied during this transformation or if
+                # this is a pointer to the C int, I think it's copied...
+                try:
+                    ceph.errno_to_exception(ret)
+                except rados.ObjectNotFound:
+                    # Object has been deleted, so this is just a stalled entry
+                    # in the OMAP listing, ignore
+                    continue
 
-        measures = self._make_measures_array()
-        for k, v in omaps:
-            measures = numpy.concatenate(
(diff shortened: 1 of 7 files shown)


jd:
Fixed Swift

jd:
Fixed Ceph

chungg:
didn't read it all completely but i do have a concern that it seems to really only benefit redis drivers with regard to number of saved i/o.

it worries me slightly that all the other drivers execute serially. seems like it'd be better if this was done using parallel_operations.

chungg:
> We can use parallel operations in the some drivers if it improves perf (but on large scale I doubt so)

i would say it only benefits large scale. threading only really matters if you're going over network

> Batching creates less operation as things can be done in a single roundtrip and speeds up the total execution time.

right, that's what i was saying. you only batch for redis driver iiuc, the other drivers are the same.

gnocchi/incoming/ceph.py
```diff
@@ -190,40 +190,44 @@ def has_unprocessed(self, metric_id):
...
         yield measures
 
         # Now clean omap
         with rados.WriteOpCtx() as op:
-            # NOTE(sileht): come on Ceph, no return code
-            # for this operation ?!!
-            self.ioctx.remove_omap_keys(op, tuple(processed_keys))
-            self.ioctx.operate_write_op(op, self.get_sack_name(sack),
-                                        flags=self.OMAP_WRITE_FLAGS)
+            for sack, keys in six.iteritems(processed_keys):
```
chungg: i'm not sure we need to group things by sack here. iiuc, this command only gets triggered by sack, never cross-sacks  as it only ever locks one sack.

gnocchi/incoming/redis.py
```diff
@@ -113,16 +113,27 @@ def has_unprocessed(self, metric_id):
...
-        # ltrim is inclusive, bump 1 to remove up to and including nth item
-        self._client.ltrim(key, item_len + 1, -1)
+    def process_measure_for_metrics(self, metric_ids):
+        measures = {}
+        pipe = self._client.pipeline(transaction=False)
+        for metric_id in metric_ids:
+            key = self._build_measure_path(metric_id)
+            self._scripts['process_measure_for_metric'](
+                keys=[key],
+                client=pipe,
```
chungg: where is client=pipe being used? it's not in script. is this some redis+lua thing?

gnocchi/incoming/s3.py
```diff
@@ -158,21 +158,23 @@ def has_unprocessed(self, metric_id):
...
-                Key=f)
-            measures = numpy.concatenate((
-                measures,
-                self._unserialize_measures(f, response['Body'].read())
-            ))
+    def process_measure_for_metrics(self, metric_ids):
+        measures = defaultdict(self._make_measures_array)
+        all_files = []
+        for metric_id in metric_ids:
+            sack = self.sack_for_metric(metric_id)
```
chungg: sack will always be the same.

gnocchi/incoming/swift.py
```diff
@@ -97,18 +97,24 @@ def has_unprocessed(self, metric_id):
...
-                f['name'],
-                self.swift.get_object(sack_name, f['name'])[1],
-            )
-            for f in files
-        ])
+    def process_measure_for_metrics(self, metric_ids):
+        measures = {}
+        all_files = defaultdict(list)
+        for metric_id in metric_ids:
+            sack = self.sack_for_metric(metric_id)
```
chungg: same

gnocchi/incoming/ceph.py
jd: This is not what the signature of the function says, so why your right in theory, it would be wrong to do that.

gnocchi/incoming/redis.py
jd: Not sure I understand the question. This makes the script executed by the pipe.

gnocchi/incoming/s3.py
jd: You don't know that based on the signature.

jd:
@chungg you want the code to make assumption it cannot make at this layer.

gnocchi/incoming/ceph.py
chungg: *shrugs* i meant to pass it in but i guess you can do it as a followup.

gnocchi/incoming/redis.py
chungg: if i look at process_measure_for_metric lua script above, it doesn't mention anything about client. i was just wondering if this was some param script takes in and if so you had a link for my personal knowledge.

chungg: i guess as a followup, why dont you just build all the keys and pass it in to script? i guess you're trying to minimise change?

gnocchi/incoming/ceph.py
jd: It's in the roadmap :)

gnocchi/incoming/redis.py
jd: Oh I get your question. The only arguments you can pass to a Lua script are `KEYS` and `ARGS`. So yeah it's a param for the pipe to execute the script (I agree the Python Redis API is not perfect in this regard).

I could pass all the keys to the script – I actually started to implement it. However it actually makes do much more Lua than Python, and I am not sure it's worth it, so I did not. It's still doable as a follow up for sure.

chungg: ack. i'll just trust that client=pipe is something. i guess the data does seem to be attached to it eventually (assuming the gates are blindly passing)

PR #733, opened 2018-02-09
redis: use regular keys and not hash to store metrics and splits

jd:
This changes the storage format so it's possible to use SETRANGE in a future
patch.

### DIFF
--- a/gnocchi/storage/redis.py
+++ b/gnocchi/storage/redis.py
@@ -22,33 +22,28 @@ from gnocchi import utils
 class RedisStorage(storage.StorageDriver):
     WRITE_FULL = True
 
-    STORAGE_PREFIX = b"timeseries"
-    FIELD_SEP = '_'
-    FIELD_SEP_B = b'_'
+    STORAGE_PREFIX = "timeseries"
 
     _SCRIPTS = {
         "list_split_keys": """
 local metric_key = KEYS[1]
 local ids = {}
 local cursor = 0
-local substring = "([^" .. ARGV[2] .. "]*)"
+local substring = "[^%s]*%s[^%s]*%s([^%s]*)"
 repeat
-    local result = redis.call("HSCAN", metric_key, cursor, "MATCH", ARGV[1])
+    local result = redis.call("SCAN", cursor, "MATCH", ARGV[1])
     cursor = tonumber(result[1])
     for i, v in ipairs(result[2]) do
-        -- Only return keys, not values
-        if i % 2 ~= 0 then
-            ids[#ids + 1] = v:gmatch(substring)()
-        end
+        ids[#ids + 1] = v:gmatch(substring)()
     end
 until cursor == 0
 if #ids == 0 and redis.call("EXISTS", metric_key) == 0 then
     return -1
 end
 return ids
-""",
+""" % (redis.SEP, redis.SEP, redis.SEP, redis.SEP, redis.SEP),
         "get_measures": """
-local results = redis.call("HMGET", KEYS[1], unpack(ARGV))
+local results = redis.call("MGET", unpack(ARGV))
 local final = {}
 for i, result in ipairs(results) do
     if result == false then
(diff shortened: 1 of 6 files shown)


jd:
Pull-request updated, HEAD is now 486bd43315846b85c713a96b3c733c99ad147f44

jd:
Can you be more precise on this? The performance improvement due to the next change using `WRITE_FULL = False` are big enough that this probably needs more than a feeling :)

chungg:
curious but what is your benchmark?

i'm just giving you context as to why i switched originally. not a feeling.

gnocchi/storage/redis.py
```diff
@@ -22,33 +22,28 @@
...
     _SCRIPTS = {
         "list_split_keys": """
 local metric_key = KEYS[1]
 local ids = {}
 local cursor = 0
-local substring = "([^" .. ARGV[2] .. "]*)"
+local substring = "[^%s]*%s[^%s]*%s([^%s]*)"
 repeat
-    local result = redis.call("HSCAN", metric_key, cursor, "MATCH", ARGV[1])
+    local result = redis.call("SCAN", cursor, "MATCH", ARGV[1])
```
chungg: for example, HSCAN and SCAN are O(N) but HSCAN's N is magnitudes smaller than SCAN.

that said, we don't call list_split_keys all the time now so maybe it's less of issue.

jd: > for example, HSCAN and SCAN are O(N) but HSCAN's N is magnitudes smaller than SCAN.

I'm curious to know if it's O(n) with a pattern though. Sounds to me that if Redis had a sorted index it could limit the output. Anyway that might not be the case, just thinking out loud.

> that said, we don't call list_split_keys all the time now so maybe it's less of issue.

And I'm also working on reducing the calls to `list_split_keys` so even if it's a small problem, it'll eventually get even smaller.

jd:
> curious but what is your benchmark?

I use a simple fake metricd to profile and optimize:
https://gist.github.com/jd/7466cebd1c7fb3a59f9c4e768572ce5d

gnocchi/storage/redis.py
chungg: > I'm curious to know if it's O(n) with a pattern though. Sounds to me that if Redis had a sorted index it could limit the output. Anyway that might not be the case, just thinking out loud.

if you believe https://stackoverflow.com/questions/41348963/does-a-redis-scan-with-a-prefix-match-pattern-scan-all-keys-in-database, it checks all keys. i imagine it's because it's iterating through and so it technically has to keep checking in case a new value came in after iterating started which is probably what i noticed when i was testing.

to give context, i have 20K metrics in my benchmark but i also have 24 aggregrate x granularity pairs, so already instead of N=20K, it's N=480K. also, i try to force at least one split so you add more keys for each split... for each metric... i believe 10K instances will garner 300K+ metrics... multiply 24... multiply splits...

my concern is how this performs against medium workloads and larger... at small scale, this is definitely fine. maybe redis is really only built for small case... or benchmark case.

jd: Thanks for the pointer. I went ahead and experimented myself and indeed, it seems there's no optimization being done. :(

If you have 300k+ metric with 8 aggregations (let's be optimistic) that's 2.4 M metric. If you have 2 splits per metric, that's close to 5M splits (keys). Each split is 3600 points * ~5 bytes = 17 Kbytes per split. That means your database is 17 KB * 5000000 ~= 83 GB. That's a lot of RAM. At this point you need sharding or something like that (e.g. Redis Cluster).

So it might be better to argue with a more likely scenario. If we consider a 1/10th of that, it's 8.3GB of RAM and therefore 500k splits (keys) for 30k metrics.

On my laptop, I filled 500k keys in Redis (content is "foobar" but I don't think it changes much):

```
abydos ~ ➔ time redis-cli --scan | wc -l
  500000
redis-cli --scan  0.58s user 0.51s system 35% cpu 3.023 total
wc -l  0.00s user 0.00s system 0% cpu 3.022 total

abydos ~ ➔ time redis-cli --scan --pattern 999?? | wc -l
     100
redis-cli --scan --pattern 999??  0.40s user 0.48s system 32% cpu 2.741 total
wc -l  0.00s user 0.00s system 0% cpu 2.740 total
```

2.7s seems quite long to list all keys… especially to retrieve data.

jd:
The `SCAN` performance issued by @chungg is worse than I thought. I'm gonna close this for now and I'll work on a `SETRANGE` implementation for hashes in Lua instead. That should allow to use `WRITE_FULL = False` without changing the current storage scheme.

chungg:
/me deletes msg i was typing.

my suggestion was just to add note "redis will probably not scale beyond x metrics"... if you have a better idea, i'm all for it :) 

i'm still curious about your comment that you can eliminate scans even more. that'd be great especially if it benefits all drivers.

PR #734, opened 2018-02-09
improve documentation on tuning the number of sacks

larsks:
The documentation suggested that the number of sacks could be tuned
via a configuration option the in gnocchi.conf file.  This commit
corrects the docs to indicate that the value can be changed only with
the gnocchi-change-sack-size command.

This commit also corrects spelling, grammar, and formatting issues in
the same section.

### DIFF
--- a/doc/source/operating.rst
+++ b/doc/source/operating.rst
@@ -227,62 +227,63 @@ metricd daemon on any number of servers.
 How to scale measure processing
 -------------------------------
 
-Measurement data pushed to Gnocchi is divided into sacks for better
-distribution. The number of partitions is controlled by the `sacks` option
-under the `[incoming]` section. This value should be set based on the
-number of active |metrics| the system will capture. Additionally, the number of
-`sacks`, should be higher than the total number of active metricd workers.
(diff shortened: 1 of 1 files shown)


doc/source/operating.rst
```diff
@@ -227,62 +227,63 @@ metricd daemon on any number of servers.
...
 
-Alternatively, to minimise API downtime::
+Alternatively, to minimize API downtime:
 
-  1. Run gnocchi-upgrade but use a new incoming storage target such as a new
-     ceph pool, file path, etc... Additionally, set |aggregate| storage to a
-     new target as well.
+1. Run gnocchi-upgrade but use a new incoming storage target such as a new
+   ceph pool, file path, etc. Additionally, set |aggregate| storage to a
+   new target as well.
```
jd: Do we want to keep that "additionally" ?

larsks: Based on the irc discussion, it sounds like targeting new aggregate storage in this step may reduce the run-time and concomitant i/o of running the command against active aggregate storage. If that's not actually the case then sure, we should probably remove it.

PR #735, opened 2018-02-10
Add Chef class to orchestrate complex actions

jd:
Some of the code is currently tied to the Storage driver, whereas it interacts
with many other components. This should be moved out.
That patch introduces the Chef, a class responsible for doing complex actions
that require interactions and orchestrations with different actors.


jd:
Pull-request updated, HEAD is now 92e9d399e76da34551f8915bf1663a31671afec9

PR #737, opened 2018-02-12
chef: import process_new_measures and refresh_metric

jd:
This moves those methods out of storage and where they belong: in the Chef!

### DIFF
--- a/gnocchi/chef.py
+++ b/gnocchi/chef.py
@@ -18,6 +18,7 @@ import itertools
 import operator
 
 import daiquiri
+import six
 
 from gnocchi import indexer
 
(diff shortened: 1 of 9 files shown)


gnocchi/chef.py
```diff
@@ -92,3 +97,41 @@ def expunge_metrics(self, sync=False):
                             raise
                         LOG.error("Unable to expunge metric %s from storage",
                                   metric, exc_info=True)
+
+    def refresh_metric(self, metric, timeout):
```
chungg: for discussion, any reason why you didn't put refresh_metric in api as it's the only place it's used?

jd: Hm that requires knowledge about locking, sacks, etc, so it seems good to be there.

PR #738, opened 2018-02-13
incoming: introduce a Sack object

jd:
Rather than manipulating integers, let's use a Sack object to abstract properly
our sack concept. This cleans the code in various places.

### DIFF
--- a/gnocchi/cli/manage.py
+++ b/gnocchi/cli/manage.py
@@ -95,7 +95,7 @@ def change_sack_size():
         LOG.error('Cannot change sack when non-empty backlog. Process '
                   'remaining %s measures and try again', remainder)
         return
-    LOG.info("Changing sack size to: %s", conf.sacks_number)
-    old_num_sacks = s.NUM_SACKS
-    s.set_storage_settings(conf.sacks_number)
-    s.remove_sack_group(old_num_sacks)
+    LOG.info("Removing current %d sacks", s.NUM_SACKS)
+    s.remove_sacks()
+    LOG.info("Creating new %d sacks", conf.sacks_number)
+    s.upgrade(conf.sacks_number)
--- a/gnocchi/incoming/__init__.py
+++ b/gnocchi/incoming/__init__.py
@@ -15,6 +15,9 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 import collections
+import functools
+import hashlib
+import operator
 
 import daiquiri
 import numpy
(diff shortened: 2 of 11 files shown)


gnocchi/cli/manage.py
```diff
@@ -95,7 +95,7 @@ def change_sack_size():
         LOG.error('Cannot change sack when non-empty backlog. Process '
                   'remaining %s measures and try again', remainder)
         return
-    LOG.info("Changing sack size to: %s", conf.sacks_number)
-    old_num_sacks = s.NUM_SACKS
-    s.set_storage_settings(conf.sacks_number)
-    s.remove_sack_group(old_num_sacks)
+    LOG.info("Removing current %d sacks", s.NUM_SACKS)
+    s.remove_sacks()
+    LOG.info("Creating new %d sacks", conf.sacks_number)
```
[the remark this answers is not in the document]
jd: Yes. `remove_sacks` now remove the current sack number based on the current configuration, so it has to be:

1. delete old sack based on the stored number of sacks
2. upgrade again to create the new sack

the old code was doing:

1. get the number of current sacks
2. create the new number of sacks and store it
3. remove the old number of sack based one 1

I think the new scheme is a bit cleaner/smoother.

jd:
Fix Ceph

gnocchi/incoming/__init__.py
```diff
@@ -166,11 +220,18 @@ def process_measure_for_metrics(metric_id):
...
+        return Sack(i, self.NUM_SACKS, self._get_sack_name(i))
+
     def sack_for_metric(self, metric_id):
-        return metric_id.int % self.NUM_SACKS
+        return self._make_sack(metric_id.int % self.NUM_SACKS)
 
-    def get_sack_name(self, sack):
-        return self.get_sack_prefix() % sack
+    def iter_sacks(self):
+        return six.moves.map(self._make_sack, six.moves.range(self.NUM_SACKS))
```
chungg: prefer list comprehension but meh.

gnocchi/incoming/file.py
```diff
@@ -49,16 +49,14 @@ def set_storage_settings(self, num_sacks):
         data = {self.CFG_SACKS: num_sacks}
         with open(os.path.join(self.basepath_tmp, self.CFG_PREFIX), 'w') as f:
             json.dump(data, f)
-        utils.ensure_paths([self._sack_path(i)
-                            for i in six.moves.range(self.NUM_SACKS)])
+        utils.ensure_paths(map(self._sack_path, self.iter_sacks()))
```
chungg: six.moves.map or list comprehension?

gnocchi/incoming/__init__.py
```diff
@@ -166,11 +220,18 @@ def process_measure_for_metrics(metric_id):
     def has_unprocessed(metric_id):
         raise exceptions.NotImplementedError
 
+    def _get_sack_name(self, number):
```
chungg: maybe drop leading '_' as you call this beyond this module.

gnocchi/incoming/redis.py
```diff
@@ -140,15 +140,15 @@ def iter_on_sacks_to_process(self):
         p = self._client.pubsub()
         db = self._client.connection_pool.connection_kwargs['db']
         keyspace = b"__keyspace@" + str(db).encode() + b"__:"
-        pattern = keyspace + self.SACK_PREFIX.encode() + b"*"
+        pattern = keyspace + self._get_sack_name("*").encode()
```
chungg: this is two different patterns. i assume it didn't matter? just asking because this is probably not tested anywhere iirc.

gnocchi/cli/manage.py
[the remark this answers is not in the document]
chungg: kk

gnocchi/incoming/__init__.py
jd: I don't call it outside, I call it in subclasses only. It's not meant to be public.

gnocchi/incoming/file.py
jd: really no difference here as it'll iterate on all so…

gnocchi/incoming/redis.py
jd: I don't understand the question, and this code path is tested in test_incoming.

jd:
Fix MySQL

jd:
Fix pep8

gnocchi/incoming/redis.py
chungg: hmm.. i was under impression greedy code was never tested

my comment was the pattern was  `__keyspace@<db>__incoming*` now it's, `__keyspace@<db>__incoming<number>-*`

i assume it doesn't matter if there's a test.

jd:
Fix Py3 encoding

gnocchi/incoming/redis.py
jd: Yeah the new pattern is actually better in this regard. :)

Test: https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/tests/test_incoming.py#L34-L66

PR #739, opened 2018-02-14
storage: replace _get_unaggregated_timeserie by _get_or_create_unaggregated_timeseries

jd:
This replaces this method in each storage driver by a new one that can retrieve
several unaggregated timeseries at the same time. The storage engine does not
leverage it and only passes a list of one metric… for now.

### DIFF
--- a/gnocchi/storage/ceph.py
+++ b/gnocchi/storage/ceph.py
@@ -183,12 +183,17 @@ class CephStorage(storage.StorageDriver):
         return (('gnocchi_%s_none' % metric.id)
                 + ("_v%s" % version if version else ""))
 
-    def _get_unaggregated_timeserie(self, metric, version=3):
+    def _get_or_create_unaggregated_timeseries_unbatched(
+            self, metric, version=3):
         try:
-            return self._get_object_content(
+            contents = self._get_object_content(
                 self._build_unaggregated_timeserie_path(metric, version))
         except rados.ObjectNotFound:
-            raise storage.MetricDoesNotExist(metric)
+            self._create_metric(metric)
+        else:
+            # _create_metric writes "" so replace it by None to indicate
+            # emptiness instead.
+            return contents or None
 
     def _store_unaggregated_timeserie(self, metric, data, version=3):
         self.ioctx.write_full(
--- a/gnocchi/storage/file.py
+++ b/gnocchi/storage/file.py
@@ -31,6 +31,12 @@ OPTS = [
                help='Path used to store gnocchi data files.'),
 ]
 
+# Python 2 compatibility
+try:
+    FileNotFoundError
+except NameError:
+    FileNotFoundError = None
+
 
 class FileStorage(storage.StorageDriver):
     WRITE_FULL = True
@@ -95,15 +101,21 @@ class FileStorage(storage.StorageDriver):
         with open(dest, "wb") as f:
             f.write(data)
 
-    def _get_unaggregated_timeserie(self, metric, version=3):
+    def _get_or_create_unaggregated_timeseries_unbatched(
+            self, metric, version=3):
         path = self._build_unaggregated_timeserie_path(metric, version)
         try:
             with open(path, 'rb') as f:
                 return f.read()
+        except FileNotFoundError:
+            pass
         except IOError as e:
-            if e.errno == errno.ENOENT:
-                raise storage.MetricDoesNotExist(metric)
-            raise
+            if e.errno != errno.ENOENT:
+                raise
+        try:
+            self._create_metric(metric)
+        except storage.MetricAlreadyExists:
+            pass
 
     def _list_split_keys(self, metric, aggregation, granularity, version=3):
         try:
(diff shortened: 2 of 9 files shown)


jd:
Fix pep8 and Ceph

jd:
Fix Ceph

jd:
Moot point IMHO. This could get gigabytes large already with only one metric. If you want to fix that, it'd be great, but it's another story.

chungg:
> Moot point IMHO. This could get gigabytes large already with only one metric. If you want to fix that, it'd be great, but it's another story.

that strengthens the reason to think about this. one metric with 1GB of backlog is bad but manageable. 100 or 1000 metrics of 1GBs is "good bye server"... we should really consider a batch size as we move down this route.

i don't think i ever benchmarked a sack with 100s of metrics but i'd be interested how much time is storage IO and how much of the time is indexer IO... my profiles have always been small but that indexer IO is usually magnitudes more than storage IO.

jd:
> that strengthens the reason to think about this. one metric with 1GB of backlog is bad but manageable. 100 or 1000 metrics of 1GBs is "good bye server"... we should really consider a batch size as we move down this route.

1 GB of backlog? I don't think fetching and storing a *single* object (whatever the driver) of 1 GB is gonna work anyway.

I agree we need to solve that point at some point by limiting the size of the back window, but this is a different issue than what I'm trying to achieve here. This problem *already* exists. This branch is not trying to solve it. Maybe it'll make it more visible, but that's a not a good reason to block this. It's a good reason to solve it.

jd:
I forgot one point:

> we should really consider a batch size as we move down this route.

Yes, that's exactly what I thought about (for different reasons but it maps to this one). At this stage I consider the sack to be the batch, because that seems like the simplest way to scale. Maybe we'll need a dynamic number of sacks in the future to adjust.

chungg:
> Maybe it'll make it more visible

agreed. :)

> if you process one sack with measures for 10k metrics in it, the problem is not gonna be the indexer. Replacing the 10k get of the unagg ts by one is a huge saver.

well the docs specifically guide users to have sacks with roughly <500 metrics. my comment was really a question so i could understand better. while i agree the hypothetical case of saving/parallelising 9999GETs will be beneficial, all i know for sure is the small case where it's not really saving much. so yeah, just a question in case you had information i didn't have.

jd:
@chungg btw, this code is only enabling multiple loads, but it does not activate it (yet).

gnocchi/storage/file.py
```diff
@@ -95,15 +101,26 @@ def _store_unaggregated_timeserie(self, metric, data, version=3):
...
+        for metric in metrics:
+            path = self._build_unaggregated_timeserie_path(metric, version)
+            try:
+                with open(path, 'rb') as f:
+                    ts[metric] = f.read()
+            except FileNotFoundError:
+                pass
+            except IOError as e:
+                if e.errno != errno.ENOENT:
+                    raise
```
chungg: this seems wrong that one metric can cause all metrics to fail. that concern can probably be applied to all drivers. we should really be ensuring any error by one doesn't affect another.

gnocchi/storage/s3.py
```diff
@@ -216,16 +213,20 @@ def _build_unaggregated_timeserie_path(metric, version):
...
+    def _get_or_create_unaggregated_timeseries(self, metrics, version=3):
+        ts = {}
+        for metric in metrics:
+            key = self._build_unaggregated_timeserie_path(metric, version)
+            try:
+                response = self.s3.get_object(
+                    Bucket=self._bucket_name, Key=key)
+            except botocore.exceptions.ClientError as e:
+                if e.response['Error'].get('Code') != "NoSuchKey":
+                    raise
```
chungg: same thing. one should not affect all

gnocchi/storage/swift.py
```diff
@@ -186,16 +186,24 @@ def _list_split_keys(self, metric, aggregation, granularity, version=3):
...
+    def _get_or_create_unaggregated_timeseries(self, metrics, version=3):
+        ts = {}
+        for metric in metrics:
+            try:
+                headers, contents = self.swift.get_object(
+                    self._container_name(metric),
+                    self._build_unaggregated_timeserie_path(version))
+            except swclient.ClientException as e:
+                if e.http_status != 404:
+                    raise
```
chungg: same

jd:
Pull-request updated, HEAD is now 472daf15c5fd27861c8f1a1a64e855d49cb64bf8

gnocchi/storage/ceph.py
```diff
@@ -183,12 +183,17 @@ def _build_unaggregated_timeserie_path(metric, version):
...
-            return self._get_object_content(
+            contents = self._get_object_content(
                 self._build_unaggregated_timeserie_path(metric, version))
         except rados.ObjectNotFound:
-            raise storage.MetricDoesNotExist(metric)
+            self._create_metric(metric)
+        else:
+            # _create_metric writes "" so replace it by None to indicate
+            # emptiness instead.
+            return contents or None
```
chungg: curious, but why would contents be "". shouldn't every metric we create have something in or is this just a handling a case where it fails for some reason?

jd: Yes: if it fails to store for some reason, then _create_metric has written "" and then we need to return `None`. I agree that this should not happen if everything goes well :)

PR #743, opened 2018-02-14
add telemetry logging

chungg:
warn if backwindow is ~4MB. 4MB is based on default object size
used in ceph rados bench tool. 12.6B/point used as average size
based on existing BoundTimeSerie random benchmark.

### DIFF
--- a/gnocchi/storage/__init__.py
+++ b/gnocchi/storage/__init__.py
@@ -31,6 +31,10 @@ OPTS = [
     cfg.StrOpt('driver',
                default='file',
                help='Storage driver to use'),
+    cfg.BoolOpt('enable_telemetry', default=True,
+                help='Occasionally logs stats regarding service health. '
+                     'Details should be reported to developers at '
+                     'https://github.com/gnocchixyz/gnocchi/issues'),
 ]
 
 
@@ -106,8 +110,8 @@ def get_driver(conf):
 
 class StorageDriver(object):
 
-    @staticmethod
-    def __init__(conf):
+    def __init__(self, conf):
+        self.enable_telemetry = conf.enable_telemetry
         pass
 
     @staticmethod
@@ -443,6 +447,14 @@ class StorageDriver(object):
             current_first_block_timestamp = None
         else:
             current_first_block_timestamp = ts.first_block_timestamp()
+            # log if backwindow is larger than ~4MB to anticipate potential
+            # performance issue
+            if self.enable_telemetry and ts.ts.size > 320000:
+                LOG.warning("TELEMETRY: The defined backwindow size for "
+                            "metric %s may negatively affect performance: "
+                            "(size: %s, back_window: %s, granularities: %s)",
+                            metric.id, ts.ts.size, back_window,
+                            [d.granularity for d in definition])
 
         # NOTE(jd) This is Python where you need such
         # hack to pass a variable around a closure,


gnocchi/storage/__init__.py
```diff
@@ -443,6 +447,14 @@ def compute_and_store_timeseries(self, metric, measures):
             current_first_block_timestamp = None
         else:
             current_first_block_timestamp = ts.first_block_timestamp()
+            # log if backwindow is larger than ~4MB to anticipate potential
+            # performance issue
+            if self.enable_telemetry and ts.ts.size > 320000:
```
jd: Could you use `len(ts)`? It's a detail but it sounds cleaner.

gnocchi/storage/__init__.py
```diff
@@ -31,6 +31,10 @@
     cfg.StrOpt('driver',
                default='file',
                help='Storage driver to use'),
+    cfg.BoolOpt('enable_telemetry', default=True,
```
jd: I don't understand the reasoning behind this option. It sounds kinda useless to me but maybe there's something I miss.

And the option name is completely "misleading" to me.

chungg: if i don't add this, and you actually hit the warning, the warning will get thrown forever as there's no way you can disable the warning... i could add a cache to capture stuff we've already logged but a metric can be handled by multiple workers so it'll still log multiple times.... and if we ever expand telemetry, it'll be more stuff we need to cache.

i'm open to another approach or rename.

jd: I see, thanks for the explanation.

I'd say add a setting to configure this 320000 size so it can be set to `None` to be disabled. Something like `warn_backwindow_size`.

And I hope we'll be able to kill that option later. :)

gnocchi/storage/__init__.py
```diff
@@ -443,6 +447,14 @@ def compute_and_store_timeseries(self, metric, measures):
             current_first_block_timestamp = None
         else:
             current_first_block_timestamp = ts.first_block_timestamp()
+            # log if backwindow is larger than ~4MB to anticipate potential
+            # performance issue
+            if self.enable_telemetry and ts.ts.size > 320000:
+                LOG.warning("TELEMETRY: The defined backwindow size for "
```
jd: keep that as a warning but don't use "TELEMETRY", that's really confusing to me.

And use a sentence like:

"The defined backwindow size for metric … may negatively affect performance as it is bigger than %d (see warn_backwindow_size option)"

or something like that so it's easy for operators to get what this warning is there

gnocchi/storage/__init__.py
chungg: will do

gnocchi/storage/__init__.py
chungg: well operators can't do anything about it. this technically means nothing to them. this is just for us developers to understand how users are using Gnocchi (like how your browser/os captures metrics in background and sends it back). i'm not even sure 4MB is that bad, i just put it because i know i don't care if it's smaller than that. :)

i'm ok with putting this as an INFO log as well.

jd: Operators can limit the backwindow in the archive policy to limit this effect at least.

I'm good for info too.

chungg:
alternatively, i could figure a way to hijack and create a gnocchi-system resource_type which captures metrics we want but i imagine this is more intrusive and probably isn't worth it for just a single metric.

gnocchi/storage/__init__.py
chungg: more specifically, i mean, i don't want to suggest to operators to limit something that may or may not be an issue. i'll be honest, i don't know if 4MB is bad or good... it's just a value i picked based on ceph benchmark tool.

PR #745, opened 2018-02-15
rest: stop using set to validate archive policies

jd:
voluptuous does not know how to use set and now wants to check that the default
value validates against the validators.

### DIFF
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -222,10 +222,10 @@ def strtobool(varname, v):
         abort(400, "Unable to parse `%s': %s" % (varname, six.text_type(e)))
 
 
-RESOURCE_DEFAULT_PAGINATION = ['revision_start:asc',
-                               'started_at:asc']
+RESOURCE_DEFAULT_PAGINATION = [u'revision_start:asc',
+                               u'started_at:asc']
 
-METRIC_DEFAULT_PAGINATION = ['id:asc']
+METRIC_DEFAULT_PAGINATION = [u'id:asc']
 
 
 def get_pagination_options(params, default):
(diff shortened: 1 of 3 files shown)


jd:
Fix Python 2

jd:
Pull-request updated, HEAD is now 5fb4b701abbdf24c95f2026d28e31f260e1d686b

jd:
Pull-request updated, HEAD is now eea564d4e0ba81ba43311ffedf112ce24e06018d

gnocchi/rest/api.py
```diff
@@ -324,7 +324,7 @@ def post(self):
         enforce("create archive policy", {})
         # NOTE(jd): Initialize this one at run-time because we rely on conf
         conf = pecan.request.conf
-        valid_agg_methods = (
+        valid_agg_methods = list(
```
chungg: how come you added list here but not in the backports?

jd: The backport works fine and is doing the same thing (in a different line) it seems but this one still fails for docs-gnocchi.xyz target and i'm trying to understand why

jd:
Pull-request updated, HEAD is now 2ba717a13dd291635554ab8d602c30f3dca6d79f

jd:
Pull-request updated, HEAD is now b657ef56c6b91f058a40534d91179f3611608f09

jd:
Pull-request updated, HEAD is now 889c53f052c0087ff66bb1650e77d3ff3f3b4627

jd:
I'm tired

sileht:
I passed the voluptuous issue, but stable branches looks not generated, so I progress but ...

jd:
@sileht what do you mean not generated?

sileht:
The remotes/origin/stable/XXX was missing, so multiversion sphinx wasn't building them. I figure out why. It should be fixed soon.

jd:
that looks great, good job @sileht. that was a hard one. go ahead and approve (I can't since it's my patch)

PR #762, opened 2018-02-20
storage: change _delete_metric_measures to _delete_metric_splits

jd:
This new method allows to delete a bunch of splits at the same time for a
single metric, batching delete.

### DIFF
--- a/gnocchi/storage/ceph.py
+++ b/gnocchi/storage/ceph.py
@@ -91,19 +91,21 @@ class CephStorage(storage.StorageDriver):
             self.ioctx.operate_write_op(
                 op, self._build_unaggregated_timeserie_path(metric, 3))
 
-    def _delete_metric_measures(self, metric, key, aggregation, version=3):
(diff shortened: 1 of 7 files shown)


gnocchi/storage/ceph.py
```diff
@@ -91,19 +91,20 @@ def _store_metric_splits(self, metric, keys_and_data_and_offset,
...
-            # anyway.
-            pass
+                try:
+                    self.ioctx.remove_object(name)
+                except rados.ObjectNotFound:
+                    # It's possible that we already remove that object and then
+                    # crashed before removing it from the OMAP key list; then
+                    # no big deal anyway.
+                    pass
+                self.ioctx.remove_omap_keys(op, (name,))
```
chungg: why not just pass in a tuple of all names? or does rados figure this out?

jd: You're right, I missed that!

jd:
Fix Ceph

PR #768, opened 2018-02-26
carbonara: make SplitKey carry aggregation method

jd:
Why not strictly used by the SplitKey, it actually makes sense as
AggregatedTimeSerie splits key are not really compatible with different
aggregation method. This will come handy to pass less arguments in the future.

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -452,23 +452,31 @@ class SplitKey(object):
     are regularly spaced.
     """
 
+    __slots__ = (
+        'key',
+        'sampling',
+        'aggregation_method',
+    )
+
     POINTS_PER_SPLIT = 3600
 
-    def __init__(self, value, sampling):
+    def __init__(self, value, sampling, aggregation_method):
         if isinstance(value, SplitKey):
             self.key = value.key
         else:
             self.key = value
 
         self.sampling = sampling
+        self.aggregation_method = aggregation_method
 
     @classmethod
-    def from_timestamp_and_sampling(cls, timestamp, sampling):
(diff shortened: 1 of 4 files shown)


jd:
Pull-request updated, HEAD is now b7d734b75704e1af2ca95511a46162615ec6c3c9

chungg:
tbh, this is confusing why we're adding non-temporal data to a time-specific object. to me, a split key is independent to a metric let alone a specific aggregate of a metric... more like a generic helper. for example the next value is just a computation of the next numerical split, regardless if it exists for metric or not.

jd:
I agree and that's why I did not do not it in the first place. This is strictly true if you keep focused on the `SplitKey` as an independent identity.

Then, there are the `SplitKey`s emitted by the `split` method of `AggregatedTimeSerie`. They are the same, except that an `AggregatedTimeSerie` has an aggregation method and in that case, it makes sense to have a different `SplitKey`, i.e. an `AggregatedSplitKey` that would carry the aggregation method to so that it can be serialized with the split and used directly as a key. Basically a driver, could do `store(str(my_aggregatedsplitkey), split.serialize())`.

However, since `SplitKey` are only used by `AggregatedTimeSerie`, I choose not to complicate too much the logic and use `SplitKey` directly to carry that information.

I'm not strongly opiniated on this TBH. I agree it's bending a bit the rule. I can also drop it and just keep carrying `(key, aggregation_method)` everywhere rather than just `key`.

chungg:
i have a slight preference to move without it... maybe its usefulness will become more apparent as we merge more of your changes.

PR #769, opened 2018-02-26
carbonara: make fetch() return an AggregatedTimeSerie

jd:
The current return format is very specific to the Gnocchi API, whereas there
should not be any need to do the formatting at this low-level stage.


jd:
Pull-request updated, HEAD is now 00eea856b76e118c8952e9f7bb13074985e8bda6

PR #772, opened 2018-02-26
carbonara: allow eq and ne comparison to SplitKey

jd:
While you can't sort and compare SplitKey with different sampling, you can say
if they are equals or not easily.

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -491,6 +491,10 @@ class SplitKey(object):
     def _compare(self, op, other):
         if isinstance(other, SplitKey):
             if self.sampling != other.sampling:
+                if op == operator.eq:
+                    return False
+                if op == operator.ne:
+                    return True
                 raise TypeError(
                     "Cannot compare %s with different sampling" %
                     self.__class__.__name__)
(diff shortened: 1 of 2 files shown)


gnocchi/carbonara.py
```diff
@@ -490,6 +490,11 @@ def __hash__(self):
 
     def _compare(self, op, other):
         if isinstance(other, SplitKey):
+            if op == operator.eq:
+                return (op(self.key, other.key) and
+                        op(self.sampling, other.sampling))
+            if op == operator.ne:
+                return not self._compare(operator.eq, other)
             if self.sampling != other.sampling:
```
chungg: can't you just check if op is operator.eq or op is operator.ne and return False as i believe the only scenario is if sampling is different it's automatically not equivalent.

jd: I don't think so IIUC since the final call is `op(self.key, other.key)` and does not include the sampling in comparison then.

jd: That'd (almost) work too. Updating the PR.

jd:
Pull-request updated, HEAD is now da7e4f3c19ab24ef7c22949f46d3813fe7d6540e

PR #775, opened 2018-02-27
storage: list_split_keys can now list several aggregations at once

jd:
This should help to limit the number of calls make to list_split_keys in the
future.

### DIFF
--- a/gnocchi/storage/ceph.py
+++ b/gnocchi/storage/ceph.py
@@ -14,9 +14,12 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
+import collections
 
 from oslo_config import cfg
+import six
 
+from gnocchi import carbonara
 from gnocchi.common import ceph
 from gnocchi import storage
 from gnocchi import utils
@@ -153,7 +156,7 @@ class CephStorage(storage.StorageDriver):
             else:
                 raise storage.MetricDoesNotExist(metric)
 
-    def _list_split_keys(self, metric, aggregation, granularity, version=3):
+    def _list_split_keys(self, metric, aggregations, version=3):
         with rados.ReadOpCtx() as op:
             omaps, ret = self.ioctx.get_omap_vals(op, "", "", -1)
             try:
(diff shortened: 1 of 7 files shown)


jd:
Fix Ceph

jd:
Pull-request updated, HEAD is now d92d6022c5f280c782c9893a494caea1dcded2a8

gnocchi/storage/redis.py
```diff
@@ -111,14 +110,36 @@ def _get_or_create_unaggregated_timeseries(self, metrics, version=3):
         }
         return ts
 
-    def _list_split_keys(self, metric, aggregation, granularity, version=3):
+    def _list_split_keys(self, metric, aggregations, version=3):
```
sileht: Do you plan to batch also metrics here ? like a dict {metric_id: [aggregations]}

jd: Very likely. I'm not there yet, but it'd make sense when we batch metric processing per sack.

gnocchi/storage/ceph.py
```diff
@@ -173,13 +176,25 @@ def _list_split_keys(self, metric, aggregation, granularity, version=3):
...
+            if not raw_keys:
+                return keys
+            zipped = list(zip(*raw_keys))
+            k_timestamps = utils.to_timestamps(zipped[2])
+            k_methods = zipped[3]
+            k_granularities = list(map(utils.to_timespan, zipped[4]))
+
+            for timestamp, method, granularity in six.moves.zip(
+                    k_timestamps, k_methods, k_granularities):
+                for aggregation in aggregations:
```
chungg: this seems simpler to just keep original code, and just build a set of (granularity, method) pairs to represent Agg, then you can just check if omap is in the set rather than all these loops?

chungg:
or maybe just make Aggregate hash to something comparable?

gnocchi/storage/ceph.py
jd: I don't want to abuse, but i'm not sure I get it; would you mind writing the code snippet you have in mind? :)

chungg: like this: https://gist.github.com/chungg/46bb333e6a1f9dc862d9b799ffe6a874

disclaimer: i never tested.

chungg: oh, you moved the numpy broadcasting into here...

jd: This would not work as the timestamp needs to be parsed. Your code does not.

My code here converts all the timestamps in one single call (L186). I did that to make it fast because I knew otherwise you would (rightfully) complain. Yes I always think of you when I code. :)

chungg: we can still convert this to "if x in y" rather than a for loop no? or at the very least, add a break once we fined match all the additional loops concern me but it might be only way as you said.

jd: We list and parse everything, and then iterate over every key and check if it's interesting for any aggregation.

We can write this code different with a `if x in y` but this is just syntatic sugar, as it'll iterate in `y` anyway. There's no magic here.

chungg: that's fair regarding syntax but we can still break once we find match as i don't see how there can be multiple matches

jd: Oh you're right I think. I missed that.

jd:
Add break in Ceph for loop

jd:
Add break

jd:
Fix s3

PR #776, opened 2018-02-27
carbonara: use Aggregation object to create AggregatedTimeSerie

jd:
This makes usage of the Aggregation object directly into Carbonara so it's
attached to an AggregatedTimeSerie.


jd:
Pull-request updated, HEAD is now 2af1506680c1c457af2f149abf2a1e9ac15f032d

gnocchi/carbonara.py
```diff
@@ -597,54 +608,49 @@ def split(self):
...
-            yield (SplitKey(key, self.sampling),
-                   AggregatedTimeSerie(self.sampling, self.aggregation_method,
-                                       self[start:end]))
+            yield (SplitKey(key, self.aggregation.granularity),
+                   AggregatedTimeSerie(self.aggregation, self[start:end]))
             start = end
 
     @classmethod
-    def from_timeseries(cls, timeseries, sampling, aggregation_method):
+    def from_timeseries(cls, timeseries, aggregation):
```
chungg: just for discussion, there are places where we clearly require sampling to be defined but as we start passing in Aggregation object, we lose the ability to enforce this. any thoughts how we can handle this? are we relying completely on tests to validate this?

gnocchi/tests/test_carbonara.py
```diff
@@ -105,21 +105,6 @@ def test_duplicate_timestamps(self):
 
 
 class TestAggregatedTimeSerie(base.BaseTestCase):
-    @staticmethod
-    def test_base():
```
chungg: you deleting this because it tests nothing?

gnocchi/carbonara.py
jd: I don't think this patch broke any existing enforcement, so AFAIU it's already the case that we solely rely on the tests for this. If you have a more precise example I may have a more precise answer. :)

gnocchi/tests/test_carbonara.py
jd: Yes. It existed because the `3` or `"4s"` once were converted at `__init__` time to a `numpy.timedelta64` (or even a pandas type) but that's not the case anymore as this is done way before in the current code (which is faster). So rather than fixing it stupidly I went ahead and just dropped it.

gnocchi/carbonara.py
chungg: works for me. let's see if it comes back to bite us

PR #779, opened 2018-03-01
doc: Update setuptools for gnocchi.xyz

sileht:


sileht:
This is really tested only at 22 every hour, I merge it to see if it solve our issue.

PR #782, opened 2018-03-01
storage: do not return empty data set on non-existing metric

jd:
This decision is now available for the upper layer.

### DIFF
--- a/gnocchi/rest/aggregates/processor.py
+++ b/gnocchi/rest/aggregates/processor.py
@@ -52,11 +52,16 @@ class MetricReference(object):
 
 
 def _get_measures_timeserie(storage, ref, granularity, *args, **kwargs):
-    return (ref, storage._get_measures_timeserie(
-        ref.metric,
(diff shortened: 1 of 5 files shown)


jd:
Fix S3

jd:
Fix 404 handling

chungg:
what's "the upper layer"? it seems better to return 404 if something doesn't exist no?

jd:
> what's "the upper layer"? it seems better to return 404 if something doesn't exist no?

The upper layer is anything that is using it such as the REST API.

Returning 404 would break the API.

chungg:
sorry, i think i'm missing original context but if i ask for a non-existing metric, shouldn't i get 404? this seems to make sense to me.

jd:
That's already the case and this patch does not change this. The REST API checks the existence of metric via the indexer, not via storage.

What this patch changes is that if a metric does not exist in storage, the storage driver does not return `[]` but raises an error. It's really unrelated to the 404 you're talking about.

PR #785, opened 2018-03-01
doc: don't build sdist

sileht:
By default tox built a sdist with the python OS packages. We already
skip install/develop phase and manually use "pip install -e"
The sdist is useless and removing it allow to run tox on old OS
that doesn't have a recent setuptools.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,7 @@
 [tox]
 minversion = 2.4
 envlist = py{35,27}-{postgresql,mysql}{,-file,-swift,-ceph,-s3},pep8
+skipsdist = True
 
 [testenv]
 skip_install = True


setup.py
```diff
@@ -17,5 +17,5 @@
 import setuptools
 
 setuptools.setup(
-    setup_requires=['pbr', 'setuptools>=20.6.8'],
+    setup_requires=['pbr', 'setuptools>=36'],
```
jd: is this in the same change?

sileht: fixed that was a test

sileht:
Pull-request updated, HEAD is now a6402e0cf514dfa7cd7956e147a5aadcccab8010

PR #787, opened 2018-03-03
storage: introduce get_aggregated_measures and deprecate get_measures

jd:


jd:
Fix sorting

gnocchi/storage/__init__.py
```diff
@@ -199,35 +200,55 @@ def _version_check(name, v):
...
         :param to timestamp: The timestamp to get the measure to.
         :param resample: The granularity to resample to.
         """
-        agg_timeseries = utils.parallel_map(
-            self._get_measures_timeserie,
-            ((metric, ag, from_timestamp, to_timestamp)
-             for ag in aggregations))
+        timeseries = self.get_aggregated_measures(
+            metric, aggregations,
+            from_timestamp=None, to_timestamp=None)
```
chungg: how come you don't pass in timestamps here?

gnocchi/storage/__init__.py
```diff
@@ -199,35 +200,55 @@ def _version_check(name, v):
...
-                  for ts in aggts]))
-            for aggmethod, aggts
-            in itertools.groupby(agg_timeseries,
-                                 lambda v: v.aggregation.method)
+                   in timeseries[agg].fetch(from_timestamp, to_timestamp)]
+                  for agg in sorted(aggs,
+                                    key=ATTRGETTER_GRANULARITY,
+                                    reverse=True)]))
+            for aggmethod, aggs in itertools.groupby(timeseries.keys(),
+                                                     ATTRGETTER_METHOD)
```
chungg: tbh, this is some crazy looking looping but i'm guessing you're going to remove this later on?

gnocchi/storage/__init__.py
jd: Good catch.

gnocchi/storage/__init__.py
jd: It's the same as before and that's what the API returns actually. It should be a bit simpler when this method gets entirely removed and moved to the API>

jd:
Fix timestamp passing

gnocchi/storage/__init__.py
chungg: works for me.

PR #792, opened 2018-03-05
storage: stop raising AggregationDoesNotExist on measures retrieval

jd:
Raising an exception at this point is damageable as when multiple SplitKeys are
being fetched at the same time, only one being absent will cancel the call. The
caller is already solid and know show to handle None as a return value so just
use None to indicate the split is non-existing.

### DIFF
--- a/gnocchi/storage/ceph.py
+++ b/gnocchi/storage/ceph.py
@@ -153,10 +153,8 @@ class CephStorage(storage.StorageDriver):
         except rados.ObjectNotFound:
             if self._object_exists(
                     self._build_unaggregated_timeserie_path(metric, 3)):
-                raise storage.AggregationDoesNotExist(
-                    metric, aggregation.method, key.sampling)
-            else:
-                raise storage.MetricDoesNotExist(metric)
+                return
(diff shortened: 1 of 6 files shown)


gnocchi/storage/ceph.py
```diff
@@ -153,10 +153,8 @@ def _get_measures_unbatched(self, metric, key, aggregation, version=3):
         except rados.ObjectNotFound:
             if self._object_exists(
                     self._build_unaggregated_timeserie_path(metric, 3)):
```
chungg: do we still need these checks? if ObjectNotFound, isn't it a given that it should raise MetricDoesNotExist?

jd: No, it just means there's no data for this particular split. When a split is rewritten, the storage engine tries to read it and if data are there it updates. But you're right that I think we can simplify this again.

If that's ok I'll do that in a subsequent patch. I saw some things that we can simplify with that.

jd:
Pull-request updated, HEAD is now e24b8db12b8467d3030f28b6d84eb6e65de4cffd

PR #793, opened 2018-03-05
storage: remove metric existence check when retrieving splits

jd:
Splits are either retrieve for reading from:

- the REST API, where the splits are actually listed before being retrieved, we
  don't care about the metric existence or not, None is enough.

- on rewriting the split (WRITE_FULL) and in that case we know the metric
  exists


jd:
Add more tests

PR #794, opened 2018-03-05
storage: _add_measures can handle multiple aggregations at the same time

jd:
This allows to batch storage and deletion of splits for a metric.

### DIFF
--- a/gnocchi/storage/__init__.py
+++ b/gnocchi/storage/__init__.py
@@ -166,15 +166,16 @@ class StorageDriver(object):
             ((metric, data, version) for metric, data in metrics_and_data))
 
     @staticmethod
-    def _store_metric_splits(metric, keys_and_data_and_offset, aggregation,
-                             version=3):
+    def _store_metric_splits(metric, keys_aggregations_data_offset, version=3):
         """Store metric split.
 
         Store a bunch of splits for a metric.
 
         :param metric: The metric to store for
-        :param keys_and_data_and_offset: A list of (key, data, offset) tuples
-        :param aggregation: The aggregation method concerned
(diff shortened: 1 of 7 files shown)


jd:
Pull-request updated, HEAD is now 030970262ef57bbfa0fd9bac2185b6a81980db4c

jd:
Pull-request updated, HEAD is now 6a199f9edfcdbaf3750f119f07a097352e36eded

jd:
Fix comment

jd:
Pull-request updated, HEAD is now aaf5375875a56547bc643050476f1609ef25dfed

gnocchi/storage/__init__.py
```diff
@@ -329,118 +331,117 @@ def _store_timeserie_splits(self, metric, keys_and_splits,
         # stored in the case that we need to rewrite them fully.
         # First, fetch all those existing splits.
         existing_data = self._get_splits_and_unserialize(
-            metric, [(key, aggregation) for key in keys_to_rewrite])
+            metric, [(key, split.aggregation)
```
chungg: if the split has the aggregation already, why we passing in the aggregation?

jd: The answer is line 367 ;)

chungg: oh i mean that aggregation is never used anywhere in this method. it only ever uses split.aggregation.

jd: Again, aggregation is only passed because it's used as the key of the dict as explained in the comment L367.

chungg: oh. it's from the other method.

PR #796, opened 2018-03-06
storage: fix corruption scenario

jd:
If data corruption occurs, the _get_splits_and_unserialize() method might not
return the same number of arguments that it received. It's not a big deal when
concatenating the splits to retrieve the data for reading. However, when
requesting a bunch of splits for rewrite, that's a problem because the list is
zipped with other data on iteration, possibly corrupting the results.


jd:
Pull-request updated, HEAD is now 5b2f5494eafee033e5a0434c28107512643fc153

PR #797, opened 2018-03-06
Pull request for fix-aggregation-in-split-retrieval

jd:
storage: add a basic test for _get_splits_and_unserialize
carbonara: pass Aggregation object to AggregatedTimeSerie.unserialize


jd:
The answer is in the commit message :) The aggregation field on those timeseries is never used: those AggregatedTimeSerie are just concatened together to create the big real AggregatedTimeSeries which *has* the correct aggregation set :)

jd:
Rebased

PR #800, opened 2018-03-08
api: fix double = bug in search

sileht:


sileht:
Sorry for the noise, I just push it to test and see if I can reproduce in CI.

sileht:
So the bug is on grafana side :(

PR #802, opened 2018-03-08
incoming: introduce process_measures_for_sack

jd:
This adds a new method process_measures_for_sack to incoming storage driver. It
allows to read an entire sack for new measure rather than individual metrics.
This avoids doing 2 listing to process new metrics and only does one to load
the measures.

The process_new_measures(metrics) is kept for the refresh_metric() use case for
now. Some refactoring might be possible after this patch.

The S3 storage driver is a bit modified to store the incoming measures in
<sack>/<metric>/ rather than <sack><metric>/ so it's easier to list incoming
sacks and metrics.

### DIFF
--- a/gnocchi/incoming/ceph.py
+++ b/gnocchi/incoming/ceph.py
@@ -13,6 +13,7 @@
 # under the License.
 from collections import defaultdict
 import contextlib
+import daiquiri
 import datetime
 import json
 import uuid
@@ -24,6 +25,8 @@ from gnocchi.common import ceph
 from gnocchi import incoming
 
 rados = ceph.rados
+
+LOG = daiquiri.getLogger(__name__)
 
 
 class CephStorage(incoming.IncomingDriver):
@@ -212,5 +215,33 @@ class CephStorage(incoming.IncomingDriver):
                 # NOTE(sileht): come on Ceph, no return code
                 # for this operation ?!!
                 self.ioctx.remove_omap_keys(op, tuple(keys.keys()))
+                self.ioctx.operate_write_op(op, str(sack),
+                                            flags=self.OMAP_WRITE_FLAGS)
+
+    @contextlib.contextmanager
+    def process_measures_for_sack(self, sack):
+        measures = defaultdict(self._make_measures_array)
+        omaps = self._list_keys_to_process(
+            sack, prefix=self.MEASURE_PREFIX + "_")
+        for k, v in six.iteritems(omaps):
+            try:
+                metric_id = uuid.UUID(k.split("_")[1])
+            except (ValueError, IndexError):
+                LOG.warning("Unable to parse measure object name %s",
+                            k)
+                continue
+            measures[metric_id] = numpy.concatenate(
+                (measures[metric_id], self._unserialize_measures(k, v))
+            )
+
+        yield measures
+
+        # Now clean omap
+        processed_keys = tuple(omaps.keys())
+        if processed_keys:
+            with rados.WriteOpCtx() as op:
+                # NOTE(sileht): come on Ceph, no return code
+                # for this operation ?!!
+                self.ioctx.remove_omap_keys(op, tuple(processed_keys))
                 self.ioctx.operate_write_op(op, str(sack),
                                             flags=self.OMAP_WRITE_FLAGS)
--- a/gnocchi/incoming/file.py
+++ b/gnocchi/incoming/file.py
@@ -20,11 +20,14 @@ import shutil
 import tempfile
 import uuid
 
+import daiquiri
 import numpy
 import six
 
 from gnocchi import incoming
 from gnocchi import utils
+
+LOG = daiquiri.getLogger(__name__)
 
 
 class FileStorage(incoming.IncomingDriver):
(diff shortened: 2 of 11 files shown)


jd:
Fix Ceph maybe

jd:
@sileht This fails because of the Ceph driver. My patch has probably a mistake but I don't see it. I could use a different pair of eyes I think.
(I can't debug it locally as I don't have Ceph :( )

jd:
Fix Ceph parsing

jd:
Different fix

jd:
Fix Ceph

jd:
Fix OMAP iteration

jd:
Pull-request updated, HEAD is now 37d22948983a44eb7b1fc922cd2f4dc6c68ee6a3

jd:
Pull-request updated, HEAD is now 74af3704eda6ab4432158ed0d30924d445faad98

gnocchi/incoming/ceph.py
```diff
@@ -214,3 +217,32 @@ def process_measure_for_metrics(self, metric_ids):
...
+    def process_measures_for_sack(self, sack):
+        measures = defaultdict(self._make_measures_array)
+        processed_keys = set()
+        omaps = self._list_keys_to_process(
+            sack, prefix=self.MEASURE_PREFIX + "_")
+        for k, v in six.iteritems(omaps):
+            try:
+                metric_id = uuid.UUID(k.split("_")[1])
+            except (ValueError, IndexError):
+                LOG.warning("Unable to parse measure object name %s",
```
chungg: is this suppose to handle scenario where user put non-sense into measures path?

gnocchi/incoming/file.py
```diff
@@ -179,3 +182,30 @@ def process_measure_for_metrics(self, metric_ids):
...
+        measures = {}
+        processed_files = {}
+        for metric_id in self._list_target(self._sack_path(sack)):
+            try:
+                metric_id = uuid.UUID(metric_id)
+            except ValueError:
+                LOG.error("Unable to parse %s as an UUID, ignoring metric",
+                          metric_id)
+                continue
+            files = self._list_measures_container_for_metric_str(
```
chungg: this method seems to ask for metric_id as a string but we give it as a uuid? shouldn't this break?

gnocchi/incoming/s3.py
```diff
@@ -27,11 +28,13 @@
...
 botocore = s3.botocore
 
+LOG = daiquiri.getLogger(__name__)
+
 
 class S3Storage(incoming.IncomingDriver):
 
     # NOTE(gordc): override to follow s3 partitioning logic
-    SACK_NAME_FORMAT = "{number}-{total}/"
+    SACK_NAME_FORMAT = "{number}-{total}"
```
chungg: should we have a releasenote that for sure s3 will lose unprocessed measures?

gnocchi/incoming/s3.py
```diff
@@ -175,3 +170,28 @@ def process_measure_for_metrics(self, metric_ids):
...
+                Key=f)
+            measures[metric_id] = numpy.concatenate((
+                measures[metric_id],
+                self._unserialize_measures(f, response['Body'].read())
+            ))
+
+        yield measures
+
+        # Now clean objects
+        s3.bulk_delete(self.s3, self._bucket_name_measures, files)
```
chungg: files may be a superset of actual processed measures... (assuming it fails to convert to UUID)

gnocchi/incoming/swift.py
```diff
@@ -117,3 +120,27 @@ def process_measure_for_metrics(self, metric_ids):
...
+                measures[metric_id],
+                self._unserialize_measures(
+                    metric_id,
+                    self.swift.get_object(sack_name, f['name'])[1],
+                )
+            ])
+
+        yield measures
+
+        swift.bulk_delete(self.swift, sack_name, files)
```
chungg: same comment as s3

gnocchi/incoming/ceph.py
jd: Yes, just in case. It should never be hit.

gnocchi/incoming/file.py
jd: No that's good, it calls str(metric_id). I admit those methods are a bit fuzzy now, I might rework/rename them in a next patch.

gnocchi/incoming/s3.py
jd: Upgrade should always be done with unprocessed measures empty, that's in the docs.

gnocchi/incoming/s3.py
jd: Is this a problem? I imagine it's better to clean the wrong file than converting a bad file over and over again.

chungg: i agree. we should do the same for ceph then, as it only deletes what it processes.

chungg: same for file?

jd: I'll do it for Ceph. For files, it's not really possible as it's a directory with wrong format not matching a metric, so it's different.

jd:
Delete unparsable filename in Ceph

PR #803, opened 2018-03-09
Make the build reproducible

lamby:
Whilst working on the Reproducible Builds effort [0], we noticed
that gnocchi could not be built reproducibly as it iterates over
a Python set when generating documentation.

This was originally filed in Debian as #892419.

 [0] https://reproducible-builds.org/
 [1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=892419

Signed-off-by: Chris Lamb <lamby@debian.org>


lamby:
Thanks!

PR #805, opened 2018-03-09
storage: more batching of store/delete operations in add_measures

jd:
Rather than emitting store and delete on each aggregation computing, this
batches them all at once for a metric by returning the operations from
`_add_measures` directly.

### DIFF
--- a/gnocchi/storage/__init__.py
+++ b/gnocchi/storage/__init__.py
@@ -354,9 +354,29 @@ class StorageDriver(object):
 
         return self._store_metric_splits(metric, keys_aggregations_data_offset)
 
-    def _add_measures(self, metric, aggregations, grouped_serie,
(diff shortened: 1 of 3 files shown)


gnocchi/storage/__init__.py
```diff
@@ -357,6 +357,25 @@ def _store_timeserie_splits(self, metric, keys_and_aggregations_and_splits,
     def _add_measures(self, metric, aggregations, grouped_serie,
```
chungg: is there a reason you didn't change this method name? too many places to change in test? just asking because this method name is misleading from what it actually does.

jd: No, I just did not think about changing it. Let me do that! Do you have a suggestion?

jd:
Change _add_measures name

PR #820, opened 2018-03-16
chef: lock sack when using process_new_measures_for_sack

jd:
This makes it safer to call it without taking care of the lock externally.

### DIFF
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -532,8 +532,9 @@ class MetricController(rest.RestController):
                 pecan.request.chef.refresh_metric(
                     self.metric,
                     pecan.request.conf.api.operation_timeout)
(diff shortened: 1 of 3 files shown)


gnocchi/rest/api.py
```diff
@@ -1904,7 +1905,9 @@ def get_cross_metric_measures_from_objs(metrics, start=None, stop=None,
                         pecan.request.chef.refresh_metric(
                             m, pecan.request.conf.api.operation_timeout)
                     except chef.SackLockTimeoutError as e:
```
chungg: need to update exception.

jd: fixed, thanks!

jd:
Fix API except on chef.SackAlreadyLocked

PR #822, opened 2018-03-19
storage: _delete_metric_splits can now batch over multiple metrics

jd:


gnocchi/storage/ceph.py
```diff
@@ -94,23 +94,27 @@ def _store_metric_splits(self, metric, keys_aggregations_data_offset,
...
+                )
+                for name in names:
+                    try:
+                        self.ioctx.remove_object(name)
+                    except rados.ObjectNotFound:
+                        # It's possible that we already remove that object and
+                        # then crashed before removing it from the OMAP key
+                        # list; then no big deal anyway.
+                        pass
+                self.ioctx.remove_omap_keys(op, names)
```
chungg: just remove all the keys at the end outside of loop.

jd: Can't, it's per metric

chungg: I don't understand... they're keys. how would they not be unique?

jd: They are stored on different objects, e.g. `self._build_unaggregated_timeserie_path(metric, 3)`.

chungg: hm got it

jd:
Fix Ceph signature

PR #825, opened 2018-03-20
storage: batch split storage on multiple metrics

jd:


gnocchi/storage/ceph.py
```diff
@@ -81,18 +81,21 @@ def _create_metric(self, metric):
...
-                op, self._build_unaggregated_timeserie_path(metric, 3))
+            for metric, keys_aggregations_data_offset in six.iteritems(
+                    metrics_keys_aggregations_data_offset):
+                for key, agg, data, offset in keys_aggregations_data_offset:
+                    name = self._get_object_name(
+                        metric, key, agg.method, version)
+                    if offset is None:
+                        self.ioctx.write_full(name, data)
+                    else:
+                        self.ioctx.write(name, data, offset=offset)
```
chungg: just thinking, but it seems like we're synchronously writing objects and then batching keys. maybe we should thread the object writes here.

jd: I don't think it's a good idea to introduce threading in drivers directly. Especially since here Ceph bindings might be using them internally.

chungg: it really seems like this will negatively affect things as object write is the slower operation and this was something we were running in parallel before.

that said, i have no ceph to test on.

PR #829, opened 2018-03-22
metricd: publish statistics about processing speed

### LINKED ISSUE #600: Export speed statistics of metricd cluster
It would be useful that each metricd service exports its average processing speed so the speed of the cluster can be determined and provisioning and scalability is easier.

jd:
Closes #600


jd:
Add missing import

PR #830, opened 2018-03-22
metricd: use process_new_measures_for_sack in test code

jd:


gnocchi/cli/metricd.py
```diff
@@ -297,14 +297,15 @@ def metricd_tester(conf):
...
     s = storage.get_driver(conf)
     inc = incoming.get_driver(conf)
-    metrics = set()
+    c = chef.Chef(None, inc, index, s)
+    metrics_count = 0
     for sack in inc.iter_sacks():
-        metrics.update(inc.list_metric_with_measures_to_process(sack))
-        if len(metrics) >= conf.stop_after_processing_metrics:
+        try:
+            metrics_count += c.process_new_measures_for_sack(s, True)
```
chungg: this doesn't return metric count. it returns measure count

jd: I beg to differ.

chungg: oh. https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/chef.py#L172 is named poorly.

PR #833, opened 2018-03-22
chef: merge refresh_metric and process_new_measures

jd:
This simplifies the code base by just allowing to process new measures sack per
sack with the existing process_new_measures_for_sack() or to use the new
refresh_metrics() who takes a list of metrics. Since the list of metric must be
known, there's no need to use list_metric_with_measures_to_process() anymore,
so it can be removed.

### DIFF
--- a/gnocchi/chef.py
+++ b/gnocchi/chef.py
@@ -102,32 +102,31 @@ class Chef(object):
                         LOG.error("Unable to expunge metric %s from storage",
                                   metric, exc_info=True)
 
-    def refresh_metric(self, metric, timeout):
-        s = self.incoming.sack_for_metric(metric.id)
-        lock = self.get_sack_lock(s)
-        if not lock.acquire(blocking=timeout):
-            raise SackAlreadyLocked(s)
-        try:
-            self.process_new_measures([str(metric.id)])
-        finally:
-            lock.release()
+    def refresh_metrics(self, metrics, timeout=None, sync=False):
+        """Process added measures in background for some metrics only.
 
-    def process_new_measures(self, metrics_to_process, sync=False):
-        """Process added measures in background.
-
-        Some drivers might need to have a background task running that process
-        the measures sent to metrics. This is used for that.
+        :param metrics: The list of `indexer.Metric` to refresh.
+        :param timeout: Time to wait for the process to happen.
+        :param sync: If an error occurs, raise, otherwise just log it.
         """
         # process only active metrics. deleted metrics with unprocessed
         # measures will be skipped until cleaned by janitor.
-        metrics = self.index.list_metrics(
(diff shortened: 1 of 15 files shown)


gnocchi/chef.py
```diff
@@ -102,57 +102,60 @@ def expunge_metrics(self, sync=False):
...
-            LOG.error("Error processing new measures", exc_info=True)
+            except Exception:
+                if sync:
+                    raise
+                LOG.error("Error processing new measures", exc_info=True)
+            finally:
+                lock.release()
 
-    def process_new_measures_for_sack(self, sack, sync=False):
+    def process_new_measures_for_sack(self, sack, blocking=False, sync=False):
```
chungg: what's the diff between sync and blocking?

jd: `sync` makes the function raise as soon as an error occurs. It's set to `True` in tests so we are sure not to just log the error, but fails the test. `block` is to make sure the sack is processed even if it's already locked when set to `True`, otherwise it'd raise `SackLockTimeout`.

jd:
Pull-request updated, HEAD is now 97c98ba213041faa34a184a1fcf6288f9c962fc4

gnocchi/chef.py
chungg: yeah, well i get that :p. was wondering if they were the same use case so we don't need two somewhat similar params. guess not.

jd: Yeah no :)

jd:
Pull-request updated, HEAD is now 189de01c3f087f4490a5010fbd1c648a11bf7b4b

jd:
Block refresh_metrics

gnocchi/tests/base.py
```diff
@@ -390,6 +390,7 @@ def _create_metric(self, archive_policy_name="low"):
     def trigger_processing(self, metrics=None):
         if metrics is None:
             self.chef.process_new_measures_for_sack(
-                self.incoming.sack_for_metric(self.metric.id), sync=True)
+                self.incoming.sack_for_metric(self.metric.id),
+                blocking=True, sync=True)
         else:
-            self.chef.process_new_measures(metrics, sync=True)
+            self.chef.refresh_metrics(metrics, timeout=True, sync=True)
```
chungg: do we have a use case for refreshing multiple metrics at a time? if not, i'd prefer to have this loop and call refresh_metrics rather than modify existing refresh_metrics above just for tests.

jd: Other than this, I don't think we have a use case. Though refresh_metrics is not really modified, it's just renamed/merged with process_new_measures.
I can envision a refresh_metrics() for passing `?refresh=true` to the API when doing e.g. dynamic .aggregates

PR #837, opened 2018-03-27
api: catch UnexpectedResourceTypeState properly

jd:
When creating resource, this might happen if the resource type is not
completely created yet (race condition). Rather than crash and return 500,
return a 503 with a parseable reply so it can be retried.

### DIFF
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -1106,6 +1106,8 @@ class ResourcesController(rest.RestController):
             abort(400, six.text_type(e))
         except indexer.ResourceAlreadyExists as e:
             abort(409, six.text_type(e))
+        except indexer.UnexpectedResourceTypeState as e:
+            abort(500, e)
         set_resp_location_hdr("/resource/"
                               + self._resource_type + "/"
                               + six.text_type(resource.id))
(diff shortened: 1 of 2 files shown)


gnocchi/rest/api.py
```diff
@@ -1135,6 +1135,8 @@ def post(self):
             abort(400, six.text_type(e))
         except indexer.ResourceAlreadyExists as e:
             abort(409, six.text_type(e))
+        except indexer.UnexpectedResourceTypeState as e:
```
sileht: In case of e.state is "creating" we can retry create resource or maybe raise 409 also

jd: Do we want to do that server side? I'm not sure. Or maybe I'm lazy.

chungg: i'm hesitant about 503. when i see 503, it's generally that service is completely offline/unaccessible.

jd: Right. Maybe 500 would be better.

jd:
Use 500 instead of 503

gnocchi/rest/api.py
chungg: i think i agree with sileht, 5xx anything is very aggressive to me.

sileht: I don't see the point of catching a 500 for an other 500.

jd: This one has a proper exception info returned?

PR #840, opened 2018-03-28
tests: replace refresh_metrics() with trigger_processing in statsd

jd:
The current tests can fail just because the timeout it not set to None. The
base method trigger_processing is meant to be used just for that!

### DIFF
--- a/gnocchi/tests/test_statsd.py
+++ b/gnocchi/tests/test_statsd.py
@@ -73,7 +73,7 @@ class TestStatsd(tests_base.TestCase):
 
         metric = r.get_metric(metric_key)
 
-        self.chef.refresh_metrics([metric], sync=True)
+        self.trigger_processing([metric])
 
         measures = self.storage.get_measures(metric, self.aggregations)
         self.assertEqual({"mean": [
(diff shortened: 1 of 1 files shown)


jd:
Pull-request updated, HEAD is now cda6ccfbd85767a3a743e165d9500730d949c343

gnocchi/tests/functional/gabbits/aggregates-with-resources.yaml
```diff
@@ -127,7 +127,7 @@ tests:
         search: "user_id = 'A50F549C-1F1C-4888-A71A-2C5473CCCEC1'"
         operations: "(aggregate mean (metric cpu.util mean))"
       poll:
-        count: 10
+        count: 60
```
chungg: it seems like this is just hiding potentially another issue. pifpaf is polling for work every second so even 10s should be more than enough.

jd: Yeah I was just trying to see if that was the source of the problem, but clearly it's not. Not sure what it is :(

jd:
Pull-request updated, HEAD is now fb37011f933c0bee9b676569325fe91a0f358579

jd:
@chungg yup, done

PR #846, opened 2018-04-02
don't do work if no measures

chungg:
bail early if we know there's no work to be done.

### DIFF
--- a/gnocchi/chef.py
+++ b/gnocchi/chef.py
@@ -127,13 +127,14 @@ class Chef(object):
                 LOG.debug("Processing measures for %d metrics", len(metrics))
                 with self.incoming.process_measure_for_metrics(
                         metrics) as metrics_and_measures:
-                    self.storage.add_measures_to_metrics({
-                        metrics_by_id[metric]: measures
-                        for metric, measures
-                        in six.iteritems(metrics_and_measures)
-                    })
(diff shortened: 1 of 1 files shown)


chungg:
it's pretty weird that py3 jobs are consistently noticeably slower than py2... :/

gnocchi/chef.py
```diff
@@ -127,13 +127,14 @@ def refresh_metrics(self, metrics, timeout=None, sync=False):
...
                 with self.incoming.process_measure_for_metrics(
                         metrics) as metrics_and_measures:
-                    self.storage.add_measures_to_metrics({
-                        metrics_by_id[metric]: measures
-                        for metric, measures
-                        in six.iteritems(metrics_and_measures)
-                    })
-                    LOG.debug("Measures for %d metrics processed",
-                              len(metrics))
+                    if metrics_and_measures:
```
jd: Is this really worth it? Most of the code is noop then. This sounds like a micro optimization.
Just asking.

chungg: it seems safer rather than assuming all the drivers are smart enough underneath.

for instance, iiuc, i'm pretty sure the redis driver will make a request even if this is empty

below, the list_metrics command i believe will also make a request.

PR #854, opened 2018-04-07
block on refresh_metrics for tests

chungg:
block so we guarantee required data is processed and we don't raise
an error because of other workers.


chungg:
for about #840

PR #892, opened 2018-05-20
support resampling on calendar dates

chungg:
support ability to group by year, half, quarter, month, and week.

### DIFF
--- a/doc/source/rest.j2
+++ b/doc/source/rest.j2
@@ -187,6 +187,18 @@ can be resampled to a new |granularity|.
 
 {{ scenarios['get-measures-resample']['doc'] }}
 
+Time-series data can also be grouped by calendar dates beyond a standard day.
+The resulting groupings are tied to the leading date of the group. For example,
+grouping on month returns a monthly aggregate linked to the first of the month.
+Available calendar groups are::
+
+   `Y` - by year.
+   `H` - by half.
+   `Q` - by quarter.
+   `M` - by month.
+   `W` - by week, starting on Sunday.
+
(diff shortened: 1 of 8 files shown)


jd:
That does not seem to be plugged to anything yet?
Also wondering if there's no way to use something that already does all of that but I guess you checked already.

chungg:
yeah, i need to add this to `GroupedTimeSerie`... i actually haven't looked at how it will work as some of them i imagine won't work for nested resample using aggregation endpoint (ie. the ones that are not bound to a year)

i didn't actually look at other projects. Pandas has most (if not all) but i don't feel like bringing in Pandas. the values can also be computed via datetime but it's 2x-4x slower (but we're only talking 200us vs 400us) so yeah, if you have a solution i'm open to it. i think regardless, it will need to be connected to our `GroupedTimeSerie` class.

chungg:
hmm... yeah i just took a quick look at how to connect calendar to `GroupedTimeSerie` and our api and i'm not entirely sure how to implement the calendar groupings that aren't related to a date (ie. day of the year, day of the week, week of the year, etc...).

our api and data structure revolves around datetime so in the above cases, we will probably need quite a few conditionals to specially handle those... or we create a new endpoint and grouping model to group and return number+value vs time+value.

chungg:
note: this doesn't work currently for groups unrelated to timestamp (ie day of week, month, day of year... )

chungg:
ok. just fyi, i will probably disable the those non-time related groupings until i can figure out a better way. (and i might rename the resample names)

Issue #509, opened 2017-11-22
Swift connection pool size is too small


jd:
When Swift is used as an incoming measures driver, the number of connections opened by httplib3 is too small to succeed. On large batches, the error occurs:

```
[Wed Nov 22 16:17:50.622612 2017] [:error] [pid 31427] 2017-11-22 16:17:50,622 [31427] WARNING  urllib3.connectionpool: Connection pool is full, discarding connection: 10.41.11.12
[Wed Nov 22 16:17:50.625786 2017] [:error] [pid 31427] [remote ::1:80] mod_wsgi (pid=31427): Exception occurred processing WSGI script '/bin/gnocchi-api'.
[Wed Nov 22 16:17:50.625846 2017] [:error] [pid 31427] [remote ::1:80] Traceback (most recent call last):
[Wed Nov 22 16:17:50.625898 2017] [:error] [pid 31427] [remote ::1:80]   File "/usr/lib/python2.7/site-packages/webob/dec.py", line 131, in __call__
[Wed Nov 22 16:17:50.625950 2017] [:error] [pid 31427] [remote ::1:80]     resp = self.call_func(req, *args, **self.kwargs)
[Wed Nov 22 16:17:50.625961 2017] [:error] [pid 31427] [remote ::1:80]   File "/usr/lib/python2.7/site-packages/webob/dec.py", line 196, in call_func
[Wed Nov 22 16:17:50.625975 2017] [:error] [pid 31427] [remote ::1:80]     return self.func(req, *args, **kwargs)
[Wed Nov 22 16:17:50.625991 2017] [:error] [pid 31427] [remote ::1:80]   File "/usr/lib/python2.7/site-packages/oslo_middleware/base.py", line 131, in __call__
[Wed Nov 22 16:17:50.626008 2017] [:error] [pid 31427] [remote ::1:80]     response = req.get_response(self.application)
[Wed Nov 22 16:17:50.626016 2017] [:error] [pid 31427] [remote ::1:80]   File "/usr/lib/python2.7/site-packages/webob/request.py", line 1327, in send
[Wed Nov 22 16:17:50.626031 2017] [:error] [pid 31427] [remote ::1:80]     application, catch_exc_info=False)
[Wed Nov 22 16:17:50.626039 2017] [:error] [pid 31427] [remote ::1:80]   File "/usr/lib/python2.7/site-packages/webob/request.py", line 1291, in call_application
[...]

PR #897, opened 2018-05-22
swift: avoid Connection aborted

sileht:
This change removes usage of threads with swift driver.

This avoids to get "Connection aborted" because a thread is stuck
and the server side decide to break the connection.

Related-bug: #509


sileht:
Pull-request updated, HEAD is now 292c6059e92b7e0be28a55ab652cb430467099e9

PR #899, opened 2018-05-24
storage: allow to batch splits retrieval for multiple metrics

jd:
It also fixes the parallel_map call to be unique, making _get_splits a little
bit faster for unbatched drivers.

### DIFF
--- a/gnocchi/storage/redis.py
+++ b/gnocchi/storage/redis.py
@@ -13,8 +13,6 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-import collections
-
 import six
 
 from gnocchi import carbonara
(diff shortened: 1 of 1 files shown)


jd:
Pull-request updated, HEAD is now 3808c50deed1f1efa266f0574a3de0c701a0981f

chungg:
i should've asked this earlier but did we ever do the math of how much memory it takes to defer writing to a single batch? i imagine we are storing hundreds of metrics + 12 series each per bucket?

jd:
@chungg your question is not related to this patch that is about reading, right?

For memory on writing, it takes more memory, but the problem is not different that it was before, it's just potentially amplified by the number of metrics.

In Gnocchi <= 4.2 you can already send 1 billion of data point for just 1 second of timespan (nanosecond precision, remember). That's not problably gonna work.

jd:
Wrong PR.

chungg:
word. well i guess we can consider batch sizes but we can look at that later if needed.

PR #900, opened 2018-05-29
file: allow to create a subdir to create less root directory

sileht:
Some filesystem have a limit in term of number of directory a directory
can have. Also in performance point of view, some filesystem are slower
when too much directories are created in one directory.

To improve the situation is such case, this change adds an option
to create subdirectories for storing metrics.

for example file_subdir_len=2 will create:

de/de3a72c4-5055-43b4-bed1-f9bb1395c52a

### DIFF
--- a/gnocchi/storage/file.py
+++ b/gnocchi/storage/file.py
@@ -17,11 +17,14 @@
 import collections
 import errno
 import itertools
+import json
 import operator
 import os
 import shutil
 import tempfile
+import uuid
 
+import daiquiri
 from oslo_config import cfg
 import six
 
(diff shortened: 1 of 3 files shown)


gnocchi/storage/file.py
```diff
@@ -34,10 +37,16 @@
     cfg.StrOpt('file_basepath',
                default='/var/lib/gnocchi',
                help='Path used to store gnocchi data files.'),
+    cfg.IntOpt('file_subdir_len',
```
jd: I'm not sure that solves the problem you had in mind entirely. If a filesystem is limited to 65535 items per directory, then splitting on anything more than the 2 first bytes (`2^16`) could lead to a problem. So you have to split on those first 2 bytes of the UUID to be sure your root directory does not exceed 65k items.
Then, in each subdirectory, you'll get the same problem.

Sounds like to we should create a directory every 2 bytes. A UUID being 32 chars long, that means having a structure of 16 directories long for each metric.

wdyt?

sileht: yeah, this is what I was about to do, should we just hardcode 2 and that is ? Or should I keep the configuration option and put 2 by default ?

sileht:
Pull-request updated, HEAD is now 2c11f9531fd6fe440d03bbe7c9d98fd48186365c

gnocchi/storage/file.py
jd: I think you can keep it and set 2 by default.

sileht:
Pull-request updated, HEAD is now 63d159b73f084d540da0fd12aed811be456da52c

sileht:
Pull-request updated, HEAD is now 806f32621dacca33f5a77014940c3d101212634b

PR #903, opened 2018-06-01
storage: allow to list split keys for different metric at the same time

jd:


gnocchi/storage/redis.py
```diff
@@ -97,35 +97,44 @@ def _get_or_create_unaggregated_timeseries(self, metrics, version=3):
...
-                in six.moves.zip(timestamps, granularities)
-            }
+        keys = collections.defaultdict(dict)
+        for metric in metrics:
+            metric_exists_p = results.pop(0)
+            if not metric_exists_p:
+                raise storage.MetricDoesNotExist(metric)
+            aggregations = metrics_and_aggregations[metric]
+            keys_for_aggregations = results[:len(aggregations)]
+            del results[:len(aggregations)]
```
chungg: can we just save (and offset) the start rather than having to del from the list.

gnocchi/storage/redis.py
```diff
@@ -97,35 +97,44 @@ def _get_or_create_unaggregated_timeseries(self, metrics, version=3):
...
+            for aggregation, k in six.moves.zip(
+                    aggregations, keys_for_aggregations):
+                if not k:
+                    keys[metric][aggregation] = set()
+                    continue
+                timestamps, methods, granularities = list(zip(*k))
```
chungg: what's `methods`?

can't we expand this when we define for loop? like 
`for aggregation, (timestamps, granularities) in six.moves.zip(aggregations, keys_for_aggregations):`

gnocchi/storage/redis.py
jd: Done

gnocchi/storage/redis.py
jd: As a note, this is just how the code was written previously, not new code.

> what's `methods`?

The aggregation method as a string parsed out from the Redis key name by the Lua script.

> can't we expand this when we define for loop? like
for aggregation, (timestamps, granularities) in six.moves.zip(aggregations, keys_for_aggregations)

No, I think you misread the code.

jd:
Changed `del` to offset

gnocchi/storage/redis.py
chungg: word. i was reviewing in unified view and didn't notice the original.

PR #904, opened 2018-06-04
Avoid using distutils at runtime

thomasgoirand:
Please consider merging this patch, which is removing distutils from
runtime use. Using distutils at runtime is annoying for distributions,
and both Debian and Ubuntu are using this patch.

### DIFF
--- a/gnocchi/utils.py
+++ b/gnocchi/utils.py
@@ -15,7 +15,6 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 import datetime
-import distutils.util
 import errno
 import itertools
 import multiprocessing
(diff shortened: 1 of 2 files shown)


gnocchi/utils.py
```diff
@@ -203,10 +202,17 @@ def ensure_paths(paths):
...
-    if isinstance(v, bool):
-        return v
-    return bool(distutils.util.strtobool(v))
+def strtobool(val):
+    if isinstance(val, bool):
+        return val
+    # copied from distutils.util ...
+    val = val.lower()
+    if val in ('y', 'yes', 't', 'true', 'on', '1'):
+        return 1
```
chungg: this should return a boolean, not a boolean-like a value.

jd:
Would you mind saying "why" it's a problem?

PR #905, opened 2018-06-04
redis: use hiredis parser

chungg:
redis-py has a python and c parser[1]. install the c parser. it offers
~10%-90% performance improvement depending on task using benchmark from
redis-py[2].

[1] https://github.com/andymccurdy/redis-py#parsers
[2] https://github.com/andymccurdy/redis-py/blob/master/benchmarks/basic_operations.py


chungg:
i'm indifferent. i think by default it looks for hiredis but i agree. it's not magically better. the benchmark also can be done against pipelines which might be more realistic to our work flow.

chungg:
lol i lack conviction.

PR #925, opened 2018-06-27
Delete stale metrics before policy delete

i-newton:
Usually stale metrics are soft deleted. Later they cleanud up by
metrics daemon. This leads to errors if  smb tries to delete
policy right after metric delete because there is foreign key
restriction between metrics and policies.

Change-Id: I9dd02b7c54b2a8086cd8cf57ff4c20bca67c42ee


i-newton:
I have the following issue:
I can't delete archive policy right after metric delete, because there is foreign key restriction between metrics and archive policies, so i need to wait some time until metricsd expunge all the data,
is there any way(except proposed in PR) to allow delete archive policy after metric delete?

jd:
Not right now. What could be done is to delete the archive policy asynchronously by using a `status` field like it's done with metrics.

PR #931, opened 2018-07-13
Pull request for rework-find-measures

jd:
storage: move MeasureQuery in REST API
storage: remove find_measures
storage: make find_measures batched


gnocchi/rest/api.py
```diff
@@ -1484,35 +1579,49 @@ def post(self, metric_id, start=None, stop=None, aggregation='mean',
...
                 reverse=True)
 
-        results = {}
+        metrics_and_aggregations = collections.defaultdict(list)
+
+        for metric in metrics:
+            if granularity is None:
+                granularity = sorted((
+                    d.granularity
+                    for d in metric.archive_policy.definition),
```
chungg: i imagine we can just check aggregations here while we're looping through the list.

jd: I don't understand what you want to check here?

jd: Yeah I don't think it matters at that point :)

jd:
Pull-request updated, HEAD is now 10e6237700aa45114620248192907446026efcd4

PR #934, opened 2018-07-16
redis: fix list_split_keys with multiple metrics

jd:
The iteration over the results was wrongly computed and did not return data for
the second metric and following.


jd:
Fix pep8

PR #935, opened 2018-07-16
Test against python3.7

sileht:


sileht:
I tested it locally with just mysql-file, and tests passed, I was lazy to test all backends.

PR #939, opened 2018-07-24
Force installation of pbr<4

jd:
For whatever reason, tox/pip do not honor the dependency during the
installation, breaking the WSGI script gnocchi-api.


jd:
Not the right fix apparently.

PR #946, opened 2018-07-25
Stop using deprecated storage.get_measures

vinaykns:
This stops using deprecated storage.get_measures and
instead uses storge.get_aggregated_measures

### DIFF
--- a/gnocchi/storage/__init__.py
+++ b/gnocchi/storage/__init__.py
@@ -281,7 +281,8 @@ class StorageDriver(object):
         return name.split("_")[-1] == 'v%s' % v
 
     def get_aggregated_measures(self, metrics_and_aggregations,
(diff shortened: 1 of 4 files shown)


gnocchi/storage/__init__.py
```diff
@@ -276,8 +276,9 @@ def _version_check(name, v):
         """
         return name.split("_")[-1] == 'v%s' % v
 
-    def get_aggregated_measures(self, metrics_and_aggregations,
-                                from_timestamp=None, to_timestamp=None):
+    def get_aggregated_measures(self, metric, aggregations,
```
sileht: Can we keep the previous signature that allows to pass multiple metrics too.

jd: Yes this is meant to be batched :)

PR #957, opened 2018-08-03
storage: Remove useless fetch() in get_measures()

sileht:


sileht:
@vinaykns looks like the fetch is useless.

vinaykns:
@sileht  Yep.! Agreed.

Issue #959, opened 2018-08-17
Metrics with multiple archive policy granularities cause TypeErrors


openvdro:
Before reporting an issue on Gnocchi, please be sure to provide all necessary
information.

### Which version of Gnocchi are you using
using it on devstack, commit 8099dfc2a30ddf305d9f904de0106e4dd5e56147

### How to reproduce your problem
Create an archive policy with multiple granularities and use it for example to send nova server measures.

### What is the result that you get
The high archive policy created on devstack by default has granularities of 1 second, 1 hour and 1 day. After an hour, the following starts appearing in the logs:

```
Aug 16 17:00:04 localhost gnocchi-metricd[1997677]: 2018-08-16 17:00:04,312 [1997705] ERROR    gnocchi.chef: Error processing new measures
Aug 16 17:00:04 localhost gnocchi-metricd[1997677]: Traceback (most recent call last):
Aug 16 17:00:04 localhost gnocchi-metricd[1997677]:   File "/opt/stack/devstack/src/gnocchi/gnocchi/chef.py", line 165, in process_new_measures_for_sack
Aug 16 17:00:04 localhost gnocchi-metricd[1997677]:     for metric in metrics
Aug 16 17:00:04 localhost gnocchi-metricd[1997677]:   File "/opt/stack/devstack/src/gnocchi/gnocchi/storage/__init__.py", line 703, in add_measures_to_metrics
Aug 16 17:00:04 localhost gnocchi-metricd[1997677]:     before_truncate_callback=_map_compute_splits_operations,
Aug 16 17:00:04 localhost gnocchi-metricd[1997677]:   File "/opt/stack/devstack/src/gnocchi/gnocchi/carbonara.py", line 364, in set_values
Aug 16 17:00:04 localhost gnocchi-metricd[1997677]:     return_value = before_truncate_callback(self)
Aug 16 17:00:04 localhost gnocchi-metricd[1997677]:   File "/opt/stack/devstack/src/gnocchi/gnocchi/storage/__init__.py", line 691, in _map_compute_splits_operations
Aug 16 17:00:04 localhost gnocchi-metricd[1997677]:     new_first_block_timestamp)
Aug 16 17:00:04 localhost gnocchi-metricd[1997677]:   File "/opt/stack/devstack/src/gnocchi/gnocchi/storage/__init__.py", line 555, in _compute_split_operations
[...]

PR #960, opened 2018-08-17
Fix computation for metrics with multiple granularities

openvdro:
The high archive policy created by default contains three different
granularity definitions - 1 second, 1 hour and one day. For such
metrics the split computation has been done incorrectly as it was
using oldest keys not taking into account those keys granularity.
It caused TypeErrors when comparing two SplitKeys with different
granularities.

closes: gnocchixyz/gnocchi#959

### DIFF
--- a/gnocchi/storage/__init__.py
+++ b/gnocchi/storage/__init__.py
@@ -488,16 +488,17 @@ class StorageDriver(object):
         )
 
         aggregations_needing_list_of_keys = set()
+        oldest_values = {}
 
         for aggregation, ts in six.iteritems(aggregations_and_timeseries):
             # Don't do anything if the timeseries is empty
             if not ts:
                 continue
 
-            if aggregation.timespan:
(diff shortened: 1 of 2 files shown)


openvdro:
Hey Julien, not very familiar with the code, but I'll try to add something. Current test failures seem to happen because of tenacity though.

jd:
tenacity should be ok now.

chungg:
this probably needs to be backported as well.

openvdro:
yup it also affects 4.3 branch

jd:
Thanks a lot @openvdro !

openvdro:
Hmm, no idea why some tests failed for s3 driver only on postgres

openvdro:
Looking at the failed tests output it almost seems like this call https://github.com/openvdro/gnocchi/blob/master/gnocchi/tests/functional/gabbits/aggregates-with-resources.yaml#L71-L78 did not happen

PR #989, opened 2018-10-04
Update to Mergify v2 engine

jd:


jd:
v1 is now deprecated, can we merge this?

PR #994, opened 2018-10-05
Switch from oslo.middleware to internal proxy converter

jd:


jd:
Pull-request updated, HEAD is now 7204dbd7ff53816ff6a7b6a3602b7a254ed098e2

PR #1001, opened 2018-10-23
Support numpy clip

vinaykns:


gnocchi/rest/aggregates/operations.py
```diff
@@ -150,6 +154,47 @@ def handle_binary_operator(nodes, granularity, timestamps,
     return granularity, timestamps, values, is_aggregated
 
 
+def handle_ternary_operator(nodes, granularity, timestamps,
+                            initial_values, is_aggregated, references):
+    op = nodes[0]
+    g1, t1, v1, is_a1 = evaluate(nodes[1], granularity, timestamps,
+                                 initial_values, is_aggregated, references)
+    g2, t2, v2, is_a2 = evaluate(nodes[2], granularity, timestamps,
```
chungg: i think this will fail if you pass `None` for the min? can you add a test to verify?

gnocchi/rest/aggregates/operations.py
```diff
@@ -150,6 +154,47 @@ def handle_binary_operator(nodes, granularity, timestamps,
...
+
+    elif isinstance(v2, numpy.ndarray):
+        timestamps = t2
+        granularity = g2
+    elif isinstance(v3, numpy.ndarray):
+        timestamps = t3
+        granularity = g3
+    else:
+        timestamps = t1
+        granularity = g1
```
chungg: sort of torn regarding the code from lines 170 to 192... None of it actually is applicable to numpy.clip and i'm not sure what other operations would actually require it. thoughts?

specifically, there is function specific conditions. for example, if nodes[2] or nodes[3] aren't the same length as nodes[1], this will throw an error.

gnocchi/rest/aggregates/operations.py
vinaykns: Yes, currently if we pass None to min and a number to max the operation is getting invalid.

gnocchi/rest/aggregates/operations.py
vinaykns: I agree that nodes[2&3] aren't the same length as nodes[1] then why do we do the same sort of verification in handle_binary_operator as nodes[1] and nodes[2] don't have same length as well.

chungg: yeah, you're right. the above code does check the shape.

i guess the only part that is irrelevant is the granularity check `g1 != g2` and  `g2 != g3`. granularity should not really matter but i guess this code is still relevant.

gnocchi/tests/functional/gabbits/aggregation.yaml
```diff
@@ -361,3 +361,13 @@ tests:
     - name: get measure aggregates with no resource name
       POST: /v1/aggregation/resource/generic/metric
       status: 405
+
+    - name: get measure aggregates with None as min in clip operation
+      POST: /v1/aggregates?fill=0&granularity=1
+      data:
+        resource_type: generic
+        search: {}
+        operations: '(clip (metric mean) None 10)'
```
chungg: sorry, i meant we should support this case ideally,  not make it fail.

vinaykns: I guess the reason it is failing because, we are forcing the value to be float.
https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/rest/aggregates/api.py#L111

chungg: sorry for delay. yes, you're right. i imagine you'd need to look at adding something like: https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/rest/aggregates/api.py#L73-L82

i can try to take a look at this over weekend if anything.

sileht:
Can you squash all the commits ?

Issue #1013, opened 2019-01-17
gnocchi aggregates returns 400 while grouping by id


DilipRenkila:
I am running gnocchi version 4.2.5. I expected gnocchi to return aggregates while using groupby id option but it didn't seem to work in my case. It returns no metrics found 400 error. But i am able to get aggregates without groupby id . I am little bit confused to understand why it works like that. Can you guys explain me why it says no metrics ?

Without groupby id
```
root@ctrl1:~# gnocchi aggregates --resource-type volume --start "2019-01-01T00:00:00" --stop "2019-01-17T01:00:00" --fill 0 --granularity 300 "(metric volume.size max)" "project_id=2eea218eea984dd68f1378ea21c64b83"
+------------------------------------------------------+---------------------------+-------------+-------+
| name                                                 | timestamp                 | granularity | value |
+------------------------------------------------------+---------------------------+-------------+-------+
| ec1261cf-c372-4129-82a2-e299dd5ea202/volume.size/max | 2019-01-16T11:10:00+00:00 |       300.0 |   1.0 |
| ec1261cf-c372-4129-82a2-e299dd5ea202/volume.size/max | 2019-01-16T11:15:00+00:00 |       300.0 |   1.0 |
| ec1261cf-c372-4129-82a2-e299dd5ea202/volume.size/max | 2019-01-16T11:20:00+00:00 |       300.0 |   1.0 |
| ec1261cf-c372-4129-82a2-e299dd5ea202/volume.size/max | 2019-01-16T11:25:00+00:00 |       300.0 |   1.0 |
| ec1261cf-c372-4129-82a2-e299dd5ea202/volume.size/max | 2019-01-16T11:30:00+00:00 |       300.0 |   1.0 |
| ec1261cf-c372-4129-82a2-e299dd5ea202/volume.size/max | 2019-01-16T11:35:00+00:00 |       300.0 |   1.0 |
| ec1261cf-c372-4129-82a2-e299dd5ea202/volume.size/max | 2019-01-16T11:40:00+00:00 |       300.0 |   1.0 |
| ec1261cf-c372-4129-82a2-e299dd5ea202/volume.size/max | 2019-01-16T11:45:00+00:00 |       300.0 |   1.0 |
| ec1261cf-c372-4129-82a2-e299dd5ea202/volume.size/max | 2019-01-16T11:50:00+00:00 |       300.0 |   1.0 |
[...]

PR #1015, opened 2019-01-30
don't fail everything if one group missing metric

chungg:
it's possible that not all groups will contain a specific metric.
rather than fail, return the groups that do have the matching metric.
only if all groups don't have matching metrics, should we fail.

Fixes: #1013

### DIFF
--- a/gnocchi/rest/aggregates/api.py
+++ b/gnocchi/rest/aggregates/api.py
@@ -246,21 +246,32 @@ class AggregatesController(rest.RestController):
             except indexer.IndexerException as e:
                 api.abort(400, six.text_type(e))
             if not groupby:
-                return self._get_measures_by_name(
-                    resources, references, body["operations"], start, stop,
-                    granularity, needed_overlap, fill, details=details)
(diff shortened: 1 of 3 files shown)


gnocchi/rest/aggregates/api.py
```diff
@@ -246,21 +246,32 @@ def post(self, start=None, stop=None, granularity=None,
...
                 api.abort(400, six.text_type(e))
             if not groupby:
-                return self._get_measures_by_name(
-                    resources, references, body["operations"], start, stop,
-                    granularity, needed_overlap, fill, details=details)
+                try:
+                    return self._get_measures_by_name(
+                        resources, references, body["operations"], start, stop,
+                        granularity, needed_overlap, fill, details=details)
+                except indexer.NoSuchMetric as e:
```
jd: Would you mind rather adding a `jsonify` method to the `indexer.NoSuchMetric` exception?

chungg: sure thing.

PR #1041, opened 2019-07-31
try forcing newer version of six on to travis image

chungg:
it's not installing correct version when it installs twine during
deploy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,10 @@ notifications:
     channels:
       - "irc.freenode.org#gnocchi"
 
+before_deploy:
+  - pip install --user --upgrade pip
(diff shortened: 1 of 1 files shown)


chungg:
the weird thing is, the upgrade actually exists from what i can tell https://github.com/travis-ci/dpl/blob/6405ca589804d9dee961b207de31983c07fcb661/spec/provider/pypi_spec.rb#L12

i have a guess that it's either because we have `sudo: required` in our file or because we use `language: generic` and the travis docs mention including `--user` in your commands which the deploy code above does not. that said, i'm very much guessing like i did this PR.lol i'll create an issue on dpl repo to see if they know.

PR #1054, opened 2020-01-16
Use the current website

mscherer:
Since the domain expired, it would be better to at least show the doc using the VM name.


jd:
#1053 might be more complete :)

mscherer:
Yup

PR #1056, opened 2020-02-07
Enhance rest api logging

pedro-martins:
To help operators to debug some API errors, was added some logs in warning and debug level to log the responses in Gnocchi rest API.

### DIFF
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -17,6 +17,7 @@
 import collections
 import functools
 import itertools
+import logging
 import operator
 import uuid
 
@@ -42,6 +43,7 @@ from gnocchi import json
 from gnocchi import resource_type
 from gnocchi.rest.aggregates import exceptions
 from gnocchi.rest.aggregates import processor
+from gnocchi.rest import exceptions as rest_exceptions
 from gnocchi import storage
 from gnocchi import utils
 
@@ -54,6 +56,7 @@ except ImportError:
 
 
 ATTRGETTER_GRANULARITY = operator.attrgetter("granularity")
+LOG = logging.getLogger(__name__)
 
 
 def arg_to_list(value):
@@ -77,6 +80,7 @@ def abort(status_code, detail=''):
         }
     elif isinstance(detail, Exception):
         detail = detail.jsonify()
+    LOG.debug("Aborting request. Code [%s]. Details [%s]", status_code, detail)
     return pecan.abort(status_code, detail)
 
 
@@ -157,7 +161,10 @@ def deserialize(expected_content_types=None):
     try:
         params = json.load(pecan.request.body_file)
     except Exception as e:
-        abort(400, "Unable to decode body: " + six.text_type(e))
+        details = rest_exceptions.UnableToDecodeBody(e,
+                                                     pecan.request.body_file)
+        LOG.warning(details.jsonify())
+        abort(400, details)
     return params
 
 
--- a/gnocchi/tests/functional/gabbits/resource.yaml
+++ b/gnocchi/tests/functional/gabbits/resource.yaml
@@ -312,10 +312,15 @@ tests:
 
     - name: patch resource no data
       desc: providing no data is an error
+      request_headers:
+        accept: application/json
       PATCH: /v1/resource/generic/75C44741-CC60-4033-804E-2D3098C7D2E9
       status: 400
-      response_strings:
-          - "Unable to decode body:"
+      response_json_paths:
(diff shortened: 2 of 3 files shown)


gnocchi/rest/api.py
```diff
@@ -157,7 +159,10 @@ def deserialize(expected_content_types=None):
     try:
         params = json.load(pecan.request.body_file)
     except Exception as e:
-        abort(400, "Unable to decode body: " + six.text_type(e))
+        details = "Unable to decode body: %s \n" \
```
chungg: do you actually want that new line in the message? is there a common practice on how to escape them because i find multi-line logs to be unusable as when you filter, most systems end up returning only the matching line and you have an incomplete log.

gnocchi/rest/api.py
```diff
@@ -157,7 +159,10 @@ def deserialize(expected_content_types=None):
...
+                  "Body: %s" % (six.text_type(e), pecan.request.body_file)
+        LOG.warning(details)
```
chungg: just curious, is there a reason for double logging at different levels?

gnocchi/rest/api.py
rafaelweingartner: I also did not like that multi-line logging. Maybe enclosing the values in "[]" in the log string can help. Such as `Unable to decode body: [%s], request body [%s]`.

gnocchi/rest/api.py
rafaelweingartner: I talked with Pedro about that. The log in debug level of the `abort` command will affect other errors as well. However, this specific one, in our opinion, should generate a warning message as well. So, when it happens, the operators can see the error and the request body that generated it.

gnocchi/rest/api.py
chungg: that could work, i'd think log filtering tools should be able to parse that.

we did start transitioning our errors to be json (search on jsonify). this was mainly for api responses but it could also work for logs as well i believe? maybe @jd has an opinion on it but as long as there's no newline, it works for me :)

chungg:
no idea why i can't force rebuild in travisci so i'm closing and reopening

gnocchi/rest/api.py
```diff
@@ -157,7 +160,9 @@ def deserialize(expected_content_types=None):
     try:
         params = json.load(pecan.request.body_file)
     except Exception as e:
-        abort(400, "Unable to decode body: " + six.text_type(e))
+        details = UnableToDecodeBody(e, pecan.request.body_file)
+        LOG.warning(details.jsonify())
+        abort(400, details)
```
chungg: you'll need to update the testcase attached to this error.

gnocchi/rest/api.py
```diff
@@ -235,6 +240,22 @@ def get_pagination_options(params, default):
...
+        super(UnableToDecodeBody, self).__init__(body)
+
+    def jsonify(self):
+        return {
+            "cause": "Unable to decode body",
+            "detail": {
+                "text_type": self.text_type,
+                "body": self.body
+            }
+        }
```
chungg: i would change this to:

```
class UnableToDecodeBody(Exception):
    def __init__(self, exception, body):
        self.reason = six.text_type(exception)
        self.body = body
        super(UnableToDecodeBody, self).__init__("Unable to decode body")

    def jsonify(self):
        return {
            "cause": "Unable to decode body",
            "reason": self.reason,
            "detail": self.body
        }
```
and temporarily put it along side `gnocchi/rest/aggregates/exceptions.py` (or move it to `gnocchi/rest/exceptions.py`)

gnocchi/tests/functional/gabbits/resource.yaml
```diff
@@ -312,10 +312,15 @@ tests:
...
+        accept: application/json
       PATCH: /v1/resource/generic/75C44741-CC60-4033-804E-2D3098C7D2E9
       status: 400
-      response_strings:
-          - "Unable to decode body:"
+      response_json_paths:
+        $.description.cause: "Unable to decode body"
+        $.description.detail:
+          body: []
+          text_type: "Expected object or value"
```
chungg: $.description.cause: "Unable to decode body"
        $.description.reason: "Expected object or value"

gnocchi/rest/api.py
pedro-martins: Ok, I extracted this class to a new file 'gnocchi/rest/exceptions.py'

chungg:
fyi, that's a valid pep8 failure.

Issue #1049, opened 2019-10-14
Project is unmaintained


cpl:
See #1047 and #1048

jd:
I've fixed the title for you. :)

lhuang001:
I am very sad to hear this news. could you tell me how much aodh and ceilometer will be affected？

chungg:
none of the contributors to gnocchi actively participate in openstack so gnocchi's status is unrelated to ceilometer/aodh.

anyone is free to contribute to gnocchi still. we just currently don't have time to fix things like documentation and ci.

unamuno:
Maybe you should change the main README then.

berendt:
@lhuang001 Please check http://lists.openstack.org/pipermail/openstack-discuss/2019-November/010922.html. Probably MongoDB will be used for Ceilometer again.

berendt:
@chungg Can you please include a hint in the README? So it is not really obvious at the moment that there is no active further development taking place.

chungg:
> @lhuang001 Please check http://lists.openstack.org/pipermail/openstack-discuss/2019-November/010922.html. Probably MongoDB will be used for Ceilometer again.

sigh. that thread bums me out.

i don't contribute/follow openstack anymore but if anyone is undecided what to adopt (if gnocchi doesn't fit your needs), we dropped mongo and ceilometer api because it stored and returned an absurd amount of data[1] and required significant resources to be thrown at it for results ranging from 'not good' to 'hot garbage'. 

whatever solution you choose for monitoring/billing/etc... don't choose ceilometer api+mongo unless:

1. your cloud has an insignificant number of users
2. they rewrite the storage schema and change the ceilometer api response aka everything
3. you're a sadist

if any of those apply to you, even then, just do some googling and find something better.

[1] https://www.slideshare.net/GordonChung/ceilometer-to-gnocchi/3

tobias-urdin:
There was a thread back in November about this, there are probably multiple people that are willing to step up to take over the maintenance of the project, while this might not boost the development of new features and mostly mean a working product and hopefully bug fixes. [1].

It was never decided on how the proceed so need to follow up on that, either the project will be forked back to OpenStack infrastructure or hopefully somebody can get help from the Gnocchi team to be added as a maintainer which can maintain the CI and review bugfixes.

I would volunteer some of my time to keep the CI and reviewing bug, I can hopefully take some more time to dig deeper on how to resolve, hopefully sooner than later if nobody beats me to it.

[...]

PR #1059, opened 2020-03-26
Group measures by resource metadata history

pedro-martins:
Problem description
===============

When we are aggregating measures using Gnocchi, we can group the measures by resource's metadata. However, Gnocchi does not group the measures taking into account the resource's metadata history; so if some resource's metadata changed recently, it will be grouped using only the last metadata value. 

Imagine the following situation:

I have a metric with metadata called "name", so I take a resource that is monitored by this metric and I change the resource's name from "test" to "test1" at 14:45:00. Then I call the Gnocchi's Aggregation API with an interval from 11 to 17 hours grouping by "name" and it returns this:

| group                  | name       | timestamp                    | granularity | value               |
|------------------------|------------|------------------------------|-------------|---------------------|
| name: test1 | aggregated | 2020\-03\-26T11:00:00\+00:00 | 3600\.0     | 101 |
| name: test1 | aggregated | 2020\-03\-26T12:00:00\+00:00 | 3600\.0     | 104 |
| name: test1 | aggregated | 2020\-03\-26T13:00:00\+00:00 | 3600\.0     | 98  |
| name: test1 | aggregated | 2020\-03\-26T14:00:00\+00:00 | 3600\.0     | 100  |
| name: test1 | aggregated | 2020\-03\-26T15:00:00\+00:00 | 3600\.0     | 97  |
| name: test1 | aggregated | 2020\-03\-26T16:00:00\+00:00 | 3600\.0     | 109 |

Here we have a problem. For the system consuming Gnocchi API, it will look like the resource has the same attributes in all of the timestamps, when that is not true. This can have a negative impact on billing systems that rely on Gnocchi as the time series database (e.g. CloudKitty).

Proposal
=======

I propose to allow users to specify if they want or not to group their measurements taking into account the resource's metadata history; if they want, they can just add a "use_history=true" parameter in their API request and Gnocchi will do it for them. If users do not specify the "use_history" in the API, the following property will be used instead.

```property
[api]
group_measures_using_history = True/False
```

If the property or the query parameter are not defined, the "use_history" value assumes False (for backward compatibility). The new response expected with this proposal using the same situation listed in `Problem description` will be (using group_measures_using_history=true):

| group                  | name       | timestamp                    | granularity | value               |
|------------------------|------------|------------------------------|-------------|---------------------|
| name: test  | aggregated | 2020\-03\-26T11:00:00\+00:00 | 3600\.0     | 101 |
| name: test  | aggregated | 2020\-03\-26T12:00:00\+00:00 | 3600\.0     | 104 |
| name: test  | aggregated | 2020\-03\-26T13:00:00\+00:00 | 3600\.0     | 98  |
| name: test  | aggregated | 2020\-03\-26T14:00:00\+00:00 | 3600\.0     | 75  |
| name: test1 | aggregated | 2020\-03\-26T14:00:00\+00:00 | 3600\.0     | 25  |
| name: test1 | aggregated | 2020\-03\-26T15:00:00\+00:00 | 3600\.0     | 97  |
| name: test1 | aggregated | 2020\-03\-26T16:00:00\+00:00 | 3600\.0     | 109 |

As you can see, there are 2 entries at 14:00; it happens because the resource's name was changed at 14:45, that is equivalent to 75% of the granularity window (that is 1 hour), so the measure named "test" will take 75% of the total measurement of the granularity window, and the measure named "test1" will take the remaining 25% of the measurement value.

### DIFF
--- a/gnocchi/indexer/sqlalchemy_base.py
+++ b/gnocchi/indexer/sqlalchemy_base.py
@@ -202,6 +202,9 @@ class ResourceType(Base, GnocchiBase, resource_type.ResourceType):
 
 
 class ResourceJsonifier(indexer.Resource):
+    def __str__(self):
+        return str(self.jsonify())
+
     def jsonify(self, attrs=None):
         d = dict(self)
         del d['revision']
--- a/gnocchi/rest/aggregates/api.py
+++ b/gnocchi/rest/aggregates/api.py
@@ -14,9 +14,11 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import daiquiri
 import fnmatch
 import itertools
 
+import numpy
 import pecan
 from pecan import rest
 import pyparsing
@@ -30,6 +32,8 @@ from gnocchi.rest.aggregates import processor
 from gnocchi.rest import api
 from gnocchi import storage
 from gnocchi import utils
+
+LOG = daiquiri.getLogger(__name__)
 
 
 def _OperationsSubNodeSchema(v):
@@ -185,6 +189,244 @@ def get_measures_or_abort(references, operations, start,
                         "detail": [(str(e.metric.id), e.method)]})
 
 
+class MeasureGroup(object):
+    def __init__(self, group_key, resources):
+        self.resources = self.join_sequential_groups(resources)
+        self.group_key = dict(group_key)
+        self.measures = []
+        self.references = None
+
+    def add_measures(self, aggregated_measures, date_map, start, stop,
+                     details):
+        measures = aggregated_measures['measures']['aggregated']
+        if details:
+            self.references = aggregated_measures['references']
+        for measure in measures:
+            self.add_measure(list(measure), date_map, start, stop)
+
+    def add_measure(self, measure, date_map, start, stop):
+        measure = Measure(*(measure + [start, stop]))
+        if not date_map.get(measure.date, None):
+            date_map[measure.date] = []
+        date_map[measure.date].append(measure)
+        self.measures.append(measure)
+
+    def join_sequential_groups(self, group):
+        group.sort(key=lambda key: key['revision_start'])
+        new_group = []
+        last_it = None
+        for it in group:
+            if last_it and it['revision_start'] == last_it['revision_end']:
+                last_it['revision_end'] = it['revision_end']
+                continue
+
+            last_it = it
+            new_group.append(it)
+
+        return new_group
+
+    def sum_groups_same_time_values(self):
+        to_remove = []
+        last_visited = None
+        for measure in self.measures:
+            if last_visited and last_visited.date == measure.date:
+                to_remove.append(last_visited)
+                measure.value += last_visited.value
+
+            last_visited = measure
+
+        self.measures = [m for m in self.measures if m not in to_remove]
+
+
+class Measure(object):
+    def __init__(self, date, granularity, value, start, stop):
+        self.date = date
+        self.granularity = granularity
+        self.value = value
+        self.start = (numpy.datetime64(start) if start
+                      else date)
+        measure_next_window_measurement = self.date + self.granularity
+        self.stop = (numpy.datetime64(stop) if stop
+                     else measure_next_window_measurement)
+        measure_expected_beg = max(self.start, self.date)
+        measure_expected_end = min(self.stop, measure_next_window_measurement)
+        measure_delta = measure_expected_end - measure_expected_beg
+        measure_delta_ns = measure_delta.astype('timedelta64[ns]')
+        self.usage_coefficient = (measure_delta_ns.astype(float) /
+                                  self.granularity.astype(float))
+
+
+class Grouper(object):
+    def __init__(self, groups, start, end, body, sorts, attr_filter,
+                 references, granularity, needed_overlap, fill, details):
+        self.groups = groups
+        self.start = start
+        self.end = end
+        self.body = body
+        self.sorts = sorts
+        self.attr_filter = attr_filter
+        self.references = references
+        self.granularity = granularity
+        self.needed_overlap = needed_overlap
+        self.fill = fill
+        self.details = details
+        self.measures_date_map = {}
+        self.grouped_response = None
+
+    def create_history_period_filter(self):
+        period_filter = {
+            "and": [
+                {
+                    "<": {'revision_start': self.end}
+                },
+                {
+                    "or": [
+                        {">=": {'revision_end': self.start}},
+                        {"=": {'revision_end': None}}
+                    ]
+                }
+            ]
+        }
+        if not (self.start and self.end):
+            return self.attr_filter
+
+        if not self.attr_filter:
+            return period_filter
+
+        return {"and": [period_filter, self.attr_filter]}
+
+    def retrieve_resources_history(self):
+        return pecan.request.indexer.list_resources(
+            self.body["resource_type"],
+            attribute_filter=self.create_history_period_filter(),
+            history=True,
+            sorts=self.sorts)
+
+    def get_grouped_measures(self):
+        resources = self.retrieve_resources_history()
+        self.grouped_response = self.get_measures(self.group(resources))
+        response = self.format_response()
+        LOG.debug("[ Resources History: %s ]", resources)
+        LOG.debug("[ Response: %s ]", response)
+        return response
+
+    def group(self, to_group):
+        to_group.sort(key=lambda g: g['revision_start'])
+        is_first = True
+        for value in to_group:
+            self.truncate_resource_time_window(value, is_first=is_first)
+            is_first = False
+        to_group.sort(key=lambda x: tuple((attr, x[attr])
+                                          for attr in self.groups))
+        grouped_values = \
+            itertools.groupby(
+                to_group,
+                lambda x: tuple((attr, x[attr]) for attr in self.groups))
+
+        grouped = []
+        for key, values in grouped_values:
+            resources = []
+            for value in values:
+                resources.append(value)
+            grouped.append(MeasureGroup(key, resources))
+
+        return grouped
+
+    def truncate_resource_time_window(self, value, is_first=False):
+        if is_first:
+            value['revision_start'] = self.start
+        elif self.start:
+            if value['revision_start']:
+                revision_start = numpy.datetime64(value['revision_start'])
+                value['revision_start'] = max(revision_start, self.start)
+            else:
+                value['revision_start'] = self.start
+
+        if self.end:
+            if value['revision_end']:
+                revision_end = numpy.datetime64(value['revision_end'])
+                value['revision_end'] = min(revision_end, self.end)
+            else:
+                value['revision_end'] = self.end
+
+    def get_measures(self, groups):
+        for group in groups:
+            date_map = self.get_date_map(group)
+            for resource in group.resources:
+                start = numpy.datetime64(resource['revision_start']) \
+                    if resource['revision_start'] else None
+                stop = numpy.datetime64(resource['revision_end']) \
+                    if resource['revision_end'] else None
+
+                LOG.debug("[ Collecting measures from %s to %s ]",
+                          start, stop)
+
+                try:
+                    measure = AggregatesController._get_measures_by_name(
+                        [resource], self.references, self.body["operations"],
+                        start, stop, self.granularity, self.needed_overlap,
+                        self.fill, details=self.details)
+                    group.add_measures(measure, date_map, start, stop,
+                                       self.details)
+                except indexer.NoSuchMetric:
+                    continue
+
+        self.truncate_measures()
+        self.sum_sequential_group_metrics(groups)
+        return groups
+
+    def get_date_map(self, group):
+        if 'id' in group.group_key:
+            return self.measures_date_map.setdefault(group.group_key['id'], {})
+
+        return self.measures_date_map
+
+    def sum_sequential_group_metrics(self, groups):
+        for group in groups:
+            group.sum_groups_same_time_values()
+
+    def truncate_measures(self):
+        for measures in self.measures_date_map.values():
+            if isinstance(measures, list):
+                self.truncate_measure(measures)
+                continue
+
+            for measure in measures.values():
+                self.truncate_measure(measure)
+
+    def truncate_measure(self, measures_list):
+        if measures_list and len(measures_list) == 1:
+            return
+        measures_sum = 0
+        for measure in measures_list[:-1]:
+            measure_new_val = (measure.value *
+                               measure.usage_coefficient)
+            measures_sum += measure_new_val
+            measure.value = measure_new_val
+        last_measure = measures_list[-1]
+        last_measure.value = abs(measures_sum - last_measure.value)
+
+    def format_response(self):
+        measures_list = []
+        for group in self.grouped_response:
+            aggregated = []
+            measures = {
+                'measures': {'measures': {'aggregated': aggregated}},
+                'group': group.group_key
+            }
+            if group.references:
+                measures['measures']['references'] = group.references
+
+            for measure in group.measures:
+                aggregated.append((measure.date, measure.granularity,
+                                   measure.value))
+
+            if aggregated:
+                measures_list.append(measures)
+
+        return measures_list
+
+
 def ResourceTypeSchema(resource_type):
     try:
         pecan.request.indexer.get_resource_type(resource_type)
@@ -207,6 +449,8 @@ class AggregatesController(rest.RestController):
     @pecan.expose("json")
     def post(self, start=None, stop=None, granularity=None,
              needed_overlap=None, fill=None, groupby=None, **kwargs):
+
+        use_history = api.get_bool_param('use_history', kwargs)
         details = api.get_bool_param('details', kwargs)
 
         if fill is None and needed_overlap is None:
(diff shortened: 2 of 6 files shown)


jd:
Hi and thank you guys!
Right now the number one issue is https://github.com/gnocchixyz/gnocchi/issues/1049 as you might have seen.

I don't see the CI running on this PR so I feel like it really does not work anymore or has been disabled?

I feel like there's really some work to be done here before we can merge anything.

This PR seems quite large so it might require some time to review, especially since nobody's usually around.

tobias-urdin:
quite a large change, can you please recheck/reopen to get a new CI check and we can try getting this back on the road.

chungg:
@mergify rebase

gnocchi/indexer/sqlalchemy_base.py
```diff
@@ -198,6 +198,9 @@ def to_baseclass(self):
 
 
 class ResourceJsonifier(indexer.Resource):
+    def __str__(self):
+        return self.jsonify().__str__()
```
jd, line 202:
```suggestion
        return str(self.jsonify())
```

gnocchi/opts.py
```diff
@@ -193,6 +193,13 @@ def list_opts():
                        default=10, min=0,
                        help='Number of seconds before timeout when attempting '
                             'to do some operations.'),
+            cfg.BoolOpt('group_measures_using_history', default=False,
```
jd, line 196: I find it always weird to control API behaviour via global config options.
Why isn't that a only flag passed to the API?

gnocchi/rest/aggregates/api.py
```diff
@@ -206,7 +183,14 @@ class AggregatesController(rest.RestController):
 
     @pecan.expose("json")
     def post(self, start=None, stop=None, granularity=None,
-             needed_overlap=None, fill=None, groupby=None, **kwargs):
+             needed_overlap=None, fill=None, groupby=None, use_history=None,
+             **kwargs):
+
+        if use_history is None:
+            use_history = pecan.request.conf.api.group_measures_using_history
```
jd, line 190: Yeah I don't think a global option should change a default behaviour.

run-upgrade-tests.sh
```diff
@@ -70,7 +70,7 @@ pifpaf_stop(){
 
 cleanup(){
     pifpaf_stop
-    rm -rf $GNOCCHI_DATA
+    rm -rf -v $GNOCCHI_DATA
```
jd, line 73: unrelated :)

jd (changes requested):
This is pretty long and complex, so I'm unable to review it deeply. But a quick glance make it looks good so I think we can merge.
The only thing I'd change is the option which I would remove.

gnocchi/opts.py
pedro-martins: This was done to avoid changing every system that uses this API to get the new behavior, so, instread of changing lots of systems to use the new API flag, we just set the configuration in gnocchi, and if some system for some reason need the old behavior, it can simple override the config with the API flag.

gnocchi/indexer/sqlalchemy_base.py
pedro-martins: nice catch, thanks.

run-upgrade-tests.sh
pedro-martins: yea, it was added for some debugging purpose and I forgot to remove it, thanks.

gnocchi/opts.py
jd: This is not a good approach as an API should never ever change and break existing system.

pedro-martins: I will remove the config

gnocchi/rest/aggregates/api.py
pedro-martins: Ok, no problem, I will remove it.

chungg (commented):
don't understand this enough to approve, but not against since it seems leave current functionality as is.

tobias-urdin:
Possible that something changed with the new pyparsing version just released https://github.com/pyparsing/pyparsing/releases - I don't have the time investigate this straight away though, will check later.

gnocchi/rest/aggregates/measures_grouper.py
```diff
@@ -0,0 +1,326 @@
...
+                    group.add_measures(measure, date_map, start, stop,
+                                       self.details)
+                except indexer.NoSuchMetric:
+                    continue
+
+        self.truncate_measures()
+        self.sum_sequential_group_metrics(groups)
+        return groups
+
+    def get_date_map(self, group):
```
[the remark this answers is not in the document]
pedro-martins: This "get_data_map" function will return a map (dict) within all group`s measures grouped by timestamp (date). 

If the group has no measures yet, then it will set the group entry and return the dict reference to be filled with the measures in  the "get_measures" function. 

If the group has no ID them it will return all measures grouped by group and grouped by date

gnocchi/rest/aggregates/measures_grouper.py
```diff
@@ -0,0 +1,326 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import fnmatch
+import itertools
+import logging
+import numpy
+import pecan
```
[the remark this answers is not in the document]
pedro-martins: ok

gnocchi/rest/aggregates/measures_grouper.py
```diff
@@ -0,0 +1,326 @@
...
+        self.grouped_response = None
+
+    def create_history_period_filter(self):
+        period_filter = {
+            "and": [
+                {
+                    "<": {'revision_start': self.end}
+                },
+                {
+                    ">": {'revision_end': self.start}
```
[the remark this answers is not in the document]
pedro-martins: yes

gnocchi/rest/aggregates/measures_grouper.py
```diff
@@ -0,0 +1,326 @@
...
+            needed_overlap, fill)
+    }
+    if details:
+        response["references"] = set((r.resource for r in references))
+    return response
+
+
+class Measures(object):
+    def __init__(self):
+        self.groups = []
```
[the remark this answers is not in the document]
pedro-martins: removed

tobias-urdin:
Can this conflict be resolved and then we can make have a another look at this and merge it? Thanks!

tobias-urdin:
pyparsing changed the output so this needs a change like was done in https://github.com/gnocchixyz/gnocchi/pull/1185/commits/70da17a90a13151a307c51bc26bf6994c1ff222a

pedro-martins:
Thanks tobias, fix applied.

gnocchi/rest/aggregates/api.py
```diff
@@ -238,36 +219,35 @@ def post(self, start=None, stop=None, granularity=None,
 
             groupby = sorted(set(api.arg_to_list(groupby)))
             sorts = groupby if groupby else api.RESOURCE_DEFAULT_PAGINATION
-            try:
+            if not groupby:
                 resources = pecan.request.indexer.list_resources(
```
tobias-urdin, line 223: Please add this back into a try statement and catch the IndexerException and return 400 as was done before. You can move it inside the try statement below.

tobias-urdin (changes requested):
A small thing to change otherwise LGTM, maybe @jd you can take a final look as well?

gnocchi/tests/test_measures_grouper.py
```diff
@@ -0,0 +1,398 @@
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
```
tobias-urdin, line 1: I can't see that this tests is ever executed in CI testing?

tobias-urdin (changes requested):

gnocchi/rest/aggregates/api.py
pedro-martins: Yes, sure, I missed that. 
Thanks.

gnocchi/tests/test_measures_grouper.py
pedro-martins: I checked, indeed, they were not executed, I fixed it as well.

Thanks again.

gnocchi/rest/aggregates/api.py
```diff
@@ -14,7 +14,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-import fnmatch
+from distutils import util
```
tobias-urdin, line 17: Please dont import distutils here but use gnocchi.rest.api.get_bool_param as already used in this file.

gnocchi/rest/aggregates/api.py
```diff
@@ -206,7 +183,11 @@ class AggregatesController(rest.RestController):
 
     @pecan.expose("json")
     def post(self, start=None, stop=None, granularity=None,
-             needed_overlap=None, fill=None, groupby=None, **kwargs):
+             needed_overlap=None, fill=None, groupby=None, use_history="False",
+             **kwargs):
+
+        use_history = util.strtobool(use_history)
```
tobias-urdin, line 189: use api.get_bool_param

tobias-urdin (changes requested):
Some more feedback on this patch now that I've managed to fix the CI again. My intent is to merge this PR now after walking through it but have some more things, see the review.

@jd my intent is to merge this, please provide feedback on this patch if you have any.

tobias-urdin:
Please rebase against master as well to get all the changes for the CI to work as well.

gnocchi/rest/aggregates/api.py
pedro-martins: sure, thanks.

gnocchi/rest/aggregates/api.py
pedro-martins: Ok, thanks :slightly_smiling_face:

gnocchi/rest/aggregates/api.py
```diff
@@ -313,27 +295,48 @@ def groupper(r):
...
+        try:
+            grouper = measures_grouper.Grouper(groupby, start, stop, body,
+                                               sorts, attr_filter, references,
+                                               granularity, needed_overlap,
+                                               fill, details)
+            results = grouper.get_grouped_measures()
+        except indexer.NoSuchMetric:
+            pass
+        except indexer.IndexerException as e:
+            api.abort(400, six.text_type(e))
```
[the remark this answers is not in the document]
pedro-martins: We dont have a reference to LOG in this scope, I will add one using the 'logging' module, should I use the 'daiquiri' module instead? or is the 'logging' ok?

Also, for this generic exception, catching, in some cases we have already handled exceptions (that are the httpexceptions one) so I will add a new except to ignore these ones to avoid breaking some tests and API compatibilities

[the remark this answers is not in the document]
tobias-urdin: I think logging is OK since it's used in other places. I don't think you need to introduce anything other than what I suggested since it's just a nicer way of doing it and it would catch anything.

Basicly, just throwing the exception would cause the same behavior anyway, the HTTP 500 error. It's just a matter if we should "leak" the exception message in the response or not. Maybe we shouldn't, what do you think?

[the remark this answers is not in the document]
tobias-urdin: Thinking about it, we should probably not return the exception, only log it. And you can skip the weboc.HTTPExceptions

gnocchi/rest/aggregates/api.py
```diff
@@ -238,36 +220,38 @@ def post(self, start=None, stop=None, granularity=None,
...
+                    results = self.get_measures_grouping(
+                        body, details, fill, granularity, needed_overlap,
+                        references, resources, start, stop, groupby)
+
+            except indexer.NoSuchMetric as e:
+                api.abort(404, e)
+            except indexer.IndexerException as e:
+                api.abort(400, six.text_type(e))
+            except Exception as e:
+                LOG.exception(e)
```
jd, line 252: probably not the right place to log

jd (commented):
honestly I don't have enough time to review this, I'm sorry

gnocchi/rest/aggregates/api.py
pedro-martins: Hmm, maybe, where do you suggest I should put this log? I cant think in another place. Thanks :slightly_smiling_face:

gnocchi/rest/aggregates/measures_grouper.py
```diff
@@ -0,0 +1,305 @@
...
+        self.measures = []
+        self.references = None
+
+    def add_measures(self, aggregated_measures, date_map, start, stop,
+                     details):
+        measures = aggregated_measures['measures']['aggregated']
+        if details:
+            self.references = aggregated_measures['references']
+        for measure in measures:
+            self.add_measure(list(measure), date_map, start, stop)
```
chungg, line 88: hmm... the performance of this is a bit concerning. it seems to be going through each timeseries point by point and trying to slot in to a group (in python)

chungg (commented):
i imagine this would only make sense for the rate/delta metrics? kind of wondering how it'd be used for other metric types? i keep thinking arbitrarily dividing something like cpu utilisation's aggregated measure across two groups doesn't make sense and dividing something like disk.storage.bytes even less.

gnocchi/rest/aggregates/api.py
```diff
@@ -313,27 +297,46 @@ def groupper(r):
...
+                                               sorts, attr_filter, references,
+                                               granularity, needed_overlap,
+                                               fill, details)
+            results = grouper.get_grouped_measures()
+        except indexer.NoSuchMetric:
+            pass
+        return results
+
     @staticmethod
     def _get_measures_by_name(resources, metric_wildcards, operations,
```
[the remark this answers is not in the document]
tobias-urdin: Including the exception however could include information about internals that should not be exposed, so either we should just leave it and give a 500 error and treat it as a bug, or return 400 and hope that it's valid. I'm more for 500 than returning 400 and it's potentially not an input issue.

gnocchi/rest/aggregates/api.py
tobias-urdin: See my comment below on the same point.

gnocchi/rest/aggregates/api.py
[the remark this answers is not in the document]
pedro-martins: Yes, but in that case, we have some problems to discuss here:
 - About line 252
   - If we catch the Exception and abort it with an error 500 or 400, in some cases the Exception is an instance of `weboc.HTTPExceptions` that is some already handled exception, so in the case this Exception raises an HTTP 400 and we catch it to an HTTP 500, it will be an inconsistency. To try to solve it, we could catch an `weboc.HTTPExceptions` before catching the Exception itself and and ignore it.
   - If we dont catch the generic Exception, we will delegate to who called this method to handle it, and its ok in my opinion, and that is what we are doing here right now, except that we are logging the Exception before throwing it. This is the initial behavior of this procedure (we only added the logging stuff).
 - About line 331. No reason, I missed this one, sorry.

gnocchi/rest/aggregates/measures_grouper.py
```diff
@@ -0,0 +1,305 @@
...
+import numpy
+import pecan
+
+from gnocchi import indexer
+from gnocchi.rest.aggregates import exceptions
+from gnocchi.rest.aggregates import processor
+from gnocchi.rest import api
+from gnocchi import storage
+
+LOG = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
pedro-martins: Sure, no problem, thanks :slightly_smiling_face:

gnocchi/rest/aggregates/api.py
```diff
@@ -24,13 +24,14 @@
...
 from gnocchi import indexer
-from gnocchi.rest.aggregates import exceptions
+from gnocchi.rest.aggregates import measures_grouper
 from gnocchi.rest.aggregates import operations as agg_operations
 from gnocchi.rest.aggregates import processor
 from gnocchi.rest import api
-from gnocchi import storage
 from gnocchi import utils
 
+LOG = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
pedro-martins: Ok, thanks.

gnocchi/rest/aggregates/measures_grouper.py
```diff
@@ -0,0 +1,305 @@
...
+                     else measure_next_window_measurement)
+        measure_expected_beg = max(self.start, self.date)
+        measure_expected_end = min(self.stop, measure_next_window_measurement)
+        measure_delta = measure_expected_end - measure_expected_beg
+        measure_delta_ns = measure_delta.astype('timedelta64[ns]')
+        self.usage_coefficient = (measure_delta_ns.astype(float) /
+                                  self.granularity.astype(float))
+
+
+class Grouper(object):
```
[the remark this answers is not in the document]
pedro-martins: We created a class just to try to organize things up when  codding this one. Is there any problem? We could change it if necessary.

gnocchi/rest/aggregates/measures_grouper.py
chungg: got it. (wasn't aware nova kept same id when flavour changes). i do understand the problem you're trying to solve, i guess my question is what the correct behaviour is to attribute ownership when dividing up an measure.

for instance, if we're capturing cpu.time|storage.bytes which is a gauge metric that (for the most part) only grows. if we split, it with current logic say it might end up being:
- 00:00:00, value 0, meta1
- 10:00:00, value 10, meta1
- 11:00:00, value 1.1; meta1(11:00 - 11:06)
- 11:00:00, value 8,8: meta2(11:06 - 11:54)
- 11:00:00, value 1.1, meta3 (11:54 - 12:00)

which would be an odd value considering it was meta1 for 11hrs and meta2 for 48min but the values are near identical. maybe i'm over-complicating this.

gnocchi/rest/aggregates/api.py
pedro-martins: It will not return an empty result, it will raise the Exception again (and maybe return error 500, or 404, it will depends how the caller is handling exceptions, but it is how it was implemented before (except by the log stuff)

gnocchi/rest/aggregates/measures_grouper.py
pedro-martins: Got your point, and yes, your are right, this kind of behavior does not make much sense when using gauge metrics, they are cumulative. 

You should not configure metadata sensitive resources using gauge metrics when you have floating metadata (each possible metadata should be a resource itself to gauge make sense), because any time the metadata changes you should reset the measures values to zero (because it means that this is a new one) and pause the old metadata collecting until the resource changes its metadata again to the old value (togging between metadata values).

For these cases, if you want to use gauge metrics, you should create one metric for any possible metadata value in the resource and measure them separated in your data polling service. 

So if this is the case, you could continue using the normal aggregation API without sending the parameter "user_history" in the request, so the default behavior will me maintained.

chungg (commented):
if i'm honest, i'm not entirely sure this is the ideal implementation or behaviour but i'm also not as involved anymore and can see it's a valid use case so am not against it. if others are ok with it, i would suggest docs and some api tests.

pedro-martins:
> if i'm honest, i'm not entirely sure this is the ideal implementation or behaviour but i'm also not as involved anymore and can see it's a valid use case so am not against it. if others are ok with it, i would suggest docs and some api tests.

I added some gabbit tests and rest docs regarding the changes.

tobias-urdin (approved):

PR #1062, opened 2020-04-14
Enable attribute update on resource-type API.

pedro-martins:
Currently, the resource-type API, not allow operators to update
resource type's attributes, the API only allows Add and Delete.

So, if an operator desires to change a required resource type's
attribute to optional and vice-versa, he/she will need to do it
manually in the database or deleting and creating the attribute
again, but it will result in the attribute's history lose.

Therefore, to help operators in that situation, I propose to
enhance the Resource-type API to allow updates while adding a
new attribute. If the new attribute already exists and is
different from the old, it will be updated; otherwise, nothing
will happen and we will return the resource type's attribute
list as the old behavior.

The first version of this implementation will not allow updating
the resource type's attribute's type.


gnocchi/rest/api.py
```diff
@@ -901,24 +901,75 @@ def patch(self):
...
+            update_attrs_old.sort(key=lambda attr: attr.name)
+            update_attrs = []
+            for i in range(len(update_attrs_new)):
+                update_attrs.append((update_attrs_new[i],
+                                     update_attrs_old[i]))
+
+            return update_attrs
+        except Exception as e:
+            LOG.exception(e)
+            raise e
```
jd, line 945: This is likely log elsewhere then, that's not a really good pattern.

gnocchi/rest/api.py
```diff
@@ -901,24 +901,75 @@ def patch(self):
...
+        return update_attrs
+
+    def validate_types(self, attribute, new_json, old_json):
+        old_type = old_json['type']
+        new_type = new_json["attributes"][attribute]['type']
+        if new_type != old_type:
+            msg = "Type update is not available yet. Changing %s to %s " \
+                  "for attribute %s of resource %s" % (old_type, new_type,
+                                                       attribute, self._name)
+            LOG.warning(msg)
```
jd, line 970: You shouldn't log user requests.
```suggestion
```

jd (changes requested):
There's a conflict to fix too :)

gnocchi/rest/api.py
pedro-martins: I changed this one to abort the request with an error.

gnocchi/rest/api.py
pedro-martins: removed.

gnocchi/rest/api.py
```diff
@@ -901,24 +901,75 @@ def patch(self):
...
+            update_attrs_old = schema.attributes_from_dict(old_attrs)
+            update_attrs_old.sort(key=lambda attr: attr.name)
+            update_attrs = []
+            for i in range(len(update_attrs_new)):
+                update_attrs.append((update_attrs_new[i],
+                                     update_attrs_old[i]))
+
+            return update_attrs
+        except Exception:
+            abort(400, "Unable to update attributes [%s]." % update_attrs)
```
jd, line 946: That does not look like a 400: if it's a 400 you must be able to explain what's wrong to the client.
If you're unable to execute the request it's more likely a 500 and a bug in the code itself.
I'd rather not try/except anything here.

jd (changes requested):

jd (commented):
This looks good overall, but can we have a few functional tests?

mrunge (commented):
Thank you, this looks good to me.

jd (approved):

PR #1064, opened 2020-07-10
Ensure member statistics key is decoded

tobias-urdin:
In the StatusController the statistics for the
members on python3 is a byte string which cannot
be jsonified.

This changes so it loops through the statistics
for the member and ensures the key is a string
and not a byte string.


tobias-urdin:
Merge this when testing is fixed.

tobias-urdin:
requesting review @jd @chungg

chungg:
feel free to push back but it might be useful to add a test to https://github.com/gnocchixyz/gnocchi/blob/cb8a0e46c8815e0bd1dc25ff01c58eb9debe79a0/gnocchi/tests/functional/gabbits/base.yaml

i believe the functional tests do have a coordinator set up and should return 'statistics'? if not, then feel free to ignore.

tobias-urdin:
not sure what tests i can add for this, need to investigate if it's required

gnocchi/rest/api.py
```diff
@@ -2116,10 +2116,14 @@ def get(details=True):
             report_dict['metricd']['processors'] = [
                 member.decode() for member in members
             ]
-            report_dict['metricd']['statistics'] = {
-                member.decode(): cap.get()
-                for member, cap in six.moves.zip(members, caps)
-            }
+            members_data = {}
+            for member, cap in six.moves.zip(members, caps):
```
jd, line 2120: You can stop using `six` here.

jd (commented):
+1 on the test.
Also since we don't support Python 2 anymore, I'd go for a simpler version of the code maybe?

gnocchi/rest/api.py
tobias-urdin: Considering there is a lot of pending fixes, I would like that we merge fixes, do a new release for the 4.3.x series (which would still include python2 support even though we don't explicitly test it now) then prepare for 4.4 or 5.0 (better?) and then drop all compatibility (six usage) etc.

tobias-urdin: Feedback on how to test this change is appreciated.

tobias-urdin:
Testing new docker image from https://github.com/gnocchixyz/ci-tools/pull/32

tobias-urdin:
@mergifyio rebase

jd:
@Mergifyio rebase

tobias-urdin:
@Mergifyio backport-to-4.3

PR #1065, opened 2020-07-11
Fixes for testing

tobias-urdin:
Testing some changes to fix testing.

### DIFF
--- a/.testr.conf
+++ b/.testr.conf
@@ -1,5 +1,5 @@
 [DEFAULT]
-test_command=${PYTHON:-python} -m subunit.run discover -t .  ${GNOCCHI_TEST_PATH:-gnocchi/tests} $LISTOPT $IDOPTION
+test_command=${PYTHON:-python3} -m subunit.run discover -t .  ${GNOCCHI_TEST_PATH:-gnocchi/tests} $LISTOPT $IDOPTION
 test_id_option=--load-list $IDFILE
 test_list_option=--list
 group_regex=(gabbi\.suitemaker\.test_gabbi((_live_|_)([^_]+)))_
--- a/.testr.conf
+++ b/.testr.conf
@@ -1,5 +1,5 @@
 [DEFAULT]
-test_command=${PYTHON:-python3} -m subunit.run discover -t .  ${GNOCCHI_TEST_PATH:-gnocchi/tests} $LISTOPT $IDOPTION
+test_command=${PYTHON:-python} -m subunit.run discover -t .  ${GNOCCHI_TEST_PATH:-gnocchi/tests} $LISTOPT $IDOPTION
 test_id_option=--load-list $IDFILE
 test_list_option=--list
 group_regex=(gabbi\.suitemaker\.test_gabbi((_live_|_)([^_]+)))_
--- a/doc/source/contributing.rst
+++ b/doc/source/contributing.rst
@@ -66,16 +66,16 @@ a virtual environment for each test environment, so make sure you are using an
 up to date version of `virtualenv <https://pypi.python.org/pypi/virtualenv>`_.
 
 Different test environments and configurations can be found by running the
-``tox -l`` command. For example, to run tests with Python 2.7, PostgreSQL as
-indexer, and file as storage backend:
+``tox -l`` command. For example, to run tests with PostgreSQL as indexer,
+and file as storage backend:
 
 ::
 
-    tox -e py27-postgresql-file
+    tox -e py37-postgresql-file
 
 
-To run tests with Python 2.7, MySQL as indexer, and Ceph as storage backend:
+To run tests with MySQL as indexer, and Ceph as storage backend:
 
 ::
 
-    tox -e py35-mysql-ceph
+    tox -e py37-mysql-ceph
--- a/doc/source/contributing.rst
+++ b/doc/source/contributing.rst
@@ -66,8 +66,8 @@ a virtual environment for each test environment, so make sure you are using an
 up to date version of `virtualenv <https://pypi.python.org/pypi/virtualenv>`_.
 
 Different test environments and configurations can be found by running the
-``tox -l`` command. For example, to run tests with PostgreSQL as indexer,
-and file as storage backend:
+``tox -l`` command. For example, to run tests with Python 3.7, PostgreSQL as
+indexer, and file as storage backend:
 
 ::
 
--- a/gnocchi/tests/functional/gabbits/resource-type.yaml
+++ b/gnocchi/tests/functional/gabbits/resource-type.yaml
@@ -768,6 +768,8 @@ tests:
             min_length: 0
             max_length: 255
       status: 400
+      response_strings:
+       - "can't remove non-existent object 'what'"
 
     - name: patch a resource attribute replace
       PATCH: /v1/resource_type/my_custom_resource
@@ -816,6 +818,8 @@ tests:
         - op: remove
           path: /attributes/unknown
       status: 400
+      response_strings:
+       - "can't remove non-existent object 'unknown'"
 
 # Ensure we can't delete the type
 
--- a/gnocchi/tests/functional/gabbits/resource-type.yaml
+++ b/gnocchi/tests/functional/gabbits/resource-type.yaml
@@ -769,7 +769,7 @@ tests:
             max_length: 255
       status: 400
       response_strings:
-       - "can't remove non-existent object 'what'"
+       - "can't remove a non-existent object 'what'"
 
     - name: patch a resource attribute replace
       PATCH: /v1/resource_type/my_custom_resource
@@ -819,7 +819,7 @@ tests:
           path: /attributes/unknown
       status: 400
       response_strings:
-       - "can't remove non-existent object 'unknown'"
+       - "can't remove a non-existent object 'unknown'"
 
 # Ensure we can't delete the type
 
--- a/gnocchi/tests/functional/gabbits/resource-type.yaml
+++ b/gnocchi/tests/functional/gabbits/resource-type.yaml
@@ -768,8 +768,6 @@ tests:
             min_length: 0
             max_length: 255
       status: 400
-      response_strings:
-       - "can't remove non-existent object 'what'"
 
     - name: patch a resource attribute replace
       PATCH: /v1/resource_type/my_custom_resource
@@ -818,8 +816,6 @@ tests:
         - op: remove
           path: /attributes/unknown
       status: 400
-      response_strings:
-       - "can't remove non-existent object 'unknown'"
 
 # Ensure we can't delete the type
 
--- a/run-upgrade-tests.sh
+++ b/run-upgrade-tests.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
-set -xe
+set -e
 
 if [ "$1" == "postgresql-file" ]; then
-  eval $(pifpaf --debug --env-prefix INDEXER run postgresql)
+  eval $(pifpaf --env-prefix INDEXER run postgresql)
 elif [ "$1" == "mysql-ceph" ]; then
-  eval $(pifpaf --debug --env-prefix INDEXER run mysql)
-  eval $(pifpaf --debug --env-prefix STORAGE run ceph)
+  eval $(pifpaf --env-prefix INDEXER run mysql)
+  eval $(pifpaf --env-prefix STORAGE run ceph)
 else
   echo "error: unsupported upgrade type"
   exit 1
@@ -13,7 +13,7 @@ fi
 
 export GNOCCHI_DATA=$(mktemp -d -t gnocchi.XXXX)
 
-old_version=$(pip3 freeze | sed -n '/gnocchi==/s/.*==\(.*\)/\1/p')
+old_version=$(pip freeze | sed -n '/gnocchi==/s/.*==\(.*\)/\1/p')
 
 RESOURCE_IDS=(
     "5a301761-aaaa-46e2-8900-8b4f6fe6675a"
@@ -50,7 +50,7 @@ inject_data() {
 
     {
         measures_sep=""
-        MEASURES=$(python3 -c 'import datetime, random, json; now = datetime.datetime.utcnow(); print(json.dumps([{"timestamp": (now - datetime.timedelta(seconds=i)).isoformat(), "value": random.uniform(-100000, 100000)} for i in range(0, 288000, 10)]))')
+        MEASURES=$(python -c 'import datetime, random, json; now = datetime.datetime.utcnow(); print(json.dumps([{"timestamp": (now - datetime.timedelta(seconds=i)).isoformat(), "value": random.uniform(-100000, 100000)} for i in range(0, 288000, 10)]))')
         echo -n '{'
         resource_sep=""
         for resource_id in ${RESOURCE_IDS[@]} $RESOURCE_ID_EXT; do
@@ -71,6 +71,8 @@ pifpaf_stop(){
 cleanup(){
     pifpaf_stop
     rm -rf $GNOCCHI_DATA
+    indexer_stop || true
+    [ "$STORAGE_DAEMON" == "ceph" ] && storage_stop || true
 }
 trap cleanup EXIT
 
@@ -82,7 +84,7 @@ else
     STORAGE_URL=file://$GNOCCHI_DATA
 fi
 
-eval $(pifpaf --debug run gnocchi --indexer-url $INDEXER_URL --storage-url $STORAGE_URL)
+eval $(pifpaf run gnocchi --indexer-url $INDEXER_URL --storage-url $STORAGE_URL)
 export OS_AUTH_TYPE=gnocchi-basic
 export GNOCCHI_USER=$GNOCCHI_USER_ID
 original_statsd_resource_id=$GNOCCHI_STATSD_RESOURCE_ID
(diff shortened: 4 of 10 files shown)


tobias-urdin:
Waiting for this [1], let's see if a updated image helps here.

[1] https://github.com/gnocchixyz/ci-tools/pull/31

setup.cfg
```diff
@@ -84,15 +84,16 @@ prometheus =
...
     sphinx_rtd_theme
     sphinxcontrib-httpdomain
     PyYAML
     Jinja2
     reno>=1.6.2
 test =
-    pifpaf[ceph,gnocchi]>=1.0.1
-    gabbi>=1.37.0
+    pifpaf[ceph,gnocchi]>=1.0.1,<=2.6.0
+    gabbi==1.49.0
```
jd, line 96: maybe provide a range?

tox.ini
```diff
@@ -56,10 +56,12 @@ commands =
 recreate = True
 setenv = GNOCCHI_VARIANT=test,postgresql,file
 deps = gnocchi[{env:GNOCCHI_VARIANT}]>=4.3,<4.4
-  pifpaf[gnocchi]>=0.13
+  pifpaf[gnocchi]>=0.13,<=2.6.0
```
jd, line 59: I think `<3` might be more accurate.

tox.ini
```diff
@@ -56,10 +56,12 @@ commands =
...
   gnocchiclient>=2.8.0
   xattr!=0.9.4
-commands = pifpaf --env-prefix INDEXER run postgresql {toxinidir}/run-upgrade-tests.sh {posargs}
+commands =
+  pifpaf --env-prefix INDEXER run postgresql
+  {toxinidir}/run-upgrade-tests.sh {posargs}
```
jd, line 64: unrelated?

tox.ini
```diff
@@ -68,16 +70,21 @@ recreate = True
 setenv = GNOCCHI_VARIANT=test,mysql,ceph,ceph_recommended_lib
 deps = gnocchi[{env:GNOCCHI_VARIANT}]>=4.3,<4.4
   gnocchiclient>=2.8.0
-  pifpaf[ceph,gnocchi]>=0.13
+  pifpaf[ceph,gnocchi]>=0.13,<=2.6.0
```
jd, line 73: ditto

tox.ini
```diff
@@ -68,16 +70,21 @@ recreate = True
...
   xattr!=0.9.4
-commands = pifpaf --env-prefix INDEXER run mysql -- pifpaf --env-prefix STORAGE run ceph {toxinidir}/run-upgrade-tests.sh {posargs}
+commands =
+  pifpaf --env-prefix INDEXER run mysql
+  pifpaf --env-prefix STORAGE run ceph
+  {toxinidir}/run-upgrade-tests.sh {posargs}
```
jd, line 78: ditto

tox.ini
```diff
@@ -95,8 +102,10 @@ deps =
     .[test,file,postgresql,doc]
     doc8
 setenv = GNOCCHI_TEST_DEBUG=1
-commands = doc8 --ignore-path doc/source/rest.rst,doc/source/comparison-table.rst doc/source
-           pifpaf -g GNOCCHI_INDEXER_URL run postgresql -- python setup.py build_sphinx -W
+commands =
+  doc8 --ignore-path doc/source/rest.rst,doc/source/comparison-table.rst doc/source
+  pifpaf -g GNOCCHI_INDEXER_URL run postgresql
+  python setup.py build_sphinx -W
```
jd, line 108: now this is even different

jd (changes requested):

tobias-urdin:
Yeah, some random changes there. Annoying since I had everything passing locally some weeks ago but now something else is broken. The upgrade testing works locally but not in CI.

chungg:
might make our lives (marginally) easier, but can we drop py2.7 tests?

jd:
Yeah we should remove Python 2.7 support now so it'd be fine to remove the tests as a first step.

Adrian-Turjak:
Dropping py2 from testing, and then some follow up work to outright drop py2 (and support py3-isms) isn't a bad idea at all.

tobias-urdin:
The upgrade tests are executed now, but fails when it tries to inject data. For some reason the gnocchi-metricd daemon that is running doesn't process the data. If you start another gnocchi-metricd process in another window it will start processing and continue.

.testr.conf
```diff
@@ -1,5 +1,5 @@
 [DEFAULT]
-test_command=${PYTHON:-python} -m subunit.run discover -t .  ${GNOCCHI_TEST_PATH:-gnocchi/tests} $LISTOPT $IDOPTION
+test_command=${PYTHON:-python3} -m subunit.run discover -t .  ${GNOCCHI_TEST_PATH:-gnocchi/tests} $LISTOPT $IDOPTION
```
jd, line 2: This should not be needed if you only use Python with tox and Python 3.

doc/source/contributing.rst
```diff
@@ -66,16 +66,16 @@ a virtual environment for each test environment, so make sure you are using an
 up to date version of `virtualenv <https://pypi.python.org/pypi/virtualenv>`_.
 
 Different test environments and configurations can be found by running the
-``tox -l`` command. For example, to run tests with Python 2.7, PostgreSQL as
-indexer, and file as storage backend:
+``tox -l`` command. For example, to run tests with PostgreSQL as indexer,
```
jd, line 69:
```suggestion
``tox -l`` command. For example, to run tests with Python 3.7, PostgreSQL as indexer,
```

gnocchi/tests/functional/gabbits/resource-type.yaml
```diff
@@ -768,8 +768,6 @@ tests:
             min_length: 0
             max_length: 255
       status: 400
-      response_strings:
```
jd, line 771: This is actually removing a useful test. Did the error reported changed?

run-upgrade-tests.sh
```diff
@@ -1,9 +1,19 @@
...
+  eval $(pifpaf --debug --env-prefix STORAGE run ceph)
+else
+  echo "error: unsupported upgrade type"
+  exit 1
+fi
 
 export GNOCCHI_DATA=$(mktemp -d -t gnocchi.XXXX)
 
-old_version=$(pip freeze | sed -n '/gnocchi==/s/.*==\(.*\)/\1/p')
+old_version=$(pip3 freeze | sed -n '/gnocchi==/s/.*==\(.*\)/\1/p')
```
jd, line 16: ditto should not be needed

run-upgrade-tests.sh
```diff
@@ -40,7 +50,7 @@ inject_data() {
 
     {
         measures_sep=""
-        MEASURES=$(python -c 'import datetime, random, json; now = datetime.datetime.utcnow(); print(json.dumps([{"timestamp": (now - datetime.timedelta(seconds=i)).isoformat(), "value": random.uniform(-100000, 100000)} for i in range(0, 288000, 10)]))')
+        MEASURES=$(python3 -c 'import datetime, random, json; now = datetime.datetime.utcnow(); print(json.dumps([{"timestamp": (now - datetime.timedelta(seconds=i)).isoformat(), "value": random.uniform(-100000, 100000)} for i in range(0, 288000, 10)]))')
```
jd, line 53: ditto

run-upgrade-tests.sh
```diff
@@ -72,17 +82,17 @@ else
...
+eval $(pifpaf --debug run gnocchi --indexer-url $INDEXER_URL --storage-url $STORAGE_URL)
 export OS_AUTH_TYPE=gnocchi-basic
 export GNOCCHI_USER=$GNOCCHI_USER_ID
 original_statsd_resource_id=$GNOCCHI_STATSD_RESOURCE_ID
 inject_data $GNOCCHI_DATA
 dump_data $GNOCCHI_DATA/old
 pifpaf_stop
 
-new_version=$(python setup.py --version)
+new_version=$(python3 setup.py --version)
```
jd, line 93: ditto

jd (changes requested):
a lot of the python -> python3 should not be needed

gnocchi/tests/functional/gabbits/resource-type.yaml
tobias-urdin: Yes, changed the output. Will revert and check output again.

jd (commented):
Looks like upgrade fails because it's too verbose :)

tobias-urdin:
Feedback on how to change gabbits response_string if this doesn't work :)

jd:
Gentle ping on @cdent if he knows something changed in Gabbi :)

jd:
I wonder if the upgrade problem is not due to gnocchiclient recent issue?

cdent:
> Gentle ping on @cdent if he knows something changed in Gabbi :)

Can you point me to where things are going wrong?

One thing I did notice is that regex (wrapping strings with `/`) are not available in `response_strings`; it checks the entire response body for the string fragment, so that's a thing that needs to be fixed.

But if you can link me a failure I can provide more.

tobias-urdin:
@cdent https://travis-ci.org/github/gnocchixyz/gnocchi/jobs/725172818#L447

Is there a way to check if the response_body contains the value instead of what it tries there to check the complete value? Seems like the output has changed in that response from the API probably due to some dependency (uwsgi or whatever it might be).

cdent:
> Is there a way to check if the response_body contains the value instead of what it tries there to check the complete value?

Thanks for the link. Yeah, the issue there is that you're trying to use a regex in a `response_strings`. Each entry in the list provided with `response_strings` is a string fragment. You want something like:

```yaml
[...]
response_strings:
  - "can't remove a non-existent object 'what'"
[...]
```

If that doesn't work for some reason it is probably a wart with YAML quoting, in which case, try matching on a smaller string.

The `/foo/` syntax you're trying to use works in a few other places, but not in `response_strings` (since it is a list of fragments)

chungg:
> I wonder if the upgrade problem is not due to gnocchiclient recent issue?

yeah. could be related to https://github.com/gnocchixyz/python-gnocchiclient/pull/110? 

@tobias-urdin you can try forcing it to pip install gnocchiclient==7.0.6 to see if it's related to one of the PRs merged.

gnocchi/tests/functional/gabbits/resource-type.yaml
```diff
@@ -769,7 +769,7 @@ tests:
             max_length: 255
       status: 400
       response_strings:
-       - "can't remove non-existent object 'what'"
+       - can't remove non-existent object 'what'
```
cdent, line 772: The word `a` is missing from this message. See the test failure for the exact message, but I think it's supposed to be "can't remove _a_ non-existent..."

gnocchi/tests/functional/gabbits/resource-type.yaml
```diff
@@ -819,7 +819,7 @@ tests:
           path: /attributes/unknown
       status: 400
       response_strings:
-       - "can't remove non-existent object 'unknown'"
+       - can't remove non-existent object 'unknown'
```
cdent, line 822: same here:

The word `a` is missing from this message. See the test failure for the exact message, but I think it's supposed to be "can't remove _a_ non-existent..."

cdent (changes requested):

gnocchi/tests/functional/gabbits/resource-type.yaml
chungg: eagle eyes.

cdent: I'm a pterodactyl, caw caw.

jd: Says the _eye of lynx_.

http://eavesdrop.openstack.org/irclogs/%23openstack-telemetry/%23openstack-telemetry.2016-12-06.log.html#t2016-12-06T17:04:48

https://en.wiktionary.org/wiki/%C5%93il_de_lynx 😛

chungg: /me looks at date of chat log

how is this possible? i can't remember yesterday.

tobias-urdin: Wow, crazy how I could have missed that. Let's see if it works when fixing that. Thanks @cdent!

tobias-urdin:
Yay, tests passing :)

Adrian-Turjak:
Fantastic!

run-upgrade-tests.sh
```diff
@@ -40,7 +50,7 @@ inject_data() {
 
     {
         measures_sep=""
-        MEASURES=$(python -c 'import datetime, random, json; now = datetime.datetime.utcnow(); print(json.dumps([{"timestamp": (now - datetime.timedelta(seconds=i)).isoformat(), "value": random.uniform(-100000, 100000)} for i in range(0, 288000, 10)]))')
+        MEASURES=$(python -c 'import datetime, random, json; now = datetime.datetime.utcnow(); print(json.dumps([{"timestamp": (now - datetime.timedelta(seconds=i)).isoformat(), "value": random.uniform(-100000, 100000)} for i in range(0, 1000, 10)]))')
```
jd, line 53: The range was this big to be sure we tested file splitting IIRC. Might be good to keep it that high.

tox.ini
```diff
@@ -57,27 +58,28 @@ recreate = True
...
 
 [testenv:pep8]
+basepython = python3
 deps = hacking>=0.12,<0.13
 commands = flake8
 
-[testenv:py27-cover]
-commands = pifpaf -g GNOCCHI_INDEXER_URL run postgresql -- python setup.py testr --coverage --testr-args="{posargs}"
+[testenv:py37-cover]
+commands = pifpaf -g GNOCCHI_INDEXER_URL run postgresql -- python3 setup.py testr --coverage --testr-args="{posargs}"
```
jd, line 82: you shouldn't need python3 here

run-upgrade-tests.sh
```diff
@@ -1,6 +1,16 @@
 #!/bin/bash
 set -e
 
+if [ "$1" == "postgresql-file" ]; then
+  eval $(pifpaf --env-prefix INDEXER run postgresql)
+elif [ "$1" == "mysql-ceph" ]; then
+  eval $(pifpaf --env-prefix INDEXER run mysql)
```
jd, line 7: If you use `eval` here, nothing will stop mysql/postgresql for you.

You'll have to call `pifpaf_stop` IIRC to stop it before you finish this script or the process will kept hanging around.

That's why `pifpaf` was invoked from `tox`.

jd (changes requested):
Almost good. I think there are a few changes that might not be needed, like using `eval`, which will leave hanging processes once the tests are done.

run-upgrade-tests.sh
tobias-urdin: I will try adding it back see if Travis complains about output length again.

run-upgrade-tests.sh
tobias-urdin: Let me know if new change is better

run-upgrade-tests.sh
```diff
@@ -61,6 +71,8 @@ pifpaf_stop(){
 cleanup(){
     pifpaf_stop
```
jd, line 72: You have `pifpaf_stop` here not sure it works if `indexer_stop` works?
I mean there's too many stop begin called now I think :)

jd (changes requested):

run-upgrade-tests.sh
```diff
@@ -61,6 +71,8 @@ pifpaf_stop(){
...
     rm -rf $GNOCCHI_DATA
+    indexer_stop
+    [ -n $STORAGE_DAEMON ] && storage_stop
```
chungg, line 75: https://github.com/gnocchixyz/gnocchi/blob/master/setup.cfg#L94 seems to be always enabling ceph so STORAGE_DAEMON gets set. you may want to add a clause to ensure it's not file or to remove ceph from setup.cfg (although not sure what other things gets affected)

run-upgrade-tests.sh
tobias-urdin: I don't think so, the pifpaf_stop should still be the non-prefixed service running under pifpaf which is Gnocchi, while indexer_stop is, well, the indice database :)

run-upgrade-tests.sh
tobias-urdin: Ah, yeah. Thanks I'll try to do a check instead.

run-upgrade-tests.sh
```diff
@@ -61,6 +71,8 @@ pifpaf_stop(){
 cleanup(){
     pifpaf_stop
     rm -rf $GNOCCHI_DATA
+    indexer_stop
+    [ "$STORAGE_DAEMON" == "ceph" ] && storage_stop
```
chungg, line 75: just need to add `|| true` so it doesn't return non zero code.

tobias-urdin: Thanks!

tobias-urdin:
Tests are passing again, let me know if there is any additional feedback.

jd (approved):

Adrian-Turjak:
*cheers* TY Tobias!

I was gonna look at rebasing https://github.com/gnocchixyz/gnocchi/pull/995 but it seems you're already looking at it. :)

tobias-urdin:
@jd can we try to backport this to stable/4.3 shouldn't be any breaking changes

tobias-urdin:
i assume i can't send mergify commands

tobias-urdin:
@Mergifyio backport stable/4.3

jd:
@Mergifyio backport stable/4.3

jd:
@tobias-urdin you should do it manually :(

tobias-urdin:
ok, thanks :)

PR #1068, opened 2020-09-07
cradox is not supported anymore.

mrunge:
Update the docs etc. to use rados directly.

PR #1070, opened 2020-09-16
Use Bionic as dist in Travis

tobias-urdin:


chungg:
:/ seems to be hanging on testing against ceph.  might be something related to ceph+pifpaf on new image. i'll try the new image when free.

tobias-urdin:
Thanks @chungg I've been spending some time trying to understand why it's hanging but haven't figured it out yet.

tobias-urdin:
@chungg latest build is without cradox, stops earlier but still hangs

chungg:
yeah. i added a note in  #1068 but i think it's because python-rados is not a pypi package. we install it in test image via apt but i've no idea how to get tox to use that copy instead of trying to find it in pypi.

tobias-urdin:
dont need to install anything, as long as python3-rados is installed on your test system and cradox is not gnocchi should use that

chungg:
oh. didn't know that's how it works, thought the venv would ignore all packages.

i did notice in the ci image, that apt install puts python3-rados as a python3.6 package. could that be it?

tobias-urdin:
could be, but there is no package for 3.7, closest one is if we were to move to ubuntu focal where i think 3.8 i default so that would have python3-rados being 3.8 i think

tobias-urdin:
maybe try adding back a py36 job and see if that one works

tobias-urdin:
Atleast a little closer. It fails on: `gnocchi/tests/test_indexer.py::TestIndexerDriver::test_update_resource_metrics_append`

```
=================================== FAILURES ===================================
____________ TestIndexerDriver.test_update_resource_metrics_append _____________
'NoneType' object is not iterable
During handling of the above exception, another exception occurred:
NOTE: Incompatible Exception Representation, displaying natively:
testtools.testresult.real._StringException: Empty attachments:
  stderr
  stdout
logs: {{{
2020-09-25 00:55:11,175 [9701] INFO     gnocchi.common.ceph: Ceph storage backend use 'rados' python library
2020-09-25 00:55:11,190 [9701] ERROR    gnocchi.utils: Unable to initialize storage driver
  File "/home/tester/src/.tox/py36-postgresql/lib/python3.6/site-packages/tenacity/__init__.py", line 412, in call
    result = fn(*args, **kwargs)
  File "/home/tester/src/gnocchi/storage/__init__.py", line 102, in get_driver
    conf.storage)
  File "/home/tester/src/gnocchi/storage/ceph.py", line 52, in __init__
    self.rados, self.ioctx = ceph.create_rados_connection(conf)
  File "/home/tester/src/gnocchi/common/ceph.py", line 72, in create_rados_connection
    ioctx = conn.open_ioctx(conf.ceph_pool)
  File "rados.pyx", line 508, in rados.requires.wrapper.validate_func
  File "rados.pyx", line 1200, in rados.Rados.open_ioctx
rados.ObjectNotFound: [errno 2] error opening pool 'b'51da834fe78947d1a53004af51146efd''
```

tobias-urdin:
One step closer, the Ceph issue is fixed. The `rados mkpool` command failed with `(34) Numerical result out of range` but we never saw the error hence the pool didn't exist. The issue related to the amount of PG per OSD value when there is a large amount of pools, the issue is resolved by simply cleaning up the pools after ourselves in `tearDown()`.

The py36-postgresql job is passing, but the py36-mysql job fails because the returned data (the asserts) differs, haven't figured out why yet.

The patch:
- Changes testing from py37 -> py36 (because we need to import sitepackages to test with rados python bindings and not cradox)
- Cleanup Ceph pools after ourselves when testing
- Removes cradox for dependencies

Things left to do:
- Fix failure of py36-mysql job
- Figure out; should we remove cradox from deps in setup.cfg or change the import ordering in gnocchi/common/ceph.py to favor rados instead of cradox? (ie. what should we do to make this backportable)

Any help checking out failure for the py36-mysql job is appreciated, will continue when I have some time available.

tobias-urdin:
Seems like this was just transient issues, probably timing issues where gnocchi-metricd never had time to process it in time, and postgresql causing Tooz locking issues. @jd @chungg please have a look at let me know.

The only part I'm somewhat twisted about is the removal of cradox from setup.cfg deps, maybe we should instead adjust in gnocchi/common/ceph.py to prefer rados import first.

chungg:
if cradox works (in some cases), i agree with you, maybe it's best to still include it but change it to check for rados first

mrunge:
I have seen consistent crashes with cradox on python3.x

tobias-urdin:
Let's keep the dependency but prefer the rados import, then address the removal of cradox in a follow up, it's a mess with all the changes merged together into one huge PR that's just suppose to fix testing.

chungg (approved):

PR #1078, opened 2020-11-02
Add incoming/indexer/storage information to V1 API's status endpoint

unexceptable:
Rebase of https://github.com/gnocchixyz/gnocchi/pull/995 with fixed for pep8.

### DIFF
--- a/gnocchi/common/ceph.py
+++ b/gnocchi/common/ceph.py
@@ -12,7 +12,9 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+from collections import OrderedDict
 import errno
+import six
 
 import daiquiri
 
@@ -97,3 +99,22 @@ def errno_to_exception(ret):
             raise rados.Error("Unhandled error '%s'" % ret)
         else:
             raise getattr(rados, name)
+
+
+def get_ceph_health_status(driver):
+    """Return ceph status.
+
+    Include ceph stats.
+    """
+    response = OrderedDict([
+        ('name', driver.__class__.__name__)
+    ])
+    try:
+        stats = driver.rados.get_cluster_stats()
+    except Exception as e:
+        response['is_available'] = False
+        response['error'] = six.text_type(e)
+    else:
+        response['is_available'] = True
+        response['stats'] = stats
+    return response
--- a/gnocchi/common/file.py
+++ b/gnocchi/common/file.py
@@ -1,0 +1,24 @@
+# -*- coding:Utf-8 -*-
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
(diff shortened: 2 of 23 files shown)


Adrian-Turjak:
@chungg @tobias-urdin I've rebased this myself since it wasn't able to be cleanly done. Plus it looks like @leo-naeka might not be interested in carrying this anymore, so I might take over.

Adrian-Turjak:
the `gnocchi.tests.indexer.sqlalchemy.test_migrations.ModelsMigrationsSync.test_models_sync` failure is unrelated to this, and also failed on master.

So we should fix that and rebase this onto that. Meanwhile will also deal with the comments by @jd in the previous pull request.

Adrian-Turjak:
Getting tests to run locally is proving to be a bit of a pain, but `py37-postgresql-file` is passing locally, so I'm not sure what exactly us causing the 503 failures I'm seeing in travis-ci.

jd:
Some errors seems explicit:
```
    b'      "error": "\'FakeSwiftClient\' object has no attribute \'head_account\'"'
```

Adrian-Turjak:
> Some errors seem explicit:
> 
> ```
>     b'      "error": "\'FakeSwiftClient\' object has no attribute \'head_account\'"'
> ```

Good catch! I've added that now to the FakeSwiftClient, but for now only returning an empty dict. Will have to think about what will be a more appropriate default test return value.

Adrian-Turjak:
@jd Still have a few more of your comments from the other pull request to do, and need to add some more tests, but it's slowly starting to make sense to me. At least this code, and the project itself a little. 

Although gabbi confuses me a little because finding the specific test it ran isn't a simple grep/text search due to the test name in repo appearing differently to the name in the failure. :( It might be worth renaming all the tests with _ rather than spaces to make them easier to search, or I wonder if gabbi has a setting for more verbose test names in failures to make that clearer.

jd:
@cdent might know about Gabbi. Usually, I didn't find those hard to find with a `git grep` and replacing the underscore.

cdent:
> replacing the underscore

That's what I tend to do.

I agree that it's kind of a pain, and an extra step, but there's a conflict between trying to make the yaml files somewhat readable (they are supposed to be as much a thing for humans as for computers) and the demands of `unittest`, `stestr` and other parts of the puzzle.

Certainly if one wanted to rename the `name` values in the gabbi file to use `_` instead of space, that would be an option.

There's another complicating factor: Yesterday lbragstad noticed that when testtools was removed from gabbi, output and logging capture fixtures stopped working. That may make finding and dealing with failures a bit more messy. If that proves to be an issue, downgrading to `gabbi<2.0.0` will fix it until https://github.com/cdent/gabbi/issues/287 is resolved.

Adrian-Turjak:
@cdent I think for me gabbi is weird because I'm used to the python module.class.function path for tests, and gabbi fakes that. Maybe when a test fails it could also print out the relative file path location/test name? Or have a verbosity setting to enable that. 

I found my way around it ultimately, but just felt a little weird for someone who tends to stick with more standard python unit testing.

cdent:
> used to the python module.class.function path for tests

The code that launches the tests is responsible for setting the "faked" name, see https://gabbi.readthedocs.io/en/latest/gabbi.html#gabbi.driver.build_tests

From that it's usually possible to make the test name point to the yaml file, or at least very close.

tobias-urdin:
@mergifyio rebase

tobias-urdin:
@unexceptable Hello 👋 are you still working on this?

tobias-urdin:
Closing, feel free to reopen.

Issue #1083, opened 2020-11-05
Metric status does not work with Python3


paramite:
### Which version of Gnocchi are you using
# rpm -qa *gnocchi*
python3-gnocchiclient-7.0.4-0.20200310205157.64814b9.el8ost.noarch
gnocchi-common-4.3.5-0.20200310232918.8cbc9ee.el8ost.noarch
gnocchi-api-4.3.5-0.20200310232918.8cbc9ee.el8ost.noarch   
python3-gnocchi-4.3.5-0.20200310232918.8cbc9ee.el8ost.noarch

### How to reproduce your problem
Deploy Gnochi on Python 3.6.8 and run "openstack metric status" via python-3-gnocchi client.

### What is the result that you get
Gnocchi API throws following traceback:

[Tue Sep 29 13:39:24.857604 2020] [wsgi:error] [pid 25] [remote 10.120.21.71:60074] mod_wsgi (pid=25): Exception occurred processing WSGI script '/var/www/cgi-bin/gnocchi/app'.
[Tue Sep 29 13:39:24.863003 2020] [wsgi:error] [pid 25] [remote 10.120.21.71:60074] Traceback (most recent call last):
[Tue Sep 29 13:39:24.863161 2020] [wsgi:error] [pid 25] [remote 10.120.21.71:60074]   File "/usr/lib/python3.6/site-packages/webob/dec.py", line 129, in __call__
[Tue Sep 29 13:39:24.863168 2020] [wsgi:error] [pid 25] [remote 10.120.21.71:60074]     resp = self.call_func(req, *args, **kw)
[Tue Sep 29 13:39:24.863176 2020] [wsgi:error] [pid 25] [remote 10.120.21.71:60074]   File "/usr/lib/python3.6/site-packages/webob/dec.py", line 193, in call_func
[Tue Sep 29 13:39:24.863180 2020] [wsgi:error] [pid 25] [remote 10.120.21.71:60074]     return self.func(req, *args, **kwargs)
[Tue Sep 29 13:39:24.863187 2020] [wsgi:error] [pid 25] [remote 10.120.21.71:60074]   File "/usr/lib/python3.6/site-packages/oslo_middleware/base.py", line 130, in __call__
[Tue Sep 29 13:39:24.863191 2020] [wsgi:error] [pid 25] [remote 10.120.21.71:60074]     response = req.get_response(self.application)
[...]

PR #1084, opened 2020-11-05
Sanitize response in case error

paramite:
This patch makes Gnocchi app ensure bytes values are not used in response
which is going to be JSON serialized.

Fixes: Issue #1083

### DIFF
--- a/gnocchi/rest/app.py
+++ b/gnocchi/rest/app.py
@@ -136,6 +136,27 @@ class NotImplementedMiddleware(object):
                 "not implement this feature 😞")
 
 
+def sanitize(obj):
+    """Avoids using bytes in given object."""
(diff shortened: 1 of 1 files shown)


chungg:
can we fix the specific request? https://github.com/gnocchixyz/gnocchi/blob/bd2cafadd0961231748cf4b67efd28c6db277b04/gnocchi/rest/api.py#L2117

my only concern is i believe this renderer is used for all requests and i have some (possibly unwarranted concern) regarding recursively fixing all payloads.

paramite:
Oh, here it is. Yes, we can definitely fix that on the side, where data is generated. I just did not have much time for the fix, so came up with this quick and dirty one. Will update this PR. Thanks for the pointer!

PR #1090, opened 2020-12-04
Set json template for DELETE requests

tobias-urdin:
We hardcode the content-type on all DELETE requests
otherwise Pecan will throw NotAcceptable errors when
doing requests to the APIs without any content-type set
and will default to text/html.

### DIFF
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -279,7 +279,7 @@ class ArchivePolicyController(rest.RestController):
         except indexer.UnsupportedArchivePolicyChange as e:
             abort(400, six.text_type(e))
 
-    @pecan.expose()
+    @pecan.expose('json')
     def delete(self):
         # NOTE(jd) I don't think there's any point in fetching and passing the
         # archive policy here, as the rule is probably checking the actual role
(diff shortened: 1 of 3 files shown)


tobias-urdin:
This fixes python-gnocchiclient issues.

jd (commented):
LGTM that should be tested though :/

tobias-urdin:
Setting template to expose() to json is cleaner than hardcoding application/json (pecan will set that content type if template is json). I don't know any good way to test it since it doesn't change the actual response but only the internal logic that Pecan doesn't error out when it doesn't know the content-type is should use (since we are setting guess_content_type_from_ext=False in pecan config).

jd (approved):
That could be tested with Gabbi?

tobias-urdin:
Let me check.

tobias-urdin:
@jd something like this? I verified the tests passes before the expose() change.

jd:
Yes, something like that :)

tobias-urdin:
Adding oslo.policy pin here, might be the one causing postgresql job to have issues.

tobias-urdin:

PR #1092, opened 2020-12-07
Fix misspelled asserts

tobias-urdin:


jd (approved):

tobias-urdin:
Think this should have gone in stable branch since it's only for an older version of gabbi, will try again when I have more time to fix stable CI.

PR #1093, opened 2020-12-07
Set GNOCCHI_TEST_DEBUG

tobias-urdin:
Test enabling debug logging for jobs
to see if we can make the Travis CI
limit.


tobias-urdin:
The output is still to long, in the future enable stdout for jobs but skip DEBUG logs.

PR #1097, opened 2020-12-10
Revert "Pin oslo.policy<=3.4.0"

mrunge:
It broke e.g ceilometer/aodh on devstack.
This reverts commit a2d84f5b1b0555ad83b14e931d9bd7065f2f7b6a.

I acknowledge, this is controversial, see 
https://github.com/gnocchixyz/gnocchi/commit/a2d84f5b1b0555ad83b14e931d9bd7065f2f7b6a#comments
for discussion

Resolves: https://github.com/gnocchixyz/gnocchi/issues/1095

PR #1099, opened 2020-12-11
Fix yaml deprecation warning

mrunge:
YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated,
as the default Loader is unsafe. Please read https://msg.pyyaml.org/load
for full details.

### DIFF
--- a/gnocchi/gendoc.py
+++ b/gnocchi/gendoc.py
@@ -194,7 +194,7 @@ def setup(app):
 
     # TODO(jd) Do not hardcode doc/source
     with open("doc/source/rest.yaml") as f:
-        scenarios = ScenarioList(yaml.load(f))
+        scenarios = ScenarioList(yaml.load(f, Loader=yaml.FullLoader))
 
     test = test_rest.RestTest()
     test.auth_mode = "basic"
--- a/setup.cfg
+++ b/setup.cfg
@@ -87,6 +87,7 @@ prometheus =
 amqp1:
     python-qpid-proton>=0.17.0
 doc =
+    chardet<4
     sphinx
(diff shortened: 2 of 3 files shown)


jd (approved):

gnocchi/gendoc.py
```diff
@@ -194,7 +194,7 @@ def setup(app):
 
     # TODO(jd) Do not hardcode doc/source
     with open("doc/source/rest.yaml") as f:
-        scenarios = ScenarioList(yaml.load(f))
+        scenarios = ScenarioList(yaml.load(f), Loader=yaml.FullLoader)
```
jd, line 197: syntax error I think

jd (changes requested):

gnocchi/gendoc.py
mrunge: embarrassing. Yes, of course!

jd (approved):

mrunge:
Thank you for looking at this Julien!

mrunge:
https://github.com/gnocchixyz/gnocchi/pull/1097 also contains a commit to pin chardet<4, which seems to be necessary here as well? Let's see if the other patch passes.

mrunge:
It seems pinning chardet is necessary here for the time being.

@jd may I ask for a review again?

jd:
@mergifyio refresh

mrunge:
Thank you.

tobias-urdin:
@mergifyio backport-to-4.3

tobias-urdin:
@Mergifyio backport stable/4.3

PR #1100, opened 2020-12-15
Python 3.7 is removed

mrunge:
from tox, it should also get removed from docs.


jd (approved):

mrunge:
I have seen https://github.com/gnocchixyz/gnocchi/pull/1097 fail with a comparable(?) error message. I wonder if there is a different issue in the CI system.

mrunge:
ugh, this was a bit quick. I did not assume this was merged with any additional change.

PR #1108, opened 2020-12-29
Use yaml for policy file by default

kajinamit:
In OpenStack projects we are deprecating usage of json for policy file
and will use yaml by default[1]. This change introduces the default
yaml file and replace the default of policy_file to follow that
migration because Gnocchi depends on the same oslo.policy library.

The old json file is kept because there are several distros like RDO
which uses the file, but it will be removed in a future release.

[1] https://governance.openstack.org/tc/goals/selected/wallaby/migrate-policy-format-from-json-to-yaml.html


jd (changes requested):
Can you add a release note?

kajinamit:
@jd Thanks for your feedback. I added a release note to explain the change as requested.

jd (approved):

PR #1109, opened 2021-01-04
Implement default policies in code

kajinamit:
This change implements default policies in code. With this change, operators no longer need to define the default policy rules in policy configuration file(policy.yaml) unless they need any customized rules.

### DIFF
--- a/gnocchi/rest/policies.py
+++ b/gnocchi/rest/policies.py
@@ -1,0 +1,48 @@
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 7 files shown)


gnocchi/gnochi-policy-generator.conf
```diff
@@ -0,0 +1,2 @@
+[DEFAULT]
```
jd, line 1: typo in filename

jd (changes requested):

gnocchi/rest/policies.py
```diff
@@ -0,0 +1,413 @@
...
+        description='List all metrics',
+        operations=[
+            {
+                'path': '/v1/metric',
+                'method': 'GET'
+            }
+        ]
+    ),
+    policy.DocumentedRuleDefault(
+        name="Search metric",
```
kajinamit, line 363: This should be "search metric"

gnocchi/gnochi-policy-generator.conf
kajinamit: Done

gnocchi/rest/policies.py
kajinamit: Done

jd (approved):
LGTM
I didn't check all the listing so I'll trust you on this.

PR #1112, opened 2021-01-11
Re-enable tests using swift and s3 backend

mrunge:
The tests were disabled in a previous commit, but the goal is to have more and more stable tests.


mrunge:
Fails due to docker limit
```
toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

PR #1116, opened 2021-01-20
Dismiss reviews after pushing a new commit

mrunge:
to prevent accidental merges of non-reviewed
changes.

Related: https://github.com/Mergifyio/mergify-engine/issues/1860

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -23,8 +23,7 @@ pull_request_rules:
   - name: dismiss reviews on additional commit
     actions:
       dismiss_reviews: {}
-    contitions:
-    - .*
+    conditions: []
 
 # Backports to stable branches
   - actions:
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -20,6 +20,11 @@ pull_request_rules:
     - label!=work-in-progress
     - author=mergify[bot]
     - status-success=Travis CI - Pull Request
+  - name: dismiss reviews on additional commit
+    actions:
+      dismiss_reviews: {}
+    contitions:
+    - .*
```
jd, line 27:
```suggestion
    conditions: []
```

jd (changes requested):

.mergify.yml
mrunge: Thanks, will do that

mrunge:
Shouldn't the update also dismiss the review?

jd:
It will, once this is merged.

mrunge:
Okay, thank you Julien!

PR #1118, opened 2021-02-02
Fix up merge conflicts 

mrunge:
This should fix up merge issues with https://github.com/gnocchixyz/gnocchi/pull/1107

Also gnocchi/amqp1d.py is a feature from 4.4 branch


mrunge:
so, this failed twice, once during mysql tests, the other time during postgresql tests.

PR #1119, opened 2021-02-17
Catch Redis server errors

mrunge:
There seems to happen a bad interaction with Redis over
HAProxy, and Redis connections are reported as closed,
which is only temporary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,10 @@ env:
 
   - TARGET: py36-mysql-file
   - TARGET: py36-mysql-swift
-  - TARGET: py36-mysql-ceph
+  - TARGET: py36-mysql-s3
   - TARGET: py36-postgresql-file
   - TARGET: py36-postgresql-swift
-  - TARGET: py36-postgresql-ceph
+  - TARGET: py36-postgresql-s3
 
 before_script:
   # NOTE(sileht): We need to fetch all tags/branches for documentation.
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,10 @@ env:
 
   - TARGET: py36-mysql-file
   - TARGET: py36-mysql-swift
-  - TARGET: py36-mysql-s3
+  - TARGET: py36-mysql-ceph
   - TARGET: py36-postgresql-file
   - TARGET: py36-postgresql-swift
-  - TARGET: py36-postgresql-s3
+  - TARGET: py36-postgresql-ceph
 
 before_script:
   # NOTE(sileht): We need to fetch all tags/branches for documentation.
--- a/gnocchi/cli/metricd.py
+++ b/gnocchi/cli/metricd.py
@@ -86,7 +86,7 @@ class MetricProcessBase(cotyledon.Service):
             with utils.StopWatch() as timer:
                 try:
                     self._run_job()
-                except ConnectionError as ce:
+                except ConnectionError:
                     LOG.debug("Redis server closed connection. Retrying.")
                 except Exception:
                     LOG.error("Unexpected error during %s job",
@@ -195,7 +195,7 @@ class MetricProcessor(MetricProcessBase):
                     self.wakeup()
         except exceptions.NotImplementedError:
             LOG.info("Incoming driver does not support notification")
-        except ConnectionError as ce:
+        except ConnectionError:
             LOG.debug("Redis server closed connection. Retrying.")
         except Exception as e:
             LOG.error(
@@ -246,7 +246,7 @@ class MetricProcessor(MetricProcessBase):
                 s_count += 1
                 self.incoming.finish_sack_processing(s)
                 self.sacks_with_measures_to_process.discard(s)
-            except ConnectionError as ce:
+            except ConnectionError:
                 LOG.debug("Redis server closed connection. Retrying.")
             except Exception:
                 LOG.error("Unexpected error processing assigned job",
--- a/gnocchi/cli/metricd.py
+++ b/gnocchi/cli/metricd.py
@@ -23,6 +23,7 @@ import cotyledon
 from cotyledon import oslo_config_glue
 import daiquiri
 from oslo_config import cfg
+from redis.exceptions import ConnectionError
 import tenacity
 import tooz
 from tooz import coordination
@@ -85,6 +86,9 @@ class MetricProcessBase(cotyledon.Service):
             with utils.StopWatch() as timer:
                 try:
                     self._run_job()
+                except ConnectionError as ce:
+                    LOG.debug("Redis server closed connection. Retrying.")
+                    raise tenacity.TryAgain(ce)
                 except Exception:
                     LOG.error("Unexpected error during %s job",
                               self.name,
@@ -192,6 +196,9 @@ class MetricProcessor(MetricProcessBase):
                     self.wakeup()
         except exceptions.NotImplementedError:
             LOG.info("Incoming driver does not support notification")
+        except ConnectionError as ce:
+            LOG.debug("Redis server closed connection. Retrying.")
+            raise tenacity.TryAgain(ce)
         except Exception as e:
             LOG.error(
                 "Error while listening for new measures notification, "
@@ -240,6 +247,9 @@ class MetricProcessor(MetricProcessBase):
                 s_count += 1
                 self.incoming.finish_sack_processing(s)
                 self.sacks_with_measures_to_process.discard(s)
+            except ConnectionError as ce:
+                LOG.debug("Redis server closed connection. Retrying.")
+                raise tenacity.TryAgain(ce)
             except Exception:
                 LOG.error("Unexpected error processing assigned job",
                           exc_info=True)
--- a/gnocchi/cli/metricd.py
+++ b/gnocchi/cli/metricd.py
@@ -240,7 +240,6 @@ class MetricProcessor(MetricProcessBase):
                 s_count += 1
                 self.incoming.finish_sack_processing(s)
                 self.sacks_with_measures_to_process.discard(s)
-
             except Exception:
                 LOG.error("Unexpected error processing assigned job",
                           exc_info=True)
--- a/gnocchi/cli/metricd.py
+++ b/gnocchi/cli/metricd.py
@@ -88,7 +88,6 @@ class MetricProcessBase(cotyledon.Service):
                     self._run_job()
                 except ConnectionError as ce:
                     LOG.debug("Redis server closed connection. Retrying.")
-                    raise tenacity.TryAgain(ce)
                 except Exception:
                     LOG.error("Unexpected error during %s job",
                               self.name,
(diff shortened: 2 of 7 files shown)


gnocchi/cli/metricd.py
```diff
@@ -240,6 +247,9 @@ def _run_job(self):
                 s_count += 1
                 self.incoming.finish_sack_processing(s)
                 self.sacks_with_measures_to_process.discard(s)
+            except ConnectionError as ce:
+                LOG.debug("Redis server closed connection. Retrying.")
+                raise tenacity.TryAgain(ce)
```
jd, line 252: The function is not directly decorated with tenacity so I'd avoid raising this exception here.
Maybe it'd be better to decorate this with a tenacity decorator?

gnocchi/cli/metricd.py
```diff
@@ -192,6 +196,9 @@ def _fill_sacks_to_process(self):
                     self.wakeup()
         except exceptions.NotImplementedError:
             LOG.info("Incoming driver does not support notification")
+        except ConnectionError as ce:
+            LOG.debug("Redis server closed connection. Retrying.")
+            raise tenacity.TryAgain(ce)
```
jd, line 201: same as below

gnocchi/cli/metricd.py
```diff
@@ -85,6 +86,9 @@ def run(self):
             with utils.StopWatch() as timer:
                 try:
                     self._run_job()
+                except ConnectionError as ce:
+                    LOG.debug("Redis server closed connection. Retrying.")
+                    raise tenacity.TryAgain(ce)
```
jd, line 91: You want to ignore the exception then, the loop will retry by itself. No need to use tenacity.

jd (changes requested):

gnocchi/cli/metricd.py
mrunge: done (in the following patch)

gnocchi/cli/metricd.py
mrunge: done

gnocchi/cli/metricd.py
mrunge: done

mrunge:
Thank you for the review. The follow-up patch should address the review comments.

gnocchi/cli/metricd.py
```diff
@@ -85,6 +86,8 @@ def run(self):
             with utils.StopWatch() as timer:
                 try:
                     self._run_job()
+                except ConnectionError:
```
jd, line 89: Ok I see another problem here, is that you are violating separation of concern.
The metricd tool does not need to know about Redis, and anything else can be used. That means the exception must be managed in the redis driver itself, not here.

gnocchi/cli/metricd.py
```diff
@@ -192,6 +195,8 @@ def _fill_sacks_to_process(self):
                     self.wakeup()
         except exceptions.NotImplementedError:
             LOG.info("Incoming driver does not support notification")
+        except ConnectionError:
```
jd, line 198: same as above

gnocchi/cli/metricd.py
```diff
@@ -219,6 +224,7 @@ def _get_sacks_to_process(self):
         finally:
             return self._tasks or self.fallback_tasks
 
+    @utils.retry_on_exception.wraps
```
jd, line 227: That's unrelated to redis, it should probably not be there; it'll retry on ANY exception.

setup.cfg
```diff
@@ -34,6 +34,7 @@ install_requires =
     futures; python_version < '3'
     jsonpatch
     cotyledon>=1.5.0
+    redis >= 3.2.0 # MIT
```
jd, line 37: Redis is not mandatory to run gnocchi, do not make that change.

jd (dismissed):
Sorry it's been a long time since I did Gnocchi stuff so my brain was a bit slow the first review

gnocchi/cli/metricd.py
mrunge: underlying incoming/redis does not handle any exceptions, but it probably should. There has been a change in redis-py around 3.0, which hinted/suggested to catch and handle exceptions in applications and not in redis-py. I can not find the issue/pr in redis-py at the moment.

Where would you want the exception handled then?

jd: In such files https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/incoming/redis.py and its method.
I'm not sure which part of Redis raises `ConnectionError`, if it's all of the, you might need to decorate with a tenacity decorator to retry a few times the operations, and bail out if it fails too much.

setup.cfg
mrunge: I agree. Moving the change to the incoming redis driver should not make it a requirement here. I've moved it back to the redis section.

.travis.yml
```diff
@@ -18,10 +18,10 @@ env:
 
   - TARGET: py36-mysql-file
   - TARGET: py36-mysql-swift
-  - TARGET: py36-mysql-s3
+  - TARGET: py36-mysql-ceph
```
jd, line 21: This looks unrelated to this PR :(

gnocchi/incoming/redis.py
```diff
@@ -93,6 +96,10 @@ def add_measures_batch(self, metrics_and_measures):
                 notified_sacks.add(sack_name)
         pipe.execute()
 
+    @tenacity.retry(
+        wait=utils.wait_exponential,
+        # Never retry except when explicitly asked by raising TryAgain
+        retry=tenacity.retry_never)
```
jd, line 102: I'd go with:
```suggestion
    @tenacity.retry(
        wait=utils.wait_exponential,
        retry=tenacity.retry_if_exception_type(ConnectionError))
```
That should be simpler.
You probably want to stop at some point too though, I'd add a condition for it to stop retrying after a few tries.

jd (changes requested):

.travis.yml
mrunge: yes indeed it is.
Ideally, I would like to have all backends enabled. There seems to be a timing(?) issue with CI, and lately, the s3 tests were failing with a chance of > 50%

gnocchi/incoming/redis.py
mrunge: That looks better. Thank you for the suggestion.

I've never seen this repeated more than twice for now.

gnocchi/incoming/redis.py
```diff
@@ -106,20 +113,23 @@ def update_report(results, m_list):
...
+            else:
                 results = pipe.execute()
                 update_report(results, m_list)
-                m_list = []
-                pipe = self._client.pipeline()
-        else:
-            results = pipe.execute()
-            update_report(results, m_list)
+        except ConnectionError:
+            LOG.debug("Redis Server closed connection. Retrying.")
```
jd, line 132: you don't need this anymore (the try/except)

gnocchi/incoming/redis.py
```diff
@@ -176,18 +186,25 @@ def process_measures_for_sack(self, sack):
...
-                # Format is defined by _get_sack_name: incoming128-17
-                yield self._make_sack(int(message['channel'].split(b"-")[-1]))
+        try:
+            for message in p.listen():
+                if message['type'] == 'pmessage' and message['pattern'] == pattern:
+                    # FIXME(jd) This is awful, we need a better way to extract this
+                    # Format is defined by _get_sack_name: incoming128-17
+                    yield self._make_sack(int(message['channel'].split(b"-")[-1]))
+        except ConnectionError:
+            LOG.debug("Redis Server closed connection. Retrying.")
```
jd, line 207: ditto

jd (changes requested):
you still miss a stop condition right?

gnocchi/incoming/redis.py
```diff
@@ -176,6 +183,10 @@ def process_measures_for_sack(self, sack):
             pipe.ltrim(key, item_len + 1, -1)
         pipe.execute()
 
+    @tenacity.retry(
+        wait=utils.wait_exponential,
+        # Never retry except when explicitly asked by raising TryAgain
```
jd, line 188: wrong comment

gnocchi/incoming/redis.py
```diff
@@ -176,6 +183,10 @@ def process_measures_for_sack(self, sack):
...
+        retry=tenacity.retry_if_exception_type(ConnectionError))
```
jd, line 189: missing stop condition?

jd (dismissed):

jd:
@mrunge LGTM
Are you able to test it on your setup to see if it fixes your issue?

mrunge:
Thank you for your reviews!

Yes, I've tested this a couple of times already over each iteration. That's why I had debug logs etc and wasn't that much concerned about stop condition.

This latest proposal does not fix the issue, try/except is required, apparently.

jd:
> This latest proposal does not fix the issue, try/except is required, apparently.

That does not make sense. What's the traceback?

mrunge:
Sorry for not getting back earlier. While I got stack traces on Friday, The deployment is now running flawlessly for over a hour, no stack traces. Before, these kind of traces errors happened every 2 to 5 minutes. We should consider this PR to fix the issues.

mrunge:
Just after my last comment, I saw this trace in the logs
```
2021-02-22 08:09:20,432 [32] ERROR    gnocchi.cli.metricd: Unexpected error during reporting job
  File "/usr/lib/python3.6/site-packages/gnocchi/cli/metricd.py", line 87, in run
    self._run_job()
  File "/usr/lib/python3.6/site-packages/gnocchi/cli/metricd.py", line 127, in _run_job
    report = self.incoming.measures_report(details=False)
  File "/usr/lib/python3.6/site-packages/gnocchi/incoming/__init__.py", line 211, in measures_report
    metrics, measures, full_details = self._build_report(details)
TypeError: 'NoneType' object is not iterable
```

mrunge:
I went back to my edits in the test environment and found that I had a try/except left over.
After removal and making sure it looks like the same as in this PR, the exceptions are completely gone from from logs. The environment is running for 2 hours now, where I would have expected

mrunge:
It turned out I had a try/except left over in my test deployment where I also tried this out. It's running for quite a while now in the form of being proposed here. Sorry to cause confusion here.

Thank you for your effort here. With this PR merged being merged, I see the issue being fixed then.

jd:
Your traceback makes little sense as the `_build_report` function never returns `None`, and if the retry fails it should raise a `RetryError`. 🤔 
You sure you got this right?

mrunge:
I had wrapped https://github.com/gnocchixyz/gnocchi/blob/eb871885c8e681d2bc91e59bc115d2c4e55ec8c3/gnocchi/incoming/redis.py#L186-L190 with a try/except, which then could lead to returning none.

```
 try:
        for message in p.listen():
            if message['type'] == 'pmessage' and message['pattern'] == pattern:
                # FIXME(jd) This is awful, we need a better way to extract this
                # Format is defined by _get_sack_name: incoming128-17
                yield self._make_sack(int(message['channel'].split(b"-")[-1]))
  except ....:
      LOG.debug.... 
```

mrunge:
I understand it that way, that my additional changes/left overs were causing the exception reported in https://github.com/gnocchixyz/gnocchi/pull/1119#issuecomment-783180946.

jd:
Oh ok, I got confused, I thought this PR was doing that. Ok so it should be fine!

jd (approved):

mrunge:
thank you @jd

PR #1123, opened 2021-03-15
Use python3 in travis deploy

tobias-urdin:


.travis.yml
```diff
@@ -70,6 +70,9 @@ notifications:
       - "irc.freenode.org#gnocchi"
 
 before_deploy:
+  - python3 -m virtualenv ~/deploy-env
+  - source ~/deploy-env/bin/activate
+  - python --version
   - pip install --user --upgrade pip
   - pip install --user --upgrade six
```
jd, line 77: just use
```suggestion
  - python3 -m pip install --user --upgrade pip
  - python3 -m pip install --user --upgrade six
```
no?

tobias-urdin: This is to ensure the pypi dpl is executed with python3, see https://github.com/travis-ci/dpl/blob/master/lib/dpl/providers/pypi.rb

since it's in a venv it should catch that pip is in python3 already

mrunge:
I assume the tests fail on tenacity 7.0.0, see https://github.com/gnocchixyz/gnocchi/pull/1124/commits/485610201930e419ca5902a6e37d4b55049d3f1e

PR #1124, opened 2021-03-17
Policy in code requires oslo.policy >= 3.5.0

mrunge:
and thus we should bump the requirement.

PR #1125, opened 2021-03-18
Limit maximum metrics to be deleted simultaniously

felixhuettner:
Currently gnocchi-metricd tries to collect all metrics that should be deleted
and deletes them in one big loop. If a large amount of old metrics has
accumulated this will cause high load on the database as it needs to find
and sort all of these rows. It will also cause a high amount of memory usage
for gnocchi-metricd as it needs to also sort all of these rows in memory.

To fix this the amount of rows being deleted in one run is limited to 10000.
This is sufficient to allow gnocchi to cleanup metrics in a fast way under
normal circumstances while still allowing it to catch up if no deletion has
been performed for a while.

Signed-off-by: Felix Huettner <felix.huettner@mail.schwarz>

### DIFF
--- a/gnocchi/opts.py
+++ b/gnocchi/opts.py
@@ -172,6 +172,11 @@ def list_opts():
                        "value may improve worker utilization but may also "
                        "increase load on coordination backend. Value is "
                        "capped by number of workers globally."),
+            cfg.IntOpt('cleanup_batch_size',
+                       default=10000,
+                       min=1,
+                       help="Number of metrics that should be deleted "
+                       "simultaneously by one janitor."),
         )),
         ("api", (
             cfg.StrOpt('paste_config',
--- a/releasenotes/notes/configurable_clenaup_batch_size-a785c1aa6ee29058.yaml
+++ b/releasenotes/notes/configurable_clenaup_batch_size-a785c1aa6ee29058.yaml
@@ -1,0 +1,12 @@
+---
+
+features:
+  - |
+    Users can now configure the ``cleanup_batch_size``. This limits the amount of
+    metrics being deleted during a run of the janitor thereby preventing
+    out-of-memory errors.
+upgrade:
+  - |
+    Users having a large amount of metrics being deleted regularly might needs
+    to increase ``cleanup_batch_size`` to a higher value so that the janitor
+    can keep up.
--- a/gnocchi/chef.py
+++ b/gnocchi/chef.py
@@ -49,14 +49,17 @@ class Chef(object):
         self.index = index
         self.storage = storage
 
-    def expunge_metrics(self, sync=False):
+    def expunge_metrics(self, cleanup_batch_size, sync=False):
         """Remove deleted metrics.
 
+        :param cleanup_batch_size: The amount of metrics to delete in one
+                                   run.
         :param sync: If True, then delete everything synchronously and raise
                      on error
         :type sync: bool
         """
-        metrics_to_expunge = self.index.list_metrics(status='delete')
+        metrics_to_expunge = self.index.list_metrics(status='delete',
+                                                     limit=cleanup_batch_size)
         metrics_by_id = {m.id: m for m in metrics_to_expunge}
         for sack, metric_ids in self.incoming.group_metrics_by_sack(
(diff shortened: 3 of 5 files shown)


felixhuettner:
Hi everyone,
i'm not sure about the CI Runs but the issues do not seem to be related to the PR in any way

jd (dismissed):
Can we make this configurable?

felixhuettner:
Implemented a configuration option. Not sure about the name though

gnocchi/opts.py
```diff
@@ -172,6 +172,11 @@ def list_opts():
                        "value may improve worker utilization but may also "
                        "increase load on coordination backend. Value is "
                        "capped by number of workers globally."),
+            cfg.IntOpt('janitor_batch_size',
+                       default=10000,
+                       min=1,
+                       help="Number of metrics that should be deleted "
+                       "simultaniously by one janitor."),
```
jd, line 179:
```suggestion
                       "simultaneously by one janitor."),
```

gnocchi/opts.py
```diff
@@ -172,6 +172,11 @@ def list_opts():
                        "value may improve worker utilization but may also "
                        "increase load on coordination backend. Value is "
                        "capped by number of workers globally."),
+            cfg.IntOpt('janitor_batch_size',
```
jd, line 175:
```suggestion
            cfg.IntOpt('cleanup_batch_size',
```
there's `metric_cleanup_delay` already

jd (dismissed):

gnocchi/opts.py
felixhuettner: fixed

gnocchi/opts.py
felixhuettner: fixed

jd (commented):
Thank you!
IIRC the doc is autogenerated. Maybe adding a release note with `reno` would be cool? :)

felixhuettner:
I hope i got the release note format correctly :)

releasenotes/notes/configurable_clenaup_batch_size-a785c1aa6ee29058.yaml
```diff
@@ -0,0 +1,12 @@
+---
+
+features:
+  - |
+    Users can now configure the ``cleanup_batch_size``. This limits the amount of
+    metrics being deleted during a run of the janitor thereby preventing
+    Out-of-memory errors.
```
jd, line 7:
```suggestion
    out-of-memory errors.
```

jd (approved):
Yeah this is great! Thank you!

jd:
The CI seems broken of the latest tenacity, somebody should rewrite the current code to stop using the deprecated now-removed API.

felixhuettner:
What is then the best way to go forward here? Should we just leave it sitting here until the CI is fixed (that would be fine with me)

jd (changes requested):
Yes, feel free to give a hand to fix the CI. Not sure anybody is working on it 😢

felixhuettner:
Would you have any details where i could start taking a look? Then i'll see if i can get some time for that

jd:
The latest tenacity release https://pypi.org/project/tenacity/ 7.0.0 removed old compat code for retry functions.
It seems Gnocchi is providing its own retry function and they need to be rewritten for the new API. That should not be hard.

mrunge:
I've put together a PR to fix CI by pinning tenacity and sqlalchemy, https://github.com/gnocchixyz/gnocchi/pull/1124
There is also a bump for oslo.policy included. This patch currently fails in upgrade tests; I've also submitted the same for the previous branch, (which fails in ubgrade checks from 4.2...)

tobias-urdin:
@Mergifyio rebase

jd (approved):

jd:
@mergifyio update

felixhuettner:
:D ok that is fixed now.
Let me try this
@Mergifyio update

felixhuettner:
:D ok, let's do it manually

jd (approved):

tobias-urdin:
@Mergifyio rebase

PR #1129, opened 2021-03-25
Run upgrade testing from git

tobias-urdin:
Don't upgrade between released versions but use tags
or branches instead. When dependencies breaks it's
easier to get jobs passing without having to release
instantly.


mrunge (approved):
Thank you, looks good to me.

tobias-urdin:
@Mergifyio backport stable/4.4 stable/4.3

PR #1135, opened 2021-03-26
Drop upgrade jobs from 4.2

tobias-urdin:
We won't maintain CI for 4.2 so we won't
test upgrade.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,6 @@ env:
   - TARGET: pep8
   - TARGET: docs
 
-  - TARGET: py36-mysql-ceph-upgrade-from-4.2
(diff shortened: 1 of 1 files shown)


jd:
Well if you maintain 4.3 you still want the upgrade to be tested, no?

tobias-urdin:
Ideally yes, if anybody wan't to fix 4.2 branch that would be good but it's missing so many fixes that has went in on other places that I can't even keep track. The only good thing regarding this is that most people hopefully doesn't run 4.2 still even though some still do.

tobias-urdin:
So it's just blocking the 4.3 branch for now, but yeah, maybe I will have some time to look at it.

jd (approved):
Fine with me then!

PR #1136, opened 2021-03-29
Bump ujson>=2.0.3

mrunge:
There is a memory leak included in ujson, see
https://github.com/ultrajson/ultrajson/pull/394/commits/92c57b421052b146e4f2dfce9386c7cf13241091

PR #1137, opened 2021-03-29
Pin importlib-metadata<3.9.0

tobias-urdin:


tobias-urdin:
Is this the issue?

tobias-urdin:
@Mergifyio backport stable/4.3 stable/4.4

mrunge:
What is the issue here?

tobias-urdin:
check job result on your patch https://github.com/gnocchixyz/gnocchi/pull/1136 - new release of import-metadata that seems to break stevedore

mrunge:
ah, got you. Thanks!

mrunge (approved):

mrunge:
Taking a step back: gnocchi does not use importlib-metadata. Not sure if we should declare it then here.

tobias-urdin:
do you have any alternative fix? we can pin stevedore to <3.0.0 but that would conflict with ==3.3.0 upper-constraint for victoria and wallaby release of openstack

tobias-urdin:
maybe block it in tox.ini envs deps instead

tobias-urdin:
fixed in >=3.9.1

PR #1157, opened 2021-07-01
Prepare for s3rver version bump

tobias-urdin:
Changes required when s3rver is bumped in ci-tools.


tobias-urdin:
will be handled another way

PR #1159, opened 2021-07-04
Migrate CI to Github Actions

tobias-urdin:
This PR's is to move away from Travis CI since it's not credit based and we run out of credits after just a few CI runs.
However this also includes a lot of other fixes and changes that benefits the stability and testing of Gnocchi.

The migration plan to move to Github Actions from Travis CI is to (that this PR partly does) continue using Docker but for now build on each job run until we change DockerHub to read Dockerfile from main repo. We can then add "container" directly to our jobs in the workflow and use a hosted container image, and thus retire the ci-tools repo.

This PR will:

- Move Dockerfile from ci-tools repo to main repo (inside our gnocchi-container GitHub action)
- Updates s3rver version to 3.7.0 and uses default credentials which hopefully stabilizes the S3 jobs - hence the update to the s3rver default credentials (username and password is S3RVER)
- Migrate Docker container to Ubuntu 20.04 (defaults to Python 3.8 but we also run 3.6)
- Because of above this means we are testing newer version of Ceph, Redis etc that is installed from distro packages
- Add testing for Python 3.8 but keeps Python 3.6 testing
- Adds back Ceph testing but only for Python 3.8 - this is because the rados module (python3-rados dist package) that we import from global site-packages is only for Python 3.8, we can build a Ubuntu 18.04 image with python3-rados for Python 3.6 if we want to test that but can be a follow up.
- This fixes so that the deploy pipeline (that is the only thing left) in Travis CI uses Python 3
- Finally this contains some minor fixes to documentation, old comments, deprecation warnings and things to make it easier to run locally.

This PR has a hard dependency on https://github.com/jd/pifpaf/pull/150 to make Ceph working.

There are some follow up patches that can be done after this which is not covered in this patch.

- When DockerHub container image builds is updated we can remove our custom GitHub gnocchi-container action and instead use the "container" directive directly in the workflow which will pull the image from DockerHub instead of rebuilding on each job run
- Add Python 3.6 Ceph jobs again
- Move away the deploy pipeline from Travis CI into a GitHub actions that is executed when a new release is created


tobias-urdin:
@jd see above, all commits should have appropriate comments as well. I tested the workflow on my fork so should be working, but you might need to allow the workflow inside the gnocchi project settings (I don't have permissions I think).

tobias-urdin:
I should also mention that it only migrates the master branch to Github Actions, the stable branches needs to be migrated in a follow up.

openstacker:
Nice work. Thanks.

jd:
GitHub Actions is enabled on this project.

jd:
@tobias-urdin That looks good to me, I'm ok to merge it. You can still fix it as you go.

jd (approved):

chungg (approved):

PR #1160, opened 2021-07-07
ci: dont run actions on push and dont hardcode default branch

tobias-urdin:
Only run the actions on PR and not
on push because that runs the actions
when a PR is merged.

Dont hardcode the default branch to master
since it actions will use the same ref/commit
that it's going to test when loading workflows.

Removing the default branch makes it easier to
just backport later.


tobias-urdin:
@Mergifyio rebase

jd (approved):

PR #1169, opened 2021-07-25
Mark Python 3.6 as supported

kajinamit:
Python 3.6 is used in CI and tested.


tobias-urdin:
Fixed by https://github.com/gnocchixyz/gnocchi/pull/1166

Issue #1168, opened 2021-07-25
When python version is 3, it still find python2


HanGuangyu:
The version of Gnocchi: 4.3.4

service status is failed.
I find in 93 lines of gnocchi/cli/api.py, it use 'python' as plugin, not python3.
And if the python version is 3 in the environment, it can't find python, because the name is python3.

I want to ask whether the question should in there, should we add a judge to distinguish what version of python in use and use different plugin.

The fllowing is code:
    args = [
        "--if-not-plugin", "python", "--plugin", "python3", "--endif",
        "--%s" % conf.api.uwsgi_mode, "%s:%d" % (
            conf.host or conf.api.host,
            conf.port or conf.api.port),
        "--master",
        "--enable-threads",
        "--thunder-lock",
        "--hook-master-start", "unix_signal:15 gracefully_kill_them_all",
        "--die-on-term",
        "--processes", str(math.floor(workers * 1.5)),
        "--threads", str(workers),
        "--lazy-apps",
        "--chdir", "/",
        "--wsgi", "gnocchi.rest.wsgi",
        "--pyargv", " ".join(sys.argv[1:]),
    ]

tobias-urdin:
Fixed by https://github.com/gnocchixyz/gnocchi/pull/1171

PR #1171, opened 2021-07-26
When we use python3, it still find python2

HanGuangyu:
The args[] use "python", and if the version of python is "3",
it will can't find python, because the name is "python3" in o

Add a judge of python version, if the version of python is 3,
set plugin to "python3", if not, plugin still "python".

If the version of Python is 3, api.py can run.


HanGuangyu:
Related-Issuse: #1168

mrunge (approved):
Yes, thank you. This fixes running unit-tests or running aodh functional tests

tobias-urdin:
I guess this would only be an issue when not running inside a virtual env that maps python path to python3, but looks good to me.

tobias-urdin (approved):

tobias-urdin:
@mergifyio rebase

PR #1173, opened 2021-08-05
fix mergify job names

chungg:
so it matches github action job name


jd (approved):

tobias-urdin:
@MergifyIO backport stable/4.4

PR #1187, opened 2022-02-15
Support later versions of tenacity

tobias-urdin:
This adds support for later version of the
tenacity library and unpins the dependency.

This depends on the tooz [1] change or it
will fail.

[1] https://review.opendev.org/c/openstack/tooz/+/829412

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,7 +39,7 @@ install_requires =
     voluptuous>=0.8.10
     werkzeug
     trollius; python_version < '3.4'
-    tenacity>=4.6.0,<7.0.0
+    tenacity>=4.6.0
     WebOb>=1.4.1
     Paste
     PasteDeploy
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,7 +39,7 @@ install_requires =
     voluptuous>=0.8.10
     werkzeug
     trollius; python_version < '3.4'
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -39,7 +39,7 @@ install_requires =
     voluptuous>=0.8.10
     werkzeug
     trollius; python_version < '3.4'
-    tenacity>=4.6.0,<7.0.0
+    tenacity
```
jd, line 42: You might want to leave the min version, that should not change.

jd (changes requested):

setup.cfg
tobias-urdin: I will fix this.

mrunge:
This looks good to me.

tobias-urdin:
This is waiting for deprecation of etcd3 driver in Tooz [1] and then the fix to Tooz in [2] and a new release of Tooz.

[1] https://review.opendev.org/c/openstack/tooz/+/833107/1
[2] https://review.opendev.org/c/openstack/tooz/+/829412

tobias-urdin:
This is waiting for a new release of tooz

tobias-urdin:
tooz 2.11.0 is released which should unblock this

PR #1198, opened 2022-02-23
Stable/4.4

minruigao:
Fix issues on gnocchi-injector cli commands.

In gnocchi injecotor cli commands, it may use 'measure' instead of
'--measures', and some option value is not 'Int'.
Also, _inject need driver instead of conf.


minruigao:
> Fix issues on gnocchi-injector cli commands.
> 
> In gnocchi injecotor cli commands, it may use 'measure' instead of '--measures', and some option value is not 'Int'. Also, _inject need driver instead of conf.

PR #1202, opened 2022-03-11
dynamic aggregates - metric not found

javacruft:
Return a 404 (metric not found) rather than 400 (bad request) when
metrics associated with a dynamic aggregate are not found.


tobias-urdin:
Maybe @jd want's to comment on this behavior change.

mrunge:
This implements the correct behavior imho, but maybe I am missing something?

jd (approved):
LGTM

PR #1208, opened 2022-03-22
ci: fixes to get CI working

tobias-urdin:
Some fixes to the CI to get it working in the same PR:

* Fix permission for the checked out code in GitHub Actions to the user running inside the Docker container
* Remove distutils code from pifpaf by using the fork, PR is opened against pifpaf
* Migrate to stestr to fix python 3.8


jd (approved):

tobias-urdin:
adding debug output temporarily to jobs, perhaps we can catch why py38 is not working

PR #1234, opened 2022-05-26
Support Python 3.9

kajinamit:
This change adds Python 3.9 to supported python versions and adds
test coverage for it. CentOS Stream 9 now uses Python 3.9 by default
so the support is required to run Gnocchi in CentOS Stream 9.

This change also adds a few metadata items to setup.cfg to enforce
usage of Python >= 3.6.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 minversion = 2.4
-envlist = {py36,py38}-{postgresql,mysql}{,-file,-swift,-ceph,-s3},pep8
+envlist = {py36,py38,py39}-{postgresql,mysql}{,-file,-swift,-ceph,-s3},pep8
 skipsdist = True
 
 [testenv]
@@ -54,7 +54,7 @@ commands =
     {toxinidir}/run-tests.sh {posargs}
     {toxinidir}/run-func-tests.sh {posargs}
 
-[testenv:{py36,py38}-postgresql-file-upgrade-from-4.4]
+[testenv:{py36,py38,py39}-postgresql-file-upgrade-from-4.4]
 # We should always recreate since the script upgrade
 # Gnocchi we can't reuse the virtualenv
 recreate = True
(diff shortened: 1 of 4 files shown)


jd (dismissed):

tobias-urdin:
@mergifyio rebase

tobias-urdin (approved):
Thanks Takashi! I've spent some time fixing CI again, let's see if this becomes green so we can merge it.

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 minversion = 2.4
-envlist = {py36,py38}-{postgresql,mysql}{,-file,-swift,-ceph,-s3},pep8
+envlist = {py36,py38,py39}-{postgresql,mysql}{,-file,-swift,-ceph,-s3},pep8
 skipsdist = True
 
 [testenv]
```
tobias-urdin, line 6: need to add py39 in some more places in this file below

tobias-urdin (changes requested):

tox.ini
kajinamit: Good catch. Fixed in the latest version.

tobias-urdin:
https://github.com/gnocchixyz/gnocchi/pull/1246 is merged, please resolve conflict and rebase.

tobias-urdin:
@kajinamit please apply this diff, we need to skip Ceph jobs on Python 3.9 since Ubuntu 20.04 python3-rados bindings is only for Python 3.8 (default version in dist).

```
diff --git a/.github/workflows/gnocchi.yml b/.github/workflows/gnocchi.yml
index 58aee490..d318c0f8 100644
--- a/.github/workflows/gnocchi.yml
+++ b/.github/workflows/gnocchi.yml
@@ -88,6 +88,12 @@ jobs:
             python: py36
           - env: postgresql-ceph
             python: py36
+          - env: mysql-ceph-upgrade-from-4.4
+            python: py39
+          - env: mysql-ceph
+            python: py39
+          - env: postgresql-ceph
+            python: py39
     steps:
       - uses: actions/checkout@v2
       - run: sudo chown -R 1001:1001 $GITHUB_WORKSPACE
```

kajinamit:
@tobias-urdin That makes sense. Thanks for the suggestion ! Applied the change.

tobias-urdin (approved):
Thanks! Sorry for all the fuzz with broken CI.

kajinamit:
> Thanks! Sorry for all the fuzz with broken CI.

No problem, and thank you for your great help !

tobias-urdin:

PR #1236, opened 2022-06-01
Pin protobuf<4.0.0

tobias-urdin:
The newer release requires us to regenerate the
protos from the remote.proto file.


tobias-urdin:

PR #1239, opened 2022-06-02
Fix faulty release note

tobias-urdin:


tobias-urdin:

Issue #1231, opened 2022-05-17
PEP517


Alessandro-Barbieri:
Could you move the buildsystem to [PEP517](https://peps.python.org/pep-0517/)?

tobias-urdin:
I might look into this later but for now I don't have the time. I am open to a PR that tries to solve that. There is some custom code for the build process that might need to be looked at as well in ``gnocchi/setuptools.py``

Issue #1230, opened 2022-05-09
Push Gnocchi 4.4.2 version to pypi


tobias-urdin:
We need to push version 4.4.2 tag to pypi
https://github.com/gnocchixyz/gnocchi/releases/tag/4.4.2

tobias-urdin:
@jd please push the 4.4.2 tag to pypi, if you feel ok with it you can add me as maintainer https://pypi.org/user/tobias-urdin/

jd:
@tobias-urdin I invited you!

tobias-urdin:
Uploaded 4.4.2 to PyPI.

```
$ python3 setup.py bdist_wheel
$ python3 setup.py sdist

$ twine check dist/gnocchi-4.4.2-py2.py3-none-any.whl dist/gnocchi-4.4.2.tar.gz 
Checking dist/gnocchi-4.4.2-py2.py3-none-any.whl: PASSED
Checking dist/gnocchi-4.4.2.tar.gz: PASSED

$ twine upload dist/gnocchi-4.4.2-py2.py3-none-any.whl dist/gnocchi-4.4.2.tar.gz 
Uploading gnocchi-4.4.2-py2.py3-none-any.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 335.4/335.4 kB • 00:00 • 1.0 MB/s
Uploading gnocchi-4.4.2.tar.gz
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 990.9/990.9 kB • 00:00 • 2.2 MB/s

View at:
https://pypi.org/project/gnocchi/4.4.2/
```

tobias-urdin:
Note to self, see new build system https://github.com/gnocchixyz/gnocchi/pull/1247

PR #1247, opened 2022-06-13
Change build system

tobias-urdin:
This moves the build system to using isolated builds. Moves the custom out of gnocchi/setuptools.py into setup.py so that we can use it. Cleans up some custom integrations that minimizes the maintenance burden.

To build you now ``pip install build && python3 -m build``

The builds is placed in the ``dist`` directory. Then simply upload with something like ``twine`` as shown in https://github.com/gnocchixyz/gnocchi/issues/1230#issuecomment-1122176441

Fixes: #1231


setup.cfg
```diff
@@ -1,5 +1,6 @@
 [metadata]
 name = gnocchi
+version = 4.5
```
jd, line 3: you might want to use `setuptools_scm`

tobias-urdin: Thanks! I should've added that when porting it, solved now.

tobias-urdin:

PR #1250, opened 2022-06-13
ci: remove old travis ci config

tobias-urdin:


tobias-urdin:
We never fixed deploy to pypi with actions, I have maintainer now so I can work on that in the future without keeping this old (not working) part.

tobias-urdin:

PR #1259, opened 2022-06-21
Added rateofchangesec operator to calculate change of value per second

art-vasilyev:
Added operator to calculate change between cumulative metric measures per second.

Simplified formula:
```
rate = (v2 - v1) / (t2 - t1)
```

It may be used to calculate cpu utilisation from cumulative cpu metric:
```
gnocchi aggregates '(* (/ (rateofchangesec (metric cpu mean)) 1000000000) 100))' id=<resource_id>
```


tobias-urdin (changes requested):
Thanks! Please add a release note.

tobias-urdin (approved):

Issue #1267, opened 2022-08-05
Updated protobuf module fails Gnocchi installation in devstack


yadneshk:
Gnocchi uses `protobuf` pip module which is listed as one of the [dependencies](https://github.com/openstack/requirements/blob/master/upper-constraints.txt#L373) in global requirements repository. Recently this package and several others were updated.

During devstack installation, when `ceilometer_upgrade` is executed, gnocchi reports this error and 
```
Aug 05 08:39:03 devstack devstack@gnocchi-api.service[248241]:   File "/opt/stack/gnocchi/gnocchi/rest/api.py", line 51, in <module>
Aug 05 08:39:03 devstack devstack@gnocchi-api.service[248241]:     from gnocchi.rest.prometheus import remote_pb2
Aug 05 08:39:03 devstack devstack@gnocchi-api.service[248241]:   File "/opt/stack/gnocchi/gnocchi/rest/prometheus/remote_pb2.py", line 33, in <module>
Aug 05 08:39:03 devstack devstack@gnocchi-api.service[248241]:     _descriptor.EnumValueDescriptor(
Aug 05 08:39:03 devstack devstack@gnocchi-api.service[248241]:   File "/opt/stack/.local/lib/python3.9/site-packages/google/protobuf/descriptor.py", line 755, in __new__
Aug 05 08:39:03 devstack devstack@gnocchi-api.service[248241]:     _message.Message._CheckCalledFromGeneratedFile()
Aug 05 08:39:03 devstack devstack@gnocchi-api.service[248241]: TypeError: Descriptors cannot not be created directly.
Aug 05 08:39:03 devstack devstack@gnocchi-api.service[248241]: If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
Aug 05 08:39:03 devstack devstack@gnocchi-api.service[248241]: If you cannot immediately regenerate your protos, some other possible workarounds are:
Aug 05 08:39:03 devstack devstack@gnocchi-api.service[248241]:  1. Downgrade the protobuf package to 3.20.x or lower.
Aug 05 08:39:03 devstack devstack@gnocchi-api.service[248241]:  2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).
[...]

PR #1268, opened 2022-08-05
Regenerate remote_pb2.py with protoc 3.19.0

yadneshk:
Execute "protoc --proto_path=. --python_out=. remote.proto" to
regenerate remote_pb2.py

Fixes: #1267

Signed-off-by: Yadnesh Kulkarni <ykulkarn@redhat.com>


yadneshk:
Used the same command from https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/rest/prometheus/README to recomplie remote_pb2.py file

eyalb1:
I think you need to change the version to be` >4.0.0`
https://github.com/gnocchixyz/gnocchi/blob/master/setup.cfg#L84

kajinamit:
> I think you need to change the version to be` >4.0.0` https://github.com/gnocchixyz/gnocchi/blob/master/setup.cfg#L84

I agree that we should update the version requirement. I could not find any good material explaining the minimum version but I think >= 3.19.0 would be reasonable because the code is regenerated with protoc 3.19.0

kajinamit (changes requested):
Please update setup.cfg. See conversations for further details.

kajinamit (approved):
We might want to define lower boundary but I'm not certain about the correct version so I'm good with the current version without it.

yadneshk:
@tobias-urdin @mrunge please let me know your thoughts on this pull request

tobias-urdin:
Thanks for looking at this, I would like to know if we need to enforce a minimum version of protobuf package, anybody knows?

yadneshk:
We could set it to `protobuf>=4.21.1`

setup.cfg
```diff
@@ -81,7 +81,7 @@ ceph_alternative =
     python-rados>=12.2.0 # not available on pypi
 prometheus =
     python-snappy
-    protobuf<4.0.0
+    protobuf
```
tobias-urdin, line 84: Lets do >=3.19.0

tobias-urdin (changes requested):

tobias-urdin (approved):

PR #1269, opened 2022-08-09
Regenerate code with protoc = 3.19.0

kajinamit:
... to fix the following error raised by the recent protobuf library.

```
TypeError: Descriptors cannot not be created directly.
If this call came from a _pb2.py file, your generated code is out of
date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible
workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will
    use pure-Python parsing and will be much slower).
More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
```


kajinamit:
Duplictate of https://github.com/gnocchixyz/gnocchi/pull/1268

PR #1270, opened 2022-09-04
Fix incomplete switch to policy.yaml

kajinamit:
Commit 0d610f6dcdfa4cb95e6832aff5054da247052a10 was supposed to switch
default policy file from policy.json to policy.yaml but that change
was incomplete and still the old policy.json is searched first.

This change fixes it and ensures policy.yaml is loaded first.


jd (approved):

kajinamit:
@mergify backport stable/4.4

PR #1272, opened 2022-09-06
Revert "Fix incomplete switch to policy.yaml"

tobias-urdin:
This reverts commit a53fcfc3ab60be6e223712d46e050e8740f16c23.

### DIFF
--- a/gnocchi/cli/api.py
+++ b/gnocchi/cli/api.py
@@ -37,7 +37,7 @@ def prepare_service(conf=None):
         conf = cfg.ConfigOpts()
 
     opts.set_defaults()
-    policy_opts.set_defaults(conf, 'policy.yaml')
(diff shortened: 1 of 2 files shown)


kajinamit:
It seems bump of oslo.policy was not unnecessary. I've submitted PR to revert version bump in
master: https://github.com/gnocchixyz/gnocchi/pull/1273
stable/4.4: https://github.com/gnocchixyz/gnocchi/pull/1274

However the original change is a kind of breaking (because that changes the default policy file path) so I'm ok with reverting this if we don't prefer making that change in stable branch, which is used with multiple OpenStack stable releases. Maybe we better create stable/4.5 instead.

tobias-urdin:
Per discussion on IRC, we merge master #1273, close #1274 and merge this revert.

PR #1275, opened 2022-09-30
Always call install_scripts on pip install

tobias-urdin:


tobias-urdin:

PR #1277, opened 2022-10-14
Concurrency issues when multiple patches happen for the same resources

rafaelweingartner:
An issue was found when Ceilometer notification is stopped, and samples accumulate in the RabbitMQ. These samples might contain the same set of attributes that would trigger one revision. However, due to a concurrency issue, when Ceilometer notification is started back again, all of them are processed in almost the same moment, and then pushed back to Gnocchi API. Then, in Gnocchi API the detection to see if the resource needs a revision is executed outside of a semaphore control (e.g. read/write transaction with the DB with a table being locked). Therefore, the code detects that a revision is needed, and generated multiple revisions, which have the same content but different revision start and end.

This patches aims to resolve that situation. The solution adopted is to check again in the transaction (when the lock was acquired), before the revision is created, if the revision is really needed. It it is not needed, a log entry is generated and we stop the processing by returnign the resource data.

### DIFF
--- a/gnocchi/indexer/sqlalchemy.py
+++ b/gnocchi/indexer/sqlalchemy.py
@@ -897,12 +897,29 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
             return r
 
     @retry_on_deadlock
-    def update_resource(self, resource_type,
-                        resource_id, ended_at=_marker, metrics=_marker,
-                        append_metrics=False,
-                        create_revision=True,
+    def update_resource(self, resource_type, resource_id,
+                        ended_at=_marker, metrics=_marker,
+                        append_metrics=False, create_revision=True,
                         **kwargs):
         with self.facade.writer() as session:
+            data_to_update = kwargs.copy()
+
+            data_to_update['ended_at'] = ended_at
+            data_to_update['metrics'] = metrics
+            if create_revision:
+                resource = self.get_resource(
+                    resource_type, resource_id, with_metrics=True)
+                if not utils.is_resource_revision_needed(
+                        resource, data_to_update):
+                    LOG.info("We thought that a revision for resource "
+                             "[%s] was needed. However, after locking the "
+                             "table and checking it again, we found that it "
+                             "is not needed anymore. This is due to a "
+                             "concurrency issue that might happen. Therefore, "
+                             "no revision is going to be generated at this "
+                             "time.", data_to_update)
+                    create_revision = False
+
             mappers = self._resource_type_to_mappers(session, resource_type)
             resource_cls = mappers["resource"]
             resource_history_cls = mappers["history"]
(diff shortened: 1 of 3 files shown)


gnocchi/rest/api.py
```diff
@@ -1073,16 +1073,12 @@ def patch(self):
...
             return resource
 
-        for k, v in six.iteritems(body):
-            if k != 'metrics' and getattr(resource, k) != v:
-                create_revision = True
-                break
-        else:
+        create_revision = utils.is_resource_revision_needed(resource, body)
+        if not create_revision and 'metrics' not in body:
             if 'metrics' not in body:
```
pedro-martins, line 1078: this if could be removed

tobias-urdin:
This makes sense, it's probably very hard to test the race condition there, but do we have a test for the normal case of a new revision to verify this doesn't break normal behavior? (either unit or a gabbit)

Also please squash (cleanup) the commits into one (or more, whatever is preferred) but provide good commit messages.

gnocchi/indexer/sqlalchemy.py
```diff
@@ -897,12 +897,29 @@ def create_resource(self, resource_type, id,
...
+                if not utils.is_resource_revision_needed(
+                        resource, data_to_update):
+                    LOG.info("We thought that a revision for resource "
+                             "[%s] was needed. However, after locking the "
+                             "table and checking it again, we found that it "
+                             "is not needed anymore. This is due to a "
+                             "concurrency issue that might happen. Therefore, "
+                             "no revision is going to be generated at this "
+                             "time.", data_to_update)
+                    create_revision = False
```
chungg, line 921: i wonder if it makes sense to just move the check revision logic to https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/indexer/sqlalchemy.py#L922-L923. that way it would be behind the lock created by for_update.

that said, now it's always locking and i've no idea the impact.

rafaelweingartner: > This makes sense, it's probably very hard to test the race condition there, but do we have a test for the normal case of a new revision to verify this doesn't break normal behavior? (either unit or a gabbit)
> 
> Also please squash (cleanup) the commits into one (or more, whatever is preferred) but provide good commit messages.

Thanks for the review. I squashed the commits. Also, regarding tests, there are already gabbi tests that cover resource revisions, and all of them are passing just fine.

rafaelweingartner: @chungg I agree with you. Ideally, I Would have done that. However, I implemented this way to be less intrusive as possible in the code base. 

I can move the decision to generate revision to the update_resource method only, if you prefer.

tobias-urdin: I'm fine with this, I'll let @chungg respond and approve if he's OK with it.

rafaelweingartner:
Hello guys, 
Is there something missing here?

tobias-urdin (approved):

rafaelweingartner:
Thanks!!

PR #1279, opened 2022-12-26
Enable the upgrade to numpy 1.24

rafaelweingartner:


rafaelweingartner:
@jd and @chungg can you take a look into this one. It is an important update. Otherwise, the pipeline starts to break.

rafaelweingartner:
@tobias-urdin can you take a look into this one as well?

gnocchi/carbonara.py
```diff
@@ -196,8 +196,8 @@ def quantile(self, q):
         ordered = numpy.lexsort((self._ts['values'], self.indexes))
         min_pos = numpy.cumsum(self.counts) - self.counts
         real_pos = min_pos + (self.counts - 1) * (q / 100)
-        floor_pos = numpy.floor(real_pos).astype(numpy.int, copy=False)
-        ceil_pos = numpy.ceil(real_pos).astype(numpy.int, copy=False)
+        floor_pos = numpy.floor(real_pos).astype(numpy.integer, copy=False)
```
tobias-urdin, line 199: When was numpy.integer introduced? I.e do we need to have a new lowest supported version for numpy?

run-upgrade-tests.sh
```diff
@@ -2,14 +2,25 @@
...
+  eval $(pifpaf --debug --env-prefix INDEXER run postgresql)
 elif [ "$1" == "mysql-ceph" ]; then
-  eval $(pifpaf --env-prefix INDEXER run mysql)
-  eval $(pifpaf --env-prefix STORAGE run ceph)
+  # Installing PifPaf from source due to the lack of a new version to handle
+  # Ceph global insecure claims. The patch was introduced in PifPaf via commit
+  # https://github.com/jd/pifpaf/commit/fb376a83a47d678952672a7f5d36a02101135fb2,
+  # but it has never been released. Therefore, we need to install it here from
+  # master/main branch in the upstream repository
+  pip install install git+https://github.com/jd/pifpaf.git@51f74a3d8743a7ac33259413df7efc30df993460
```
tobias-urdin, line 13: Please set this in tox.ini instead, you can see that I have some old pins for pifpaf there you can change those as the one I pin there is probably merged already.

run-upgrade-tests.sh
```diff
@@ -2,14 +2,25 @@
...
+
+  echo "Deploying MySQL with PifPaf."
```
tobias-urdin, line 15: All --debug changes here is probably fine, back when we used Travis there were a maximum amount of lines allowed in output but it seems that Github Actions enforce this at 50 MB limit instead of amount of lines.

So if the jobs passes with this it should be fine.

run-upgrade-tests.sh
```diff
@@ -51,15 +62,15 @@ inject_data() {
 
     {
         measures_sep=""
-        MEASURES=$(python -c 'import datetime, random, json; now = datetime.datetime.utcnow(); print(json.dumps([{"timestamp": (now - datetime.timedelta(seconds=i)).isoformat(), "value": random.uniform(-100000, 100000)} for i in range(0, 288000, 10)]))')
+        MEASURES=$(python -c 'import datetime, random, json; now = datetime.datetime.utcnow(); print(json.dumps([{"timestamp": (now + datetime.timedelta(seconds=i)).isoformat(), "value": random.uniform(-100000, 100000)} for i in range(0, 288000, 10)]))')
```
tobias-urdin, line 65: Not sure why this is changed?

run-upgrade-tests.sh
```diff
@@ -85,7 +96,13 @@ else
     STORAGE_URL=file://$GNOCCHI_DATA
 fi
 
+# This downgrade of `numpy` is needed to enable the merge of PR
+# https://github.com/gnocchixyz/gnocchi/pull/1279, which is the PR that
+# introduces the support of numpy >= 1.24. After we merge it, we can remove
+# this downgrade here.
+pip install "numpy<1.24"
```
tobias-urdin, line 103: Pin this separately in setup.cfg and then propose a commit following the changes required that unpins it.

setup.cfg
```diff
@@ -26,7 +26,7 @@ packages =
 include_package_data = true
 
 install_requires =
-    numpy>=1.9.0
+    numpy>=1.19.5
```
tobias-urdin, line 29: Is this the new lowest required verison for numpy.integer?

It's very high which means we'd get issues with some distributions that ship an older version, for example CentOS Stream 8 has 1.14.3 and CentOS Stream 9 has 1.20.1, I haven't checked Ubuntu.

tobias-urdin (changes requested):

gnocchi/carbonara.py
rafaelweingartner: It seems to be first introduced in 1.19.0. New lowest, I am not sure what you mean. Do you mean in `setup.cfg`

setup.cfg
rafaelweingartner: You are right, we do not need to change the lowest. The `Integer` member has always been there. However, the `int` is the one that was removed in the 1.24.0 version. I will amend my patch

run-upgrade-tests.sh
rafaelweingartner: I am not sure I follow your idea of creating a new commit. It will not work if I do that in this PR.

run-upgrade-tests.sh
rafaelweingartner: It was changed during my tests, and I forgot to revert it.

run-upgrade-tests.sh
rafaelweingartner: ok, thanks!

I added the debug to get more meaningful information about the tests

run-upgrade-tests.sh
rafaelweingartner: I fixed in tox then.

gnocchi/carbonara.py
rafaelweingartner: I was wrong, it is already in 1.9.0. Therefore, I removed my change in lower contraints.

rafaelweingartner:
@tobias-urdin I applied your suggestions. Can you check the PR again?

run-upgrade-tests.sh
tobias-urdin: You can remove this entirely as it's not needed right?

tobias-urdin (changes requested):

rafaelweingartner:
@tobias-urdin if I remove the downgrade of numpy the tests do not pass, as we install in the upgrade tests the Gnocchi version from stable/4.4, which is not ready for numpy 1.24, which is the one installed by default.

tobias-urdin:
You should propose a commit that pins numpy, commit(s) that fixes support for newer numpy, create a commit that reverts commit 1.

rafaelweingartner:
> You should propose a commit that pins numpy, commit(s) that fixes support for newer numpy, create a commit that reverts commit 1.

I think that I did not get it. What is your proposal?
- create PR to pin numpy to be smaller than 1.24 and merge it
-  then move on with this PR and introduce the support to numpy 1.24
Is that it?

tobias-urdin:
Yes, like that, I should have explained myself better. You can do it as commits doesn't have to be a separate PR but would be preferable so that we can verify the CI gets green between changes.

tobias-urdin:
fixed in https://github.com/gnocchixyz/gnocchi/pull/1281

rafaelweingartner:
@tobias-urdin thanks for merging the proposal. I would execute the process we agreed on. However, I had no spare time this week. You seem to have done it differently from what we discussed though. I do not understand how the upgrade tests passed with your PR though...

tobias-urdin:
Hello 👋 had to pin numpy in stable/4.4 [1] since upgrade jobs on master install old version from stable/4.4 branch and then new version from master. I then simply backported the change on master to stable/4.4 to unpin numpy again on stable/4.4

[1] https://github.com/gnocchixyz/gnocchi/pull/1282

rafaelweingartner:
I see, so you implemented what we discussed. Thanks for that.

PR #1283, opened 2023-01-20
deps: set numpy>=1.14.0

tobias-urdin:
The copy kwarg on astype() was not introduced until 1.14.0 and we are using that in multiple places.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ packages =
 include_package_data = true
 
 install_requires =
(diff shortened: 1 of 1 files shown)


tobias-urdin:
@mergifyio rebase

tobias-urdin:
I will enforce this in master so for next version but not backport to stable/4.4 even though using a version older than 1.14.0 in 4.4 would not work.

PR #1286, opened 2023-01-26
Adjust metric not found exception

rafaelweingartner:
Following the discussion on [1], Gnocchi was not properly handling the metric not found exceptions, which in turn causes a problem with Gnocchi client.

The exceptions were being rendered as a JSON, and not a string inside the description field. This PR fixes that.
The problem seems to have been caused by: https://github.com/gnocchixyz/gnocchi/commit/455c1b9b60942230028a6012a674c83816afd6ef

P.S. There are some other situations that the response is a JSON inside the description attribute. I would prefer as is and fix the issue in the Client. However, I see that this would break the backward compatibility.

[1] https://github.com/gnocchixyz/python-gnocchiclient/pull/133#discussion_r1084311700

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ cover
 .coverage
 dist
 upgrade/
+
+# Ignore the IDE files
+.idea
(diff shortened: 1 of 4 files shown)


tobias-urdin:
Thanks! This makes sense.

tobias-urdin (approved):

PR #1287, opened 2023-01-27
Fix issue with None value attributes

rafaelweingartner:
With Python 3, the ``sort`` function is not handling `None` values. Therefore, when we handle resources with values as ``None`` an exception happens. This patch fixes such situations.

### DIFF
--- a/gnocchi/rest/aggregates/api.py
+++ b/gnocchi/rest/aggregates/api.py
@@ -316,7 +316,7 @@ class Grouper(object):
         for value in to_group:
             self.truncate_resource_time_window(value, is_first=is_first)
             is_first = False
-        to_group.sort(key=lambda x: tuple((attr, x[attr])
+        to_group.sort(key=lambda x: tuple((attr, str(x[attr] or ''))
                                           for attr in self.groups))
         grouped_values = \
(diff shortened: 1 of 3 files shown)


rafaelweingartner:
@tobias-urdin this is also an important one. However, its tests will only pass after the following patches have been merged:

- https://github.com/gnocchixyz/gnocchi/pull/1288
- https://github.com/gnocchixyz/gnocchi/pull/1289

tobias-urdin:
Could this be reproduced with a test? I'm trying to understand what the change will do to the grouped data, will there be entries with an empty string? Should those be filtered out instead?

rafaelweingartner:
The code is not designed in a way  that enables us to easily do a unit testing. Let me think about what we can do here.

tobias-urdin:
@mergifyio rebase

rafaelweingartner:
@tobias-urdin I added the unit tests for this one, as you suggested.

tobias-urdin (approved):

tobias-urdin:
Thanks @rafaelweingartner – let me know if you have anything else you want in master I will cut a 4.5 release sometime this week.

PR #1288, opened 2023-01-30
Pin down SQL Alchemy to <2

rafaelweingartner:
Version 2 of SQL Alchemy has been released, and it brings considerable changes to it [1]. The one that is affecting us right away is the philosophy of always using a transaction. That means, for every connection open with the database (DB), there is automatically a transaction opened. This, in turn, breaks the current code structure that we have, where we try to open a transaction every time we get a connection to write or read in the DB. The break happens because the code is trying to open a transaction inside an already existing one (the transaction opened by SQL Alchemy itself).

This whole situation only happens with SQL Alchemy >=2. Therefore, for now, the solution is to pin down its (SQL Alchemy) version until we have a final solution, which is going to have a considerable impact in the transaction management that we currently have.

[1] https://pypi.org/project/SQLAlchemy/2.0.0/

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,13 +61,13 @@ keystone =
 mysql =
     pymysql
     oslo.db>=4.29.0
-    sqlalchemy
+    sqlalchemy<2
     sqlalchemy-utils
     alembic>=0.7.6,!=0.8.1,!=0.9.0
 postgresql =
     psycopg2
     oslo.db>=4.29.0
-    sqlalchemy
+    sqlalchemy<2
     sqlalchemy-utils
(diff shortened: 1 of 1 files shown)


rafaelweingartner:
@tobias-urdin can you check this one?

It is a similar situation as we had with Numpy. We first need to merge in the `stable\4.4`, and then we can merge it upstream as well.

tobias-urdin (approved):
Thanks, let's go ahead an pin this and we can work out the other stuff later.

rafaelweingartner:
@tobias-urdin thanks!

PR #1298, opened 2023-03-20
Allow using specific CA certificates file

kajinamit:
This change allows configuring ssl_ca_certs argument of redis client using the query parameter, so that users can use a specific CA certificates file instead of system global repository.


tobias-urdin:
@mergifyio rebase

tobias-urdin:
@Mergifyio rebase

PR #1303, opened 2023-06-15
Remove Python 3.6 testing

tobias-urdin:


tobias-urdin:
@mrunge Unfortunately it also occuring on Python > 3.6 as well. There is a bug somewhere in the DB layer (issues with SQLAlchemy) where the select() for resource history doesn't return the proper rows.

mrunge:
thank you. 

It looks like this is one of the issues becoming more prominent

chungg (approved):

tobias-urdin:
@mergifyio rebase

PR #1307, opened 2023-07-10
Optimize Gnocchi aggregates API

rafaelweingartner:
Gnocchi aggregates API can present slowness with time, due to the growing number of resources and revisions in large scale OpenStack Cloud environments. This happens due to the situation where Gnocchi does not apply filters in the MySQL queries in the resource tables. I mean, start/stop filters are not applied, and others as well, which makes Gnocchi to manipulate all dataset in the MySQL database and then the filtering is executed in Python.

To cope with that situation, we are proposing an optimization to improve the query executed by Gnocchi in MySQL, and apply the filtering right away; thus, the dataset manipulated by MySQL will be drastically reduced. After applying the patch, we noticed a reduction of 5-6 folds in the response time for the Gnocchi aggregates API.

### DIFF
--- a/gnocchi/indexer/sqlalchemy.py
+++ b/gnocchi/indexer/sqlalchemy.py
@@ -1120,7 +1120,10 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
                     attribute_filters_to_use[value_index] = value_sanitized
 
         elif is_value_dict:
-            for key, value in attribute_filter:
+            all_keys = list(attribute_filter.keys())
+            for key in all_keys:
+                value = attribute_filter.get(key)
+
                 # The value is a leaf when it is not of type dict of list.
                 is_value_leaf = not (isinstance(
                     value, dict) or isinstance(value, list))
--- a/gnocchi/indexer/sqlalchemy.py
+++ b/gnocchi/indexer/sqlalchemy.py
@@ -1120,10 +1120,7 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
                     attribute_filters_to_use[value_index] = value_sanitized
 
         elif is_value_dict:
-            all_keys = list(attribute_filter.keys())
-            for key in all_keys:
-                value = attribute_filter.get(key)
-
+            for key, value in attribute_filter:
                 # The value is a leaf when it is not of type dict of list.
                 is_value_leaf = not (isinstance(
                     value, dict) or isinstance(value, list))
--- a/gnocchi/indexer/sqlalchemy.py
+++ b/gnocchi/indexer/sqlalchemy.py
@@ -14,6 +14,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import copy
 import datetime
 import itertools
 import operator
@@ -1071,7 +1072,100 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
                 q = q.options(sqlalchemy.orm.joinedload(Resource.metrics))
             return session.scalars(q).first()
 
-    def _get_history_result_mapper(self, session, resource_type):
+    def extracts_filters_for_table(self, attribute_filter,
+                                   allowed_keys_for_table=[
+                                       'creator', 'started_at', 'ended_at',
+                                       'user_id', 'project_id',
+                                       'original_resource_id', 'id', 'type']):
+        """Extracts the filters for resource history table.
+
+        Extracts the filters that can be used in the resource history table to
+        apply in the aggregates query that we execute in the database.
+        """
+
+        attribute_filters_to_use = copy.deepcopy(attribute_filter)
+
+        LOG.debug("Executing the processing of attributes filters [%s] for "
+                  "resource history table.", attribute_filters_to_use)
+
+        is_value_list = isinstance(attribute_filters_to_use, list)
+        is_value_dict = isinstance(attribute_filters_to_use, dict)
+        is_value_dict_or_list = (is_value_dict or is_value_list)
+
+        if not is_value_dict_or_list:
+            LOG.debug("Attribute filter [%s] is not of expected types [list "
+                      "or dict]. Therefore, we do not do anything with it.",
+                      attribute_filters_to_use)
+            return attribute_filters_to_use
+
+        if is_value_list:
+            for attribute in attribute_filter:
+                LOG.debug("Sending attribute filter [%s] to be processed, "
+                          "as it is part of a list of attribute filters.",
+                          attribute)
+
+                value_sanitized = self.extracts_filters_for_table(
+                    attribute, allowed_keys_for_table=allowed_keys_for_table)
+
+                if not value_sanitized:
+                    LOG.debug("Value [%s] was totally cleaned after being "
+                              "sanitized. Therefore, we remove it from our "
+                              "attribute filter list.", attribute)
+                    attribute_filters_to_use.remove(attribute)
+                else:
+                    LOG.debug("Replacing value [%s] in list with the sanitized"
+                              "value [%s] in its current position.",
+                              attribute, value_sanitized)
+                    value_index = attribute_filters_to_use.index(attribute)
+                    attribute_filters_to_use[value_index] = value_sanitized
+
+        elif is_value_dict:
+            all_keys = list(attribute_filter.keys())
+            for key in all_keys:
+                value = attribute_filter.get(key)
+
+                # The value is a leaf when it is not of type dict of list.
+                is_value_leaf = not (isinstance(
+                    value, dict) or isinstance(value, list))
+
+                if key not in allowed_keys_for_table and is_value_leaf:
+                    attribute_being_remove = attribute_filters_to_use.pop(key)
+                    LOG.debug('Removing attribute [%s] with value [%s] from '
+                              'attributes [%s] as it is not an expected key '
+                              'value [%s].', key, attribute_being_remove,
+                              attribute_filter,
+                              allowed_keys_for_table)
+                else:
+                    LOG.debug("Sending attribute [key=%s, value=%s] from "
+                              "dictionary to be processed.", key, value)
+                    value_sanitized = self.extracts_filters_for_table(
+                        value, allowed_keys_for_table=allowed_keys_for_table)
+
+                    is_value_changed = value != value_sanitized
+                    if not is_value_changed:
+                        LOG.debug("Value [%s] for key [%s] did not changed. "
+                                  "Therefore, we go for the next iteration.",
+                                  value, key)
+                        continue
+                    if not value_sanitized:
+                        LOG.debug("Value from dict [%s] was totally cleaned after"
+                                  " being sanitized. Therefore, we remove it from "
+                                  "our attribute filter dictionary.", value)
+                        attribute_filters_to_use.pop(key)
+                    else:
+                        LOG.debug("Replacing attribute [%s] in dict, with "
+                                  "sanitized data [%s]. Old value was [%s].",
+                                  key, value_sanitized, value)
+                        attribute_filters_to_use[key] = value_sanitized
+        else:
+            LOG.debug("This condition should never happen. Attribute filter [%s] "
+                      "is not of expected types [list or dict].",
+                      attribute_filters_to_use)
+        return attribute_filters_to_use
+
+    def _get_history_result_mapper(self, session, resource_type,
+                                   attribute_filter=None):
+
         mappers = self._resource_type_to_mappers(session, resource_type)
         resource_cls = mappers['resource']
         history_cls = mappers['history']
@@ -1089,8 +1183,53 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
         s1 = select(*history_cols.values())
         s2 = select(*resource_cols.values())
         if resource_type != "generic":
+            engine = session.connection()
+
+            # Get the available columns for ResourceHistory table
+            resource_history_filters_names = list(map(
+                lambda column: column.name, sqlalchemy.inspect(
+                    ResourceHistory).columns))
+
+            history_filters = self.extracts_filters_for_table(
+                attribute_filter,
+                allowed_keys_for_table=resource_history_filters_names)
+
+            LOG.debug("Filters to be used [%s] in query according to resource "
+                      "history columns [%s] extracted from [%s].",
+                      history_filters, resource_history_filters_names,
+                      attribute_filter)
+
+            # Get the available columns for Resource table
+            resource_filter_names = list(map(
+                lambda column: column.name, sqlalchemy.inspect(
+                    Resource).columns))
+
+            resource_filters = self.extracts_filters_for_table(
+                attribute_filter, allowed_keys_for_table=resource_filter_names)
+            LOG.debug("Filters to be used [%s] in query according to resource "
+                      "columns [%s] extracted from [%s].", resource_filters,
+                      resource_filter_names, attribute_filter)
+
             s1 = s1.where(history_cls.revision == ResourceHistory.revision)
+            if history_filters:
+                f1 = QueryTransformer.build_filter(
+                    engine.dialect.name, ResourceHistory, history_filters)
+                s1 = s1.filter(f1)
+            else:
+                LOG.debug("No filters supplied to be applied for the resource "
+                          "history table. Attribute filters: [%s].",
+                          attribute_filter)
+
             s2 = s2.where(resource_cls.id == Resource.id)
+
+            if resource_filters:
+                f2 = QueryTransformer.build_filter(
+                    engine.dialect.name, Resource, resource_filters)
+                s2 = s2.filter(f2)
+            else:
+                LOG.debug("No filters supplied to be applied for the resource "
+                          "table. Attribute filters: [%s].", attribute_filter)
+
         union_stmt = sqlalchemy.union(s1, s2)
         stmt = union_stmt.alias("result")
 
@@ -1124,7 +1263,7 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
         with self.facade.independent_reader() as session:
             if history:
                 target_cls = self._get_history_result_mapper(
-                    session, resource_type)
+                    session, resource_type, attribute_filter)
                 unique_keys = ["id", "revision"]
             else:
                 target_cls = self._resource_type_to_mappers(
@@ -1185,7 +1324,12 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
 
             # Always include metrics
             q = q.options(sqlalchemy.orm.joinedload(target_cls.metrics))
+
+            LOG.debug("Executing query [%s] to search for resources.", q)
             all_resources = session.scalars(q).unique().all()
+
+            LOG.debug("Resources [quantity=%s] [%s] found with query: [%s].",
+                      len(all_resources), all_resources, q)
 
             if details:
                 grouped_by_type = itertools.groupby(
@@ -1213,9 +1357,11 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
 
                         q = select(target_cls).filter(f)
                         # Always include metrics
-                        q = q.options(sqlalchemy.orm.joinedload(target_cls.metrics))
+                        q = q.options(sqlalchemy.orm.joinedload(
+                            target_cls.metrics))
                         try:
-                            all_resources.extend(session.scalars(q).unique().all())
+                            all_resources.extend(
+                                session.scalars(q).unique().all())
                         except sqlalchemy.exc.ProgrammingError as e:
                             # NOTE(jd) This exception can happen when the
                             # resources and their resource type have been
--- a/gnocchi/rest/aggregates/api.py
+++ b/gnocchi/rest/aggregates/api.py
@@ -211,18 +211,21 @@ class MeasureGroup(object):
         self.measures.append(measure)
 
     def join_sequential_groups(self, group):
-        group.sort(key=lambda key: key['revision_start'])
+        group.sort(key=lambda key: key['search_start'])
         new_group = []
         last_it = None
         for it in group:
-            if last_it and it['revision_start'] == last_it['revision_end']:
-                last_it['revision_end'] = it['revision_end']
+            if last_it and it['search_start'] == last_it['search_start']:
+                last_it['search_en'] = it['search_en']
                 continue
 
             last_it = it
             new_group.append(it)
 
         return new_group
+
+    def __str__(self):
+        return "Group keys [%s]." % self.group_key
 
     def sum_groups_same_time_values(self):
         to_remove = []
@@ -303,10 +306,17 @@ class Grouper(object):
 
     def get_grouped_measures(self):
         resources = self.retrieve_resources_history()
-        self.grouped_response = self.get_measures(self.group(resources))
+        groups_to_process_metrics = self.group(resources)
+        LOG.debug("Groups [%s] to process metrics of resources [%s].",
+                  groups_to_process_metrics, resources)
+
+        self.grouped_response = self.get_measures(groups_to_process_metrics)
+        LOG.debug("Response grouped [%s] for resources [%s].",
+                  self.grouped_response, resources)
+
         response = self.format_response()
-        LOG.debug("[ Resources History: %s ]", resources)
-        LOG.debug("[ Response: %s ]", response)
+        LOG.debug("Response to be used after group by process and formatting: "
+                  "[%s]", response)
         return response
 
     def group(self, to_group):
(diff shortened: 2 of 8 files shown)


rafaelweingartner:
@tobias-urdin can you look into this one? It is a nice to have in Gnocchi.

rafaelweingartner:
we need to merge this one first: https://github.com/gnocchixyz/gnocchi/pull/1308

tobias-urdin:
@mergifyio rebase

gnocchi/indexer/sqlalchemy.py
```diff
@@ -1083,7 +1083,81 @@ def get_resource(self, resource_type, resource_id, with_metrics=False):
                 q = q.options(sqlalchemy.orm.joinedload('metrics'))
             return q.first()
 
-    def _get_history_result_mapper(self, session, resource_type):
+    def populate_filters_for_resource_history(
+            self, attribute_filter, attribute_filters_to_use={}):
+        """Extracts the filters for resource history table.
+
+        Extracts the filters that can be used in the resource history table to
+        apply in the aggregates query that we execute in the MySQL.
```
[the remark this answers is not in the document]
rafaelweingartner: I see. Thanks for the review! 

Fixed!

gnocchi/rest/aggregates/api.py
```diff
@@ -502,6 +535,15 @@ def post(self, start=None, stop=None, granularity=None,
...
                         attribute_filter=attr_filter,
                         sorts=sorts)
+
+                    if resources:
+                        LOG.debug(
+                            "Resources found [%s] with query filter [%s].",
+                            [{'original_resource_id': r.original_resource_id,
+                              'started_at': r.started_at,
+                              'ended_at': r.ended_at} for r in resources if
+                             resources], attr_filter)
```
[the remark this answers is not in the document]
rafaelweingartner: done

gnocchi/indexer/sqlalchemy.py
```diff
@@ -1083,7 +1083,81 @@ def get_resource(self, resource_type, resource_id, with_metrics=False):
...
+                    self.populate_filters_for_resource_history(
+                        value, attribute_filters_to_use)
+        else:
+            LOG.warning("Attribute filter [%s] is not of expected types"
+                        "[list or dict].", attribute_filter)
+
+    def build_filters_for_query(
+            self, valid_filters, start='started_at', end='ended_at'):
+        filters = []
+        if start in valid_filters and end in valid_filters:
```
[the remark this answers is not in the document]
rafaelweingartner: Fixed

gnocchi/indexer/sqlalchemy.py
```diff
@@ -1083,7 +1083,81 @@ def get_resource(self, resource_type, resource_id, with_metrics=False):
...
+                filters.append({
+                    "and": time_filter_to_apply
+                })
+        if 'project_id' in valid_filters:
+            project_filter = {'=': {'project_id': valid_filters['project_id']}}
+
+            if 'creator' in valid_filters:
+                filters.append({
+                    'or': [project_filter,
+                           {'like': {'creator': valid_filters['creator']}}]})
```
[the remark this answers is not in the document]
rafaelweingartner: I do not know. We just know that it needs to be this way. Some tests are affected if we try to use as you suggested. You can see at: https://github.com/gnocchixyz/gnocchi/actions/runs/5731177719/job/15531536852?pr=1307. 

We do not know if the creator filter is expected to be used together with the project_id, and so on. That is why we implemented this way.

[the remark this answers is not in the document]
rafaelweingartner: I will work on an improved solution to use all of the possible attributes.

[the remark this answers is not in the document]
rafaelweingartner: @chungg can you check the code now? I guess it is better now.

gnocchi/indexer/sqlalchemy.py
```diff
@@ -1101,8 +1194,50 @@ def _get_history_result_mapper(self, session, resource_type):
         s1 = sqlalchemy.select(history_cols.values())
         s2 = sqlalchemy.select(resource_cols.values())
         if resource_type != "generic":
+            engine = session.connection()
+
+            resource_history_filters_names = [
```
pedro-martins, line 1199: You could use the sqlalchemy to get the Resource history table columns instead of hardcoding them
```suggestion
            # Get the available columns for ResourceHistory table
            resource_history_cols = list(
                map(lambda column: column.name, sqlalchemy.inspect(
                    ResourceHistory).columns))
```

gnocchi/indexer/sqlalchemy.py
```diff
@@ -1101,8 +1194,50 @@ def _get_history_result_mapper(self, session, resource_type):
...
+            history_filters = self.extracts_filters_for_table(
+                attribute_filter,
+                allowed_keys_for_table=resource_history_filters_names)
+
+            LOG.debug("Filters to be used [%s] in query according to resource "
+                      "history columns [%s] extracted from [%s].",
+                      history_filters, resource_history_filters_names,
+                      attribute_filter)
+
+            resource_filter_names = [
```
pedro-martins, line 1213: Same here:
```suggestion
           # Get the available columns for Resource table
            resource_cols = list(
                map(lambda column: column.name, sqlalchemy.inspect(
                    Resource).columns))
```

rafaelweingartner:
Hello @chungg and @tobias-urdin I guess this patch is ready for your reviews again.

tobias-urdin:
> Hello @chungg and @tobias-urdin I guess this patch is ready for your reviews again.

I will check sometime this week but it's probably best to get @chungg review as well.

Please rebase on the now merged changes for SQLAlchemy 2.0 support, remove some duplicate commits that seems to be on here, squash commits and add some clear commit messages.

Thanks!

rafaelweingartner:
@tobias-urdin and @chungg rebase and squash done!

gnocchi/rest/aggregates/api.py
```diff
@@ -211,19 +211,22 @@ def add_measure(self, measure, date_map, start, stop):
...
     def join_sequential_groups(self, group):
-        group.sort(key=lambda key: key['revision_start'])
+        group.sort(key=lambda key: key['search_start'])
         new_group = []
         last_it = None
         for it in group:
-            if last_it and it['revision_start'] == last_it['revision_end']:
-                last_it['revision_end'] = it['revision_end']
+            if last_it and it['search_start'] == last_it['search_start']:
+                last_it['search_en'] = it['search_en']
```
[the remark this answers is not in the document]
rafaelweingartner: Nice catch! Fixed!

rafaelweingartner:
Hello guys, 
Do we need some other amendments in this patch?

gnocchi/rest/aggregates/api.py
```diff
@@ -567,6 +701,11 @@ def groupper(r):
 
         results = []
         for key, resources in itertools.groupby(resources, groupper):
+            resources = list(resources)
+
+            LOG.debug("Executing collection of metrics for group [%s] and "
+                      "resources [%s].", key, resources)
```
chungg, line 707:
```suggestion
            LOG.debug("Executing collection of metrics for group [%s] and "
                      "resources [%s].", key, list(resources))
```

gnocchi/indexer/sqlalchemy.py
```diff
@@ -1071,7 +1073,100 @@ def get_resource(self, resource_type, resource_id, with_metrics=False):
...
+                    LOG.debug("Replacing value [%s] in list with the sanitized"
+                              "value [%s] in its current position.",
+                              attribute, value_sanitized)
+                    value_index = attribute_filters_to_use.index(attribute)
+                    attribute_filters_to_use[value_index] = value_sanitized
+
+        elif is_value_dict:
+            all_keys = list(attribute_filter.keys())
+            for key in all_keys:
+                value = attribute_filter.get(key)
```
chungg, line 1126:
```suggestion
            for key, value in attribute_filter.items():
```

gnocchi/rest/aggregates/api.py
```diff
@@ -394,16 +456,14 @@ def truncate_measures(self):
...
-        measures_sum = 0
-        for measure in measures_list[:-1]:
-            measure_new_val = (measure.value *
-                               measure.usage_coefficient)
-            measures_sum += measure_new_val
+
+        for measure in measures_list:
+            measure_new_val = (measure.value * measure.usage_coefficient)
+            LOG.debug('Trucating measure [%s] value to [%s].',
+                      measure.__dict__, measure_new_val)
```
chungg, line 465: sorry, it's not clear to me how this change relates to your requirement to filter resources?

gnocchi/rest/aggregates/api.py
```diff
@@ -365,20 +391,56 @@ def get_measures(self, groups):
...
+        for key in all_keys:
+            data_map_key += "%s=%s," % (key, group.group_key[key])
+
+        data_map_key = data_map_key[0:len(data_map_key) - 1]
+
+        date_map_to_return = self.measures_date_map.setdefault(data_map_key, {})
+        LOG.debug("Date map [%s] found for key [%s].",
+                  date_map_to_return, data_map_key)
+
+        return date_map_to_return
```
chungg, line 443: this a copy/paste issue? everything after return in 432 will never run/

gnocchi/json.py
```diff
@@ -30,8 +30,7 @@ def to_primitive(obj):
     if isinstance(obj, datetime.datetime):
         return obj.isoformat()
     if isinstance(obj, numpy.datetime64):
-        # Do not include nanoseconds if null
-        return str(obj).rpartition(".000000000")[0] + "+00:00"
+        return numpy.datetime_as_string(obj, unit='s') + "+00:00"
```
chungg, line 33: don't think this is wrong but can you confirm the issue here? is this a separate bug you're fixing?

chungg (commented):
you probably don't need me to approve. i'm just getting familiar with this again.

i guess my only other feedback would be it'd be good to have a new test rather than just making sure existing ones don't break.

gnocchi/rest/aggregates/api.py
rafaelweingartner: Why would this change be better than the way it is coded now? If resources is an iterator, when we create the list here, it will be empty, and then when the resources variable is used, there will be nothing there.

gnocchi/indexer/sqlalchemy.py
rafaelweingartner: done.

gnocchi/rest/aggregates/api.py
rafaelweingartner: While troubleshooting the problem to filter the resources, we were having a lot of issues to understand what was going on due to the lack of logs. Therefore, we added logs in a few places to helps us understand what was going on. It does not help fixing the issues per se, but it helped us to understand where the slowness was.

rafaelweingartner: While troubleshooting the problem to filter the resources, we were having a lot of issues to understand what was going on due to the lack of logs. Therefore, we added logs in a few places to helps us understand what was going on. It does not help fixing the issues per se, but it helped us to understand where the slowness was.

gnocchi/rest/aggregates/api.py
rafaelweingartner: Yes, it is. Nice catch

gnocchi/json.py
rafaelweingartner: it is another bug that we discovered while fixing this issue of slowness. The date manipulation here is quite odd, which was causing some weird outputs if the nanoseconds part of the string had some value different from `.000000000`;  we just changed that to a better approach.

gnocchi/indexer/sqlalchemy.py
rafaelweingartner: This does not work. Therefore, I reverted:
- https://github.com/gnocchixyz/gnocchi/actions/runs/6278775791/job/17053206922?pr=1307

gnocchi/rest/aggregates/api.py
chungg: valid point :)

gnocchi/json.py
chungg: i see. have a slight preference to have this in a separate PR/commit.

gnocchi/indexer/sqlalchemy.py
chungg: i think it's because you put `for key, value in attribute_filter` instead of `for key, value in attribute_filter.items()`

gnocchi/rest/aggregates/api.py
chungg: oh i meant the change to measure_sum logic, not why you added the log. tbh, i don't remember why we ahve this measure_sum logic

rafaelweingartner: Ah, I see. This is something that was added when the implementation for the partial accounting of attributes was introduced (https://github.com/gnocchixyz/gnocchi/pull/1059). While working on this patch here, we were navigating the code, and trying to figure out what was going on. Therefore, we took the opportunity to add logs in the execution flows that were in need of some. 

As a consequence of that, we can save some operators going bald due to the lack of information for a troubleshooting.

gnocchi/json.py
rafaelweingartner: This is also something that we found while troubleshooting this issue here.

rafaelweingartner:
Thank you @chungg for your reviews! I have extract the code you suggested to a new PR at https://github.com/gnocchixyz/gnocchi/pull/1361.

Besides that, are we missing something else here?

rafaelweingartner:
@chungg and @tobias-urdin I have rebased the patch. Can you guys check it?

rafaelweingartner:
Hello @tobias-urdin, @chungg, @jd do you guys think that we need to do some more adjusts in the code, so it can get merged?

jd:
There is no way I could review this TBH, it's been so long. I'm ok to approve and merge it though as I don't want to block any improvement.
Anyone else wants to weight it before I do it?

rafaelweingartner:
> There is no way I could review this TBH, it's been so long. I'm ok to approve and merge it though as I don't want to block any improvement. Anyone else wants to weight it before I do it?

Thanks! 

BTW, we have been constantly improving Gnocchi for years so far, and we have more coming. Would it be possible for one of us to be come a committer with merge permissions in the project here?

gnocchi/rest/aggregates/api.py
```diff
@@ -211,19 +211,22 @@ def add_measure(self, measure, date_map, start, stop):
         self.measures.append(measure)
 
     def join_sequential_groups(self, group):
-        group.sort(key=lambda key: key['revision_start'])
```
tobias-urdin, line 214: Can you explain these changes.

tobias-urdin (changes requested):
I don't have the time to go through this in detail or test it, but just browsing through it and the fact that the testing is green makes me OK to just merge.

I do however have some remarks.

* The commit messages is very undescriptive and it looks like this PR is addressing more things in this PR than just the optimization itself 
*  The revision_start -> search_start etc, why was that done can you explain? Did we merge broken code with the group history PR before? If so, why is that not in a separate PR and bulked into this PR that is unrelated to the optimization itself.
* This adds a lot of LOG.debug calls, in pretty much the entire code path, I understand that might be good for debugging but is it really neccesary that we merge all that?

Overall it's kind of hard to understand what parts of this change is just cleanup in formatting, what fixes bugs, I would hope that would be cleaner in the future.

I'm all for giving our more access, my primary goal is still maintenance and keeping the project alive, especially to make sure it doesn't block anything in the OpenStack ecosystem so the project gets dropped from there (like the SQLAlchemy 2.0 migration).

I do really appreciate that you're taking the effort to address larger flaws in Gnocchi, like this one, I didn't even know about it until this PR, but I would hope that a high standard in commits is still kept even if the project is in maintenance mode.

jd:
> > There is no way I could review this TBH, it's been so long. I'm ok to approve and merge it though as I don't want to block any improvement. Anyone else wants to weight it before I do it?
> 
> Thanks!
> 
> BTW, we have been constantly improving Gnocchi for years so far, and we have more coming. Would it be possible for one of us to be come a committer with merge permissions in the project here?

I give you a soft-yes and will let @tobias-urdin and others interested to weigh in.

rafaelweingartner:
> I don't have the time to go through this in detail or test it, but just browsing through it and the fact that the testing is green (hoping that most things are tested that this touches) makes me OK to just merge.
> 
> I do however have some remarks.
> 
>     * The commit messages is very undescriptive and it looks like this PR is addressing more things in this PR than just the optimization itself
> 
>     * The revision_start -> search_start etc, why was that done can you explain? Did we merge broken code with the group history PR before? If so, why is that not in a separate PR and bulked into this PR that is unrelated to the optimization itself.
> 
>     * This adds a lot of LOG.debug calls, in pretty much the entire code path, I understand that might be good for debugging but is it really neccesary that we merge all that?
> 
> 
> Overall it's kind of hard to understand what parts of this change is just cleanup in formatting, what fixes bugs, I would hope that would be cleaner in the future.
> 
> I'm all for giving our more access, my primary goal is still maintenance and keeping the project alive, especially to make sure it doesn't block anything in the OpenStack ecosystem so the project doesn't get dropped from there (keeping up with things such as the SQLAlchemy 2.0 migration).
> 
> I do really appreciate that you're taking the effort to address larger flaws in Gnocchi, like this one, I didn't even know about it until this PR, but I would hope that a high standard in commits is still kept even if the project is in maintenance mode.

Let me start by saying that Gnocchi is very convoluted and quite complicated code-wise. The idea is simple, but the way it was coded, makes it very difficult to understand what is being done, and how to fix the issues we find or the extensions that need to be developed. We are trying to address that by adding logs, changing/creating new functions with more descriptive names, and so on. However, it is a task that is going to take much more time until we reach a state where things get better.

Regarding the addition of logs, it goes towards that situation.  It is very difficult to troubleshoot Gnocchi. Therefore, the addition of those logs will help people to work with Gnocchi in the future. Furthermore, they (the logs) do not cause any harm. You can use it in INFO log level (debug=false), and then you will not see those DEBUG logs.

Those flaws that nobody knows exist that we are fixing/patching come due to the fact that most OpenStack setups are not taking full advantage of Gnocchi (as far as we have seen). We are only facing them because we have a situation where we are putting pressure on Gnocchi to the limit without cleaning/removing things (the setup) every upgrade with a newer installation and so on. Moreover, we have a generic billing pipeline that receives data not just from OpenStack, but other systems that compose the cloud.

And, regarding the description. It is clear for people using Gnocchi that have some understanding of its inner workings. Every metric pushed to Gnocchi is created for a resource. Every time we create a resource, we have the create_at field, that is marked, and every time we update it, we have a revision created for it. When we do a search, via the Gnocchi search API, before this patch, the search was going through all resources in the system. However, that is costly. We can filter the resources affected by the search using the created_at/revision_start ended_at/revision_end fields when they are provided by the user calling the API. This is the optimization that we are doing here. Gnocchi was filtering the results in Python, and loading all measure, for all resources for all timestamps. What we propose is to filter the result-set right away in the DB.

Regarding the quality standard, can you describe/show us what needs to be improved? Then, we can work to address it and make the standard to be in the same level as you expect it to be.

rafaelweingartner:
> > > There is no way I could review this TBH, it's been so long. I'm ok to approve and merge it though as I don't want to block any improvement. Anyone else wants to weight it before I do it?
> > 
> > 
> > Thanks!
> > BTW, we have been constantly improving Gnocchi for years so far, and we have more coming. Would it be possible for one of us to be come a committer with merge permissions in the project here?
> 
> I give you a soft-yes and will let @tobias-urdin and others interested to weigh in.

I do not know what that means =). However, even if we do not deserve a committer role. We just need to understand what are the requirements before a merge can be done.

We have more feature and optimizations to push to Gnocchi, and sometimes the time it takes to get them in is quite long.

jd:
I think the root problem is that the project is barely maintained at this stage. All the people who worked on Gnocchi years ago and wrote most of that code are gone in the wild.
This means nobody is smart enough or has the time to review all that code, especially because it's a huge change that seems (I did not dig through it) to address multiple problems at the same time.

I'm not concerned by the project's fate, maintenance, etc, at this stage, so I'm not going to block anything. As I said, I'm giving a soft yes meaning "I think it's better to have this merged because you seem to care about the project more than anyone else, but also, I'm not smart enough to say yay that works let's merge it."

rafaelweingartner:
> I think the root problem is that the project is barely maintained at this stage. All the people who worked on Gnocchi years ago and wrote most of that code are gone in the wild. This means nobody is smart enough or has the time to review all that code, especially because it's a huge change that seems (I did not dig through it) to address multiple problems at the same time.
> 
> I'm not concerned by the project's fate, maintenance, etc, at this stage, so I'm not going to block anything. As I said, I'm giving a soft yes meaning "I think it's better to have this merged because you seem to care about the project more than anyone else, but also, I'm not smart enough to say yay that works let's merge it."

Thanks for the message! I understand what you mean. We are trying to reduce the scope for every new PR here. Gnocchi is a very interesting system. I mean, it has an interesting design and idea/concept, that we see a lot of potential. However, code-wise it has problems that we are working to address and to improve.

rafaelweingartner:
Hello guys, 
Do you want us to execute some other changes in this PR before it is merged?

rafaelweingartner:
I was revisiting this PR, and the only file that was changed, and is not directly related to the patch proposed here is the following:
- `doc/source/rest.j2`.

I can extract that is needed, but it was an opportunity to improve a documentation, and we just did it. All of the rest is related to the changeset proposed. All of the logs were needed/helped us to understand the situation, and they are being added in DEBUG log level. Therefore, if one does not want them, it is a matter of disabling DEBUG.

rafaelweingartner:
> Yes if you have changes that can be extracted, do this, that'll help for sure!

Done. The documentation improvement was extracted to: https://github.com/gnocchixyz/gnocchi/pull/1373

rafaelweingartner:
Hello guys, can we merge this one? Or, do you think that we need some further improvements here?

The part of the documentation was already extracted to https://github.com/gnocchixyz/gnocchi/pull/1373.

gnocchi/rest/aggregates/api.py
tobias-urdin: I still don't understand the revision_* -> search_* changes?

I also think this PR is too broad in what it does and should be cleaned up and be more minimal. For example the optimization to _get_history_result_mapper and changes to the aggregate/grouping code.

rafaelweingartner: The problem is that I do not understand what you do not understand.

This PR is addressing a situation with Gnocchi aggregates API. This API can present slowness with time, due to the growing number of resources and revisions in large scale OpenStack Cloud environments.

The situation we reported happens is some cases when we have a very large number of resources in the cloud environment. The slowness happens because when the aggregate API is called, it does not apply the filters sent in the API directly in the MySQL query. It only applies some filtering. Therefore, it will bring more elements in the response than what is needed. For instance, the start/stop filters are not applied, and others as well, which makes Gnocchi to manipulate all the dataset (resources) in the MySQL database and then the filtering is executed in Python. In some situations we have seen that databases almost broke/got in a stuck/stale state.

To cope with that situation, we are proposing an optimization to improve the query executed by Gnocchi in MySQL, and apply the filtering right away. By that, I mean, applying the filters directly in the MySQL query. Therefore, we reduce the result-set size and the number of elements manipulated by MySQL query. After applying the patch, we noticed a reduction of 5-6 folds in the response time for the Gnocchi aggregates API for large scale cloud environments.

What don't you understand?

Do you need an example of query that we normally execute? This is one example of query that we used to discover the reported issue:
```
/v1/aggregates?details=False&granularity=3600&start=2023-08-11T00%3A00%3A00%2B00%3A00&stop=2023-08-11T01%3A00%3A00%2B00%3A00&use_history=True&groupby=id&groupby=display_name&groupby=flavor_id&groupby=flavor_name&groupby=user_id&groupby=project_id&groupby=revision_start&groupby=availability_zone&groupby=image_ref&groupby=flavor_vcpus&groupby=flavor_ram&groupby=operating_system_name&groupby=operating_system_distro&groupby=operating_system_type&groupby=operating_system_version&groupby=mssql_version" -H "Content-Type: application/json" -H "Accept: application/json, */*" -H "X-Auth-Token: ${ACCESS_TOKEN_KEYSTONE}" -d '{"operations":["aggregate","max",["metric","dynamic_pollster.compute.services.instance.status","max"]],"search":{"and":[{"=":{"type":"instance"}},{"=":{"project_id":"<project_id>"}}]},"resource_type":"instance"}'
```
The above query is one that might be generated by CloudKitty, for instance.

As you can see, with a lot of groupbys, MySQL will have trouble to process the large dataset generated by Gnocchi queries, as they are not properly applying the filters sent in the API request.

Furthermore, can you explain me why you think that the PR is too broad?

jd (approved):
As said, I'm ok with that since I don't think it'd be fair to block it just on the basis we don't have time to review it.

rafaelweingartner:
> As said, I'm ok with that since I don't think it'd be fair to block it just on the basis we don't have time to review it.

Thanks for your support! If there is anything else missing here, please let me know. We have more patches to open for Gnocchi, but we are doing one by one, to avoid managing/dealing with conflicts in multiple fronts.

PR #1310, opened 2023-07-17
Add more fill variants

### LINKED ISSUE #1266: Ability to set non-static fill value, based on existing data in series for aggregates

```
$ gnocchi --version
gnocchi 7.0.7
$ gnocchi server version
+---------+-------+
| Field   | Value |
+---------+-------+
| version | 4.4.2 |
+---------+-------+
```
on OpenStack Xena release.

I'm trying to get percentage of CPU and memory utilization on a scale from 1 to 100 (for CPU that means, that percent of utilization doesn't depend on a number of vCPUs in VM). 

For `cpu` I have this data in series:
```
$ gnocchi aggregates --resource-type instance --start 2022-07-27T12:00:00 '(/ (metric cpu rate:mean) 60000000000)' id=c1081247-90cb-471b-b1e3-3a927de2e042
+----------------------------------------------------+---------------------------+-------------+----------------------+
| name                                               | timestamp                 | granularity |                value |
+----------------------------------------------------+---------------------------+-------------+----------------------+
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T12:00:00+00:00 |       300.0 |               0.0696 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T12:05:00+00:00 |       300.0 |  0.06923333333333333 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T12:10:00+00:00 |       300.0 |  0.06856666666666666 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T12:15:00+00:00 |       300.0 |  0.06993333333333333 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T12:20:00+00:00 |       300.0 |               0.0723 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T12:25:00+00:00 |       300.0 |               0.0712 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T12:30:00+00:00 |       300.0 |  0.06833333333333333 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T12:35:00+00:00 |       300.0 |  0.06629166666666667 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T12:40:00+00:00 |       300.0 |  0.06883333333333333 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T12:45:00+00:00 |       300.0 |  0.07066666666666667 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T12:50:00+00:00 |       300.0 |  0.07073333333333333 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T12:55:00+00:00 |       300.0 |  0.07306666666666667 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T13:00:00+00:00 |       300.0 |               0.0712 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T13:05:00+00:00 |       300.0 |  0.07343333333333334 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T13:10:00+00:00 |       300.0 |               0.0745 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T13:15:00+00:00 |       300.0 |  0.07703333333333333 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T13:20:00+00:00 |       300.0 |  0.07103333333333334 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T13:25:00+00:00 |       300.0 |               0.0712 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T13:30:00+00:00 |       300.0 |  0.07246666666666667 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T13:35:00+00:00 |       300.0 |               0.0376 |
| c1081247-90cb-471b-b1e3-3a927de2e042/cpu/rate:mean | 2022-07-27T13:40:00+00:00 |       300.0 | 0.037333333333333336 |
+----------------------------------------------------+---------------------------+-------------+----------------------+
```
For `vcpus` I have this:
```
$ gnocchi aggregates --resource-type instance --start 2022-07-27T12:00:00 '(metric vcpus mean)' id=c1081247-90cb-471b-b1e3-3a927de2e042
+-------------------------------------------------+---------------------------+-------------+-------+
| name                                            | timestamp                 | granularity | value |
+-------------------------------------------------+---------------------------+-------------+-------+
| c1081247-90cb-471b-b1e3-3a927de2e042/vcpus/mean | 2022-07-27T12:00:00+00:00 |       300.0 |   4.0 |
| c1081247-90cb-471b-b1e3-3a927de2e042/vcpus/mean | 2022-07-27T13:00:00+00:00 |       300.0 |   4.0 |
+-------------------------------------------------+---------------------------+-------------+-------+
```

Now, when I try to aggregate those two series to get CPU utilization on a scale from 1 to 100, I get data points in the resulting set only for intersections of series where points are with the same `timestamp`:
```
$ gnocchi aggregates --resource-type instance --start 2022-07-27T12:00:00 '(/ (aggregate mean (/ (metric cpu rate:mean) (metric vcpus mean))) 60000000000))' id=c1081247-90cb-471b-b1e3-3a927de2e042
+------------+---------------------------+-------------+--------+
| name       | timestamp                 | granularity |  value |
+------------+---------------------------+-------------+--------+
| aggregated | 2022-07-27T12:00:00+00:00 |       300.0 | 0.0174 |
| aggregated | 2022-07-27T13:00:00+00:00 |       300.0 | 0.0178 |
+------------+---------------------------+-------------+--------+
```
This works as expected, as this should be fixed by using `fill` parameter, but I can only set a static value for `fill`, whereas I need missing values from the `vcpus` series to be filled with "same value as previous non-null/none/NaN" value, for this kind of aggregate to work properly. 

Previously, this could be achieved by creating same percentage metrics with `transform` in `sink`, which got deprecated and consequetively removed. 

I expect a way to get CPU and memory utilization in percent for all available data point in `cpu` and `memory.usage` series on a scale from 0 to 100 by using calculated, based on existing data in series, value for `fill`.

Some time ago there were attempts to fix this missing bit within ceilometer:
* https://review.opendev.org/c/openstack/ceilometer/+/799963 for `cpu_util`
* https://review.opendev.org/c/openstack/ceilometer/+/597054 for `memory_util`
and the first one of these raised a valid question for why does percentage would be stored as cumulative value, so Gnocchi looks like proper place to add this functionality. missing

dbalagansky:
I'm not an expert in numpy, so this was borrowed from SO: https://stackoverflow.com/a/54508024.

I probably should rename this from "previous" to "similar" (or maybe some other better word?), as this thing covers the case when first point in resulting series could be NaN with backwards fill in addition to the standard forward fill. 

Fixes #1266

### DIFF
--- a/doc/source/rest.j2
+++ b/doc/source/rest.j2
@@ -793,12 +793,17 @@ Backfill
 ~~~~~~~~
 
 The ability to fill in missing points from a subset of time series is supported
-by specifying a `fill` value. Valid fill values include any float, `dropna` or
-`null`. In the case of `null`, Gnocchi will compute the aggregation using only
-the existing points. `dropna` is like `null` but remove NaN from the result.
-The `fill` parameter will not backfill timestamps which contain no points in
-any of the time series. Only timestamps which have datapoints in at least one
-of the time series is returned.
+by specifying a `fill` value. Valid fill values include any float, `dropna`,
+`null`, `ffill`, `bfill` `full_ffill` or `full_bfill`. In the case of `null`,
+Gnocchi will compute the aggregation using only the existing points. `dropna` is
+like `null` but remove NaN from the result. `ffill` fills NaN measures in one
+metric with previous non-NaN value, `bfill` fills NaN measures with next non-NaN
+value, so if the metric starts (or, in case of `bfill`, ends) with NaNs, those
+wouldn't change and would be excluded from the resulting set, just like `dropna`
+does. To fill the remaining NaN values, producing a metric with as much
+timestamps as there are in all metrics combined, one can use `full_ffill` and
+`full_bfill` variants. In the case of `full_ffill`, it applies a forward then backward
+fill and for `full_bfill`, a backward then forward fill.
 
 {{ scenarios['get-aggregates-by-metric-ids-fill']['doc'] }}
 
@@ -1039,12 +1044,17 @@ expects 100% overlap. If this percentage is not reached, an error is returned.
    boundary to the first or last timestamp common across all series.
 
 The ability to fill in missing points from a subset of time series is supported
-by specifying a `fill` value. Valid fill values include any float, `dropna` or
-`null`. In the case of `null`, Gnocchi will compute the aggregation using only
-the existing points. `dropna` is like `null` but remove NaN from the result.
-The `fill` parameter will not backfill timestamps which contain no points in
-any of the time series. Only timestamps which have datapoints in at least one
-of the time series is returned.
+by specifying a `fill` value. Valid fill values include any float, `dropna`,
+`null`, `ffill`, `bfill` `full_ffill` or `full_bfill`. In the case of `null`,
+Gnocchi will compute the aggregation using only the existing points. `dropna` is
+like `null` but remove NaN from the result. `ffill` fills NaN measures in one
+metric with previous non-NaN value, `bfill` fills NaN measures with next non-NaN
+value, so if the metric starts (or, in case of `bfill`, ends) with NaNs, those
+wouldn't change and would be excluded from the resulting set, just like `dropna`
+does. To fill the remaining NaN values, producing a metric with as much
+timestamps as there are in all metrics combined, one can use `full_ffill` and
+`full_bfill` variants. In the case of `full_ffill`, it applies a forward then backward
+fill and for `full_bfill`, a backward then forward fill.
 
 {{ scenarios['get-across-metrics-measures-by-metric-ids-fill']['doc'] }}
 
--- a/gnocchi/rest/aggregates/processor.py
+++ b/gnocchi/rest/aggregates/processor.py
@@ -157,7 +157,7 @@ def aggregated(refs_and_timeseries, operations, from_timestamp=None,
             return_inverse=True)
 
         # create nd-array (unique series x unique times) and fill
-        filler = (numpy.NaN if fill in [None, 'null', 'dropna']
+        filler = (numpy.NaN if fill in [None, 'null', 'dropna', 'ffill', 'bfill', 'full_ffill', 'full_bfill']
                   else fill)
         val_grid = numpy.full((len(series[sampling]), len(times)), filler)
         start = 0
@@ -165,6 +165,21 @@ def aggregated(refs_and_timeseries, operations, from_timestamp=None,
             size = len(split)
             val_grid[i][indices[start:start + size]] = split['values']
             start += size
+
+        if fill == "ffill":
+            val_grid = utils.forward_fill(val_grid)
+
+        if fill == "bfill":
+            val_grid = utils.backward_fill(val_grid)
+
+        if fill == "full_ffill":
+            val_grid = utils.forward_fill(val_grid)
+            val_grid = utils.backward_fill(val_grid)
+
+        if fill == "full_bfill":
+            val_grid = utils.backward_fill(val_grid)
+            val_grid = utils.forward_fill(val_grid)
+
         values = val_grid.T
 
         if fill is None:
@@ -200,7 +215,7 @@ def aggregated(refs_and_timeseries, operations, from_timestamp=None,
         output = {"aggregated": []}
         for sampling in sorted(result, reverse=True):
             granularity, times, values, references = result[sampling]
-            if fill == "dropna":
+            if fill in ("dropna", "ffill", "bfill", "full_ffill", "full_bfill"):
                 pos = ~numpy.logical_or(numpy.isnan(values[0]),
                                         numpy.isinf(values[0]))
                 v = values[0][pos]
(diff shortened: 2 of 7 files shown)


tobias-urdin (commented):
Please add testing to show functional behaviour, a release note would be good to, see [1] for using reno to add a release note.

[1] https://docs.openstack.org/reno/latest/user/usage.html#creating-new-release-notes

tobias-urdin:
@Mergifyio rebase

chungg (commented):
sort of surprised we didn't have this but you should update doc/source/rest.j2 as well.

not blocker but:
- `similar` is a odd keyword. not sure if there's existing keyword we can copy from other libs? bidirectional seems more accurate but maybe too long?
- would be good to have `forward` and `backward` as well. i suspect `forward` is the most common option.

dbalagansky:
> * would be good to have `forward` and `backward` as well. i suspect `forward` is the most common option.
> 
I've been thinking about these for some time and the problem with plain `backward` and `forward` fill variants is that, given this series: `[NaN, NaN, 1, 2, NaN, 4, 5, NaN, NaN]`:
* doing just `forward` fill would result in: `[NaN, NaN, 1, 2, 2, 4, 5, 5, 5]`
* doing just `backward` fill would result in: `[1, 1, 1, 2, 4, 4, 5, NaN, NaN]`

So, as you can see, in some cases it wouldn't be enough to do `forward` or `backward` fill to produce fully populated series (without `NaN`'s) and, maybe, it would be a good idea to:
* add something like `full_ffill` for doing `forward` than `backward` (for overwriting first two `NaN`'s in the example above after normal `forward` fill) and same thing with `full_bfill`: `backward` than `forward` fill. This way there would be no need for `similar` or `bidirectional` fill variants as the one that is implemented in this PR is just `full_ffill` one
* add `ffill` and `bfill`, which would leave `NaN`'s in the resulting series alone

chungg:
>     * add something like `full_ffill` for doing `forward` than `backward` (for overwriting first two `NaN`'s in the example above after normal `forward` fill) and same thing with `full_bfill`: `backward` than `forward` fill. This way there would be no need for `similar` or `bidirectional` fill variants as the one that is implemented in this PR is just `full_ffill` one
> 
>     * add `ffill` and `bfill`, which would leave `NaN`'s in the resulting series alone

yeah, this makes sense to me. i didn't even consider the differences in order of your full fill solution. i'm ok with the naming. don't have anything better to suggest

dbalagansky:
`py311-postgresql-file-upgrade-from-4.4` fail doesn't look related (and it does pass for me locally). 

This should be ready for review, hope I didn't miss any points, but feel free to point them out :)

One more thing that came to my mind during the final tests was that this probably should interact with resampling down (e.g., doing aggregates `resample mean 5` on a metric with measurements every minute) by creating additional measurements and filling them, but that sound like a separate PR.

doc/source/rest.j2
```diff
@@ -793,12 +793,17 @@ Backfill
...
+`null`, `ffill`, `bfill` `full_ffill` or `full_bfill`. In the case of `null`,
+Gnocchi will compute the aggregation using only the existing points. `dropna` is
+like `null` but remove NaN from the result. `ffill` fills NaN measures in one
+metric with previous non-NaN value, `bfill` fills NaN measures with next non-NaN
+value, so if the metric starts (or, in case of `bfill`, ends) with NaNs, those
+wouldn't change and would be excluded from the resulting set, just like `dropna`
+does. To fill those remaning NaN too, producing resulting metric with as much
+timestamps as there are in all metrics combined, one can use `full_ffill` and
+`full_bfill` variants, which, in case of `full_ffill` do forward, than backward
+fill or `backward`, than `forward` fill for `full_bfill`.
```
chungg, line 806:
```suggestion
does. To fill the remaining NaN values, producing a metric with as much
timestamps as there are in all metrics combined, one can use `full_ffill` and
`full_bfill` variants. In the case of `full_ffill`, it applies a forward then backward
fill and for `full_bfill`, a backward then forward fill.
```

doc/source/rest.j2
```diff
@@ -1039,12 +1044,17 @@ expects 100% overlap. If this percentage is not reached, an error is returned.
...
```
chungg, line 1057:
```suggestion
does. To fill the remaining NaN values, producing a metric with as much
timestamps as there are in all metrics combined, one can use `full_ffill` and
`full_bfill` variants. In the case of `full_ffill`, it applies a forward then backward
fill and for `full_bfill`, a backward then forward fill.
```

chungg (commented):
lgtm. minor wording update

chungg (approved):

tobias-urdin (approved):

PR #1311, opened 2023-07-18
Fix easy dependency

rafaelweingartner:


rafaelweingartner:
No, we do not. I only created because the jobs in 4.4 were using it. However, you removed in your patch. That is why we do not need it anymore.

PR #1315, opened 2023-07-31
Fix multiple CI issues

tobias-urdin:
This fixes multiple CI issues, see the commits.


rafaelweingartner:
@tobias-urdin that is what we have noticed, that MySQL version is causing some issues with Gnocchi.

PR #1317, opened 2023-08-01
db: add sqlalchemy 2.0 support

tobias-urdin:
This adds support for SQLAlchemy 2.0 but does not solve the removal of autocommit which will need to be handled in the future.


gnocchi/indexer/sqlalchemy.py
```diff
@@ -521,9 +524,9 @@ def fill_null_attribute_values(self, engine, name, rt, session,
                     f = QueryTransformer.build_filter(
```
stephenfin, line 524: GitHub won't let me leave comments but I think the use of `str(url)` on line 301 above also needs to be updated. By default this will sanitise the URL and replace passwords with `***`. To avoid this, you need to use `url.render_as_string(hide_password=False)` instead. We encountered this in [nova also](https://review.opendev.org/c/openstack/nova/+/880669/4/nova/cmd/manage.py#761).

gnocchi/indexer/sqlalchemy.py
```diff
@@ -1334,10 +1350,15 @@ class QueryTransformer(object):
...
     def _handle_multiple_op(cls, engine, table, op, nodes):
-        return op(*[
+        args = [
             cls.build_filter(engine, table, node)
             for node in nodes
-        ])
+        ]
+
+        if len(args) == 0:
+            return op(sqlalchemy.true())
```
stephenfin, line 1359: nit: You could just add this anyway. It's effectively a noop:

```
return op(sqlalchemy.true(), *args)
```

stephenfin (changes requested):
Only spotted one potential issue. The rest looks good though we'll surely spot other issues when we resolve the autocommit thing and use SQLAlchemy 2.x (and oslo.db 13.x).

Out of curiosity, would it be worth enabling warnings in our test like we've done for most OpenStack projects?

gnocchi/indexer/sqlalchemy.py
tobias-urdin: Don't follow if you talk about the code in dress_url() or about __str__(), the code in __str__() could most definitely be cleaned with that approach but the other function is nothing exposed.

I will look into it and perhaps fix that in a separate PR.

tobias-urdin: Oh reading it again, now I understand. I will check that.

stephenfin:
> > Only spotted one potential issue. The rest looks good though we'll surely spot other issues when we resolve the autocommit thing and use SQLAlchemy 2.x (and oslo.db 13.x).
> > Out of curiosity, would it be worth enabling warnings in our test like we've done for most OpenStack projects?
> 
> Thanks, yeah the autocommit part will really suck when it hits there is no real control of the transactions so more cleanup required there, the testing right now actually uses SQAlchemy 2.x and newer oslo.db already because we have unpinned dependencies that is not restricted by any upper-constraints.

Oh really? I wonder how it's working though since autocommit isn't available in SQLAlchemy 2.x (at least outside of [driver-level implementations](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#driver-level-autocommit-remains-available)). Weird.

> 
> I did look into enabling warnings with a warnings fixtures to the tests, used it while doing this work but didn't get any noticeable warnings except for the autocommit IIRC.
> 
> I opted for unpinning SQLAlchemy pinning and instead add a separate SQLAlchemy 1.4 job to make sure we don't break >=1.4<2 users.

Sounds good.

tobias-urdin:
> > > Only spotted one potential issue. The rest looks good though we'll surely spot other issues when we resolve the autocommit thing and use SQLAlchemy 2.x (and oslo.db 13.x).
> > > Out of curiosity, would it be worth enabling warnings in our test like we've done for most OpenStack projects?
> > 
> > 
> > Thanks, yeah the autocommit part will really suck when it hits there is no real control of the transactions so more cleanup required there, the testing right now actually uses SQAlchemy 2.x and newer oslo.db already because we have unpinned dependencies that is not restricted by any upper-constraints.
> 
> Oh really? I wonder how it's working though since autocommit isn't available in SQLAlchemy 2.x (at least outside of [driver-level implementations](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#driver-level-autocommit-remains-available)). Weird.

I wish I had an answer for that but really I'm just glad it even works, I'm hoping nothing major is overlooked and that it actually works as well, it's still unreleased code so atleast not a catastrophe if this were to break someones CI.
> 
> > I did look into enabling warnings with a warnings fixtures to the tests, used it while doing this work but didn't get any noticeable warnings except for the autocommit IIRC.
> > I opted for unpinning SQLAlchemy pinning and instead add a separate SQLAlchemy 1.4 job to make sure we don't break >=1.4<2 users.
> 
> Sounds good.

Do you have any other feedback for this or do you approve?

stephenfin:
Nope, no other feedback. This looks good to me as-is now.

PR #1325, opened 2023-08-03
Drop use of build_sphinx distutils command

stephenfin:
Drop use of the `build_sphinx` distutils command, which was dropped in Sphinx 7.x. distutils itself is deprecated and will be removed in Python 3.12.

While we're here, we also fix up some other issues with tox and the setup files in general. The commit messages hold more information about this.


stephenfin:
Depends on #1329.

tobias-urdin:
@stephenfin thanks, this all looks good to me except for removal of `docs-gnocchi-web` tox env. I think that's used somehow to roll out https://gnocchi.osci.io but I don't have access to that so I'm not sure. It's also building that but ignoring warnings to not fail I guess(?).

Can we leave that in and fix the conflict and I'll pull this in asap to unblock sphinx7, thanks!

stephenfin:
> @stephenfin thanks, this all looks good to me except for removal of `docs-gnocchi-web` tox env. I think that's used somehow to roll out [gnocchi.osci.io](https://gnocchi.osci.io) but I don't have access to that so I'm not sure. It's also building that but ignoring warnings to not fail I guess(?).
> 
> Can we leave that in and fix the conflict and I'll pull this in asap to unblock sphinx7, thanks!

Done. I think the failure is unrelated but I can't rerun it to check.

.github/workflows/gnocchi.yml
```diff
@@ -21,7 +21,6 @@ jobs:
       matrix:
         env:
           - docs
-          - docs-gnocchi-web
```
tobias-urdin, line 24: Please also add back this

.mergify.yml
```diff
@@ -2,7 +2,6 @@ queue_rules:
   - name: default
     conditions:
     - check-success=doc (docs)
-    - check-success=doc (docs-gnocchi-web)
```
tobias-urdin, line 5: And these two

tobias-urdin (changes requested):

tobias-urdin:
Figured out it's handled from here: https://gitlab.com/osci/community-cage-infra-ansible/-/blob/master/roles/gnocchi_builder/templates/gnocchi_builder

.mergify.yml
stephenfin: Done

.github/workflows/gnocchi.yml
stephenfin: Done

tobias-urdin:
Sorry for all the resistance @stephenfin – please rebase after revert https://github.com/gnocchixyz/gnocchi/pull/1343

stephenfin:
@tobias-urdin Anything else needed here or can we merge this (I'd like to get it off my TODO list :smile:)

tobias-urdin:
@stephenfin Thanks!

PR #1329, opened 2023-08-03
CI: Use Python 3.11 for tox

stephenfin:
Recent versions of tox do not support Python 3.6.

### DIFF
--- a/images/Dockerfile.ci
+++ b/images/Dockerfile.ci
@@ -60,7 +60,7 @@ RUN locale-gen $LANG
 
 #NOTE(sileht): Upgrade python dev tools
 RUN python3 -m pip install -U pip tox virtualenv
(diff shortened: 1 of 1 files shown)


tobias-urdin:
We're still on Ubuntu 20.04 so a lot of deadsnakes stuff to get newer versions so probably need python3.11-distutils or something in the image as well. I've opted for starting to move it to 3.9 first in the SQALchemy 2.0 PR https://github.com/gnocchixyz/gnocchi/pull/1317/commits/15fcaee34b7f98d9ff931d4798f589fdd7d58e1d

tobias-urdin:
We'll distutils is gone in 3.11 so my bad, I guess we can get it to 3.9 atleast for now.

PR #1361, opened 2023-12-01
Fix time format for datetime objects

rafaelweingartner:
The date manipulation in Gnocchi to generate the response  is quite odd, which is causing some weird outputs if the nanoseconds part of the string has some value different from .000000000; this patch is changing that to a better approach.

This was extract from https://github.com/gnocchixyz/gnocchi/pull/1307#discussion_r1330867035

### DIFF
--- a/gnocchi/json.py
+++ b/gnocchi/json.py
@@ -30,8 +30,7 @@ def to_primitive(obj):
     if isinstance(obj, datetime.datetime):
         return obj.isoformat()
     if isinstance(obj, numpy.datetime64):
-        # Do not include nanoseconds if null
-        return str(obj).rpartition(".000000000")[0] + "+00:00"
+        return numpy.datetime_as_string(obj, unit='s') + "+00:00"
     if isinstance(obj, numpy.timedelta64):
         return obj / numpy.timedelta64(1, 's')
     if isinstance(obj, datetime.timedelta):


pedro-martins:
@rafaelweingartner, this PR https://github.com/gnocchixyz/gnocchi/pull/1362 fixes the errors, you can rebase your PR from it, it should fix the tests errors

rafaelweingartner:
> @rafaelweingartner, this PR #1362 fixes the errors, you can rebase your PR from it, it should fix the tests errors

Thanks! I already reviewed, and approved!

gnocchi/json.py
```diff
@@ -30,8 +30,7 @@ def to_primitive(obj):
     if isinstance(obj, datetime.datetime):
         return obj.isoformat()
     if isinstance(obj, numpy.datetime64):
-        # Do not include nanoseconds if null
-        return str(obj).rpartition(".000000000")[0] + "+00:00"
+        return numpy.datetime_as_string(obj, unit='s') + "+00:00"
```
chungg, line 33:
```suggestion
        return str(obj.astype('datetime64[s]')) + "+00:00"
```
is a bit faster but this serializer doesn't seem particularly efficient anyways so it doesn't matter.

chungg (approved):

gnocchi/json.py
rafaelweingartner: I see. I do find the way it is right now easier to read, and more direct to understand. However, if you guys prefer this other way, I can change it.

chungg: lgtm too. datetime_as_string seems to be part of stable api from what i can tell.

PR #1366, opened 2023-12-20
Remove the use of the future module

thomasgoirand:
This module is buggy with Python 3.11 and 3.12, and is being actively removed from distributions (like Debian and Ubuntu). The absolute_import things are remaining of the past, in the Python 2.7 age. Let's get rid of them and remove future from setup.cfg.


tobias-urdin (approved):

tobias-urdin:

PR #1370, opened 2024-01-24
Remove example policy files

kajinamit:
The default policy file has been changed to policy.yaml in 4.5, so we can remove this file assuming it may no longer be used.

Also, the policy.yaml file is empty, thus is ineffective. Let's drop the file and then oslo.policy skips loading the default policy file (and use the rules implemented in code).


tobias-urdin:
I wonder if this is used by any packging, probably not, and it's also a mess since we including data files in a python module. I'm fine just deleting this on master and hopefully nothing will break, thought?

kajinamit:
It seems packages in distributions(such as RDO or Canonical) installs policy.json but we can drop it because the file is not loaded by gnocchi by default.

I just noticed we may not need even poliy.yaml, because it's just an empty file and oslo.policy skips loading the file in case the file does not exist. I've added another commit to remove the file as well. I'll update RDO spec file once this is merged (and probably when they bump gnocchi version). Ubuntu packages may be taken care by canonical folks.

kajinamit:
We merged https://review.rdoproject.org/r/c/openstack/gnocchi-distgit/+/51386 for RDO which stops copying policy files into /etc/gnocchi. Gnocchi still loads the files in /usr/share/python3.9/site-packages/gnocchi but once this change is merged that part will be also skipped.

tobias-urdin (approved):

PR #1371, opened 2024-02-04
Drop redundant required property

kajinamit:
An option doesn't need to be marked required if it has a default value.


rafaelweingartner (approved):
LGTM

tobias-urdin (approved):

PR #1374, opened 2024-02-29
Redis: Fix sentinel fallbacks with IPv6 address

kajinamit:
This fixes how the logic parses host and port from sentinel fallbacks when the host is IPv6 address. We should not use bare split by ':' because ':' is used within address representations like [::1]:6379.


tobias-urdin (approved):

tobias-urdin:

PR #1383, opened 2024-04-13
Remove unused dependency

kajinamit:
The trollius library is required only in Python <= 3.4, but currently Python 3.8 is the minimum supported version.

### DIFF
--- a/gnocchi/statsd.py
+++ b/gnocchi/statsd.py
@@ -12,13 +12,10 @@
 # implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import asyncio
 import itertools
 import uuid
 
-try:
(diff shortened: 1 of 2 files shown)


rafaelweingartner:
Isn't this used here?
>     import trollius as asyncio

kajinamit:
> Isn't this used here?
> 
> > ```
> > import trollius as asyncio
> > ```
> 
> https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/statsd.py#L21

Oops. Good catch. I removed the ImportError fallback because the asyncio module is available in all of the python versions currently supported (python >= 3.8) .

tobias-urdin (approved):

PR #1384, opened 2024-04-17
Fix Gnocchi manager upgrade command for File

rafaelweingartner:
When running the upgrade command manually, we noticed that the "tmp" folder in the "conf.file_basepath" is not created, which will lead to an error later in the execution flow.

### DIFF
--- a/gnocchi/incoming/file.py
+++ b/gnocchi/incoming/file.py
@@ -35,17 +35,15 @@ class FileStorage(incoming.IncomingDriver):
         self.basepath = conf.file_basepath
         self.basepath_tmp = os.path.join(self.basepath, 'tmp')
 
-        # We need to make sure that the 'tmp' path exists; otherwise,
-        # we will have an issue when running the method "set_storage_settings"
-        # that created the temporary JSON file under the 'tmp' folder.
-        utils.ensure_paths([self.basepath_tmp])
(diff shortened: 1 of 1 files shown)


pedro-martins:
It looks good to me. It will help in future deploys using file driver, so the gnocchi-upgrade command will create the tmp dir, not need to do it manually.

gnocchi/incoming/file.py
```diff
@@ -35,6 +35,11 @@ def __init__(self, conf, greedy=True):
         self.basepath = conf.file_basepath
         self.basepath_tmp = os.path.join(self.basepath, 'tmp')
 
+        # We need to make sure that the 'tmp' path exists; otherwise,
+        # we will have an issue when running the method "set_storage_settings"
+        # that created the temporary JSON file under the 'tmp' folder.
+        utils.ensure_paths([self.basepath_tmp])
```
chungg, line 41: this seems to already exists in the upgrade method? i guess we need to have it run before the super() call.
i'd lean more to fixing it there so we don't create folders whenever we create instance.

rafaelweingartner: I see. Done.

Thanks for the review!

chungg (approved):
thanks!

PR #1386, opened 2024-05-28
Automatically detect deleted resources

rafaelweingartner:
This patch is created on top of https://github.com/gnocchixyz/gnocchi/pull/1385, https://github.com/gnocchixyz/gnocchi/pull/1381,  https://github.com/gnocchixyz/gnocchi/pull/1387, and https://github.com/gnocchixyz/gnocchi/pull/1388. Therefore, we need to merge them first. Then, the extra commits here will disappear.

While executing some Gnocchi optimizations (https://github.com/gnocchixyz/gnocchi/pull/1307), we noticed that some deleted/removed resources do not have the "ended_at" field with a datetime. This can cause slowness with time, as more and more "zombie" resources are left there, and this has a direct impact in the MySQL queries executed with the aggregates API.

This patch introduces a new parameter called `metric_inactive_after`, which defines for how long a metric can go without receiving new data points until we consider it as inactive. Then, when all metrics of a resource are in inactive state, we can mark/consider the resource as removed.

### DIFF
--- a/gnocchi/chef.py
+++ b/gnocchi/chef.py
@@ -18,9 +18,11 @@ import hashlib
 
 import daiquiri
 import random
+import datetime
 
 from gnocchi import carbonara
 from gnocchi import indexer
+from gnocchi import utils
 
 LOG = daiquiri.getLogger(__name__)
 
@@ -50,6 +52,79 @@ class Chef(object):
         # which means, database connector.
         self.index = index
         self.storage = storage
+
+    def resource_ended_at_normalization(self, metric_inactive_after):
+        """Marks resources as ended at if needed.
+
+        This method will check all metrics that have not received new
+        datapoints after a given period. The period is defined by
+        'metric_inactive_after' parameter. If all metrics of resource are in
+        inactive state, we mark the ended_at field with a timestmap. Therefore,
+        we consider that the resource has ceased existing.
+
+        In this process we will handle only metrics that are considered as
+        inactive, according to `metric_inactive_after` parameter. Therefore,
+        we do not need to lock these metrics while processing, as they are
+        inactive, and chances are that they will not receive measures anymore.
+        """
+
+        momment_now = utils.utcnow()
+        momment = momment_now - datetime.timedelta(
+            seconds=metric_inactive_after)
+
+        inactive_metrics = self.index.list_metrics(
+            attribute_filter={"<": {
+                "last_measure_timestamp": momment}},
+            resource_policy_filter={"==": {"ended_at": None}}
+        )
+
+        LOG.debug("Inactive metrics found for processing: [%s].",
+                  inactive_metrics)
+
+        metrics_by_resource_id = {}
+        for metric in inactive_metrics:
+            resource_id = metric.resource_id
+            if metrics_by_resource_id.get(resource_id) is None:
+                metrics_by_resource_id[resource_id] = []
+
+            metrics_by_resource_id[resource_id].append(metric)
+
+        for resource_id in metrics_by_resource_id.keys():
+            if resource_id is None:
+                LOG.debug("We do not need to process inactive metrics that do "
+                          "not have resource. Therefore, these metrics [%s] "
+                          "will be considered inactive, but there is nothing "
+                          "else we can do in this process.",
+                          metrics_by_resource_id[resource_id])
+                continue
+            resource = self.index.get_resource(
+                "generic", resource_id, with_metrics=True)
+            resource_metrics = resource.metrics
+            resource_inactive_metrics = metrics_by_resource_id.get(resource_id)
+
+            all_metrics_are_inactive = True
+            for m in resource_metrics:
+                if m not in resource_inactive_metrics:
+                    all_metrics_are_inactive = False
+                    LOG.debug("Not all metrics of resource [%s] are inactive. "
+                              "Metric [%s] is not inactive. The inactive "
+                              "metrics are [%s].",
+                              resource, m, resource_inactive_metrics)
+                    break
+
+            if all_metrics_are_inactive:
+                LOG.info("All metrics [%s] of resource [%s] are inactive."
+                         "Therefore, we will mark it as finished with an"
+                         "ended at timestmap.", resource_metrics, resource)
+                if resource.ended_at is not None:
+                    LOG.debug(
+                        "Resource [%s] already has an ended at value.", resource)
+                else:
+                    LOG.info("Marking ended at timestamp for resource "
+                             "[%s] because all of its metrics are inactive.",
+                             resource)
+                    self.index.update_resource(
+                        "generic", resource_id, ended_at=momment_now)
 
     def clean_raw_data_inactive_metrics(self):
         """Cleans metrics raw data if they are inactive.
--- a/gnocchi/chef.py
+++ b/gnocchi/chef.py
@@ -17,8 +17,8 @@
 import hashlib
 
 import daiquiri
-import random
 import datetime
+import random
 
 from gnocchi import carbonara
 from gnocchi import indexer
--- a/gnocchi/chef.py
+++ b/gnocchi/chef.py
@@ -82,26 +82,26 @@ class Chef(object):
         LOG.debug("Inactive metrics found for processing: [%s].",
                   inactive_metrics)
 
-        metrics_by_resource_id = {}
+        inactive_metrics_by_resource_id = {}
         for metric in inactive_metrics:
             resource_id = metric.resource_id
-            if metrics_by_resource_id.get(resource_id) is None:
-                metrics_by_resource_id[resource_id] = []
+            if inactive_metrics_by_resource_id.get(resource_id) is None:
+                inactive_metrics_by_resource_id[resource_id] = []
 
-            metrics_by_resource_id[resource_id].append(metric)
+            inactive_metrics_by_resource_id[resource_id].append(metric)
 
-        for resource_id in metrics_by_resource_id.keys():
+        for resource_id in inactive_metrics_by_resource_id.keys():
             if resource_id is None:
                 LOG.debug("We do not need to process inactive metrics that do "
                           "not have resource. Therefore, these metrics [%s] "
                           "will be considered inactive, but there is nothing "
                           "else we can do in this process.",
-                          metrics_by_resource_id[resource_id])
+                          inactive_metrics_by_resource_id[resource_id])
                 continue
             resource = self.index.get_resource(
                 "generic", resource_id, with_metrics=True)
             resource_metrics = resource.metrics
-            resource_inactive_metrics = metrics_by_resource_id.get(resource_id)
+            resource_inactive_metrics = inactive_metrics_by_resource_id.get(resource_id)
 
             all_metrics_are_inactive = True
             for m in resource_metrics:
--- a/gnocchi/chef.py
+++ b/gnocchi/chef.py
@@ -14,8 +14,8 @@
 # implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import hashlib
 import datetime
+import hashlib
 
 import daiquiri
 import random
--- a/gnocchi/chef.py
+++ b/gnocchi/chef.py
@@ -82,12 +82,9 @@ class Chef(object):
         LOG.debug("Inactive metrics found for processing: [%s].",
                   inactive_metrics)
 
-        inactive_metrics_by_resource_id = {}
+        inactive_metrics_by_resource_id = collections.defaultdict(list)
         for metric in inactive_metrics:
             resource_id = metric.resource_id
-            if inactive_metrics_by_resource_id.get(resource_id) is None:
-                inactive_metrics_by_resource_id[resource_id] = []
-
             inactive_metrics_by_resource_id[resource_id].append(metric)
 
         for resource_id in inactive_metrics_by_resource_id.keys():
--- a/gnocchi/chef.py
+++ b/gnocchi/chef.py
@@ -66,6 +66,7 @@ class Chef(object):
         inactive, according to `metric_inactive_after` parameter. Therefore,
         we do not need to lock these metrics while processing, as they are
         inactive, and chances are that they will not receive measures anymore.
+        Moreover, we are only touching metadata, and not the actual data.
         """
 
         momment_now = utils.utcnow()
--- a/gnocchi/chef.py
+++ b/gnocchi/chef.py
@@ -15,11 +15,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import hashlib
+import datetime
 
 import daiquiri
-import datetime
 import random
 
+from collections import defaultdict
 from gnocchi import carbonara
 from gnocchi import indexer
 from gnocchi import utils
@@ -69,20 +70,20 @@ class Chef(object):
         Moreover, we are only touching metadata, and not the actual data.
         """
 
-        momment_now = utils.utcnow()
-        momment = momment_now - datetime.timedelta(
+        moment_now = utils.utcnow()
+        moment = moment_now - datetime.timedelta(
             seconds=metric_inactive_after)
 
         inactive_metrics = self.index.list_metrics(
             attribute_filter={"<": {
-                "last_measure_timestamp": momment}},
+                "last_measure_timestamp": moment}},
             resource_policy_filter={"==": {"ended_at": None}}
         )
 
         LOG.debug("Inactive metrics found for processing: [%s].",
                   inactive_metrics)
 
-        inactive_metrics_by_resource_id = collections.defaultdict(list)
+        inactive_metrics_by_resource_id = defaultdict(list)
         for metric in inactive_metrics:
             resource_id = metric.resource_id
             inactive_metrics_by_resource_id[resource_id].append(metric)
@@ -122,7 +123,7 @@ class Chef(object):
                              "[%s] because all of its metrics are inactive.",
                              resource)
                     self.index.update_resource(
-                        "generic", resource_id, ended_at=momment_now)
+                        "generic", resource_id, ended_at=moment_now)
 
     def clean_raw_data_inactive_metrics(self):
         """Cleans metrics raw data if they are inactive.
--- a/gnocchi/indexer/alembic/versions/f89ed2e3c2ec_create_last_measure_timestamp_column.py
+++ b/gnocchi/indexer/alembic/versions/f89ed2e3c2ec_create_last_measure_timestamp_column.py
@@ -19,11 +19,10 @@ Revises: 18fff4509e3e
 Create Date: 2024-04-24 09:16:00
 """
 
+import datetime
 from alembic import op
 
 import sqlalchemy
-
-from sqlalchemy.sql import func
 
 # revision identifiers, used by Alembic.
 revision = 'f89ed2e3c2ec'
@@ -33,7 +32,9 @@ depends_on = None
 
 
 def upgrade():
-    op.add_column(
-        "metric", sqlalchemy.Column(
+    sql_column = sqlalchemy.Column(
             "last_measure_timestamp", sqlalchemy.DateTime,
-            nullable=False, server_default=func.current_timestamp()))
+            nullable=False, default=datetime.datetime.now(datetime.UTC),
+            server_default=sqlalchemy.sql.func.current_timestamp())
+    op.add_column(
+        "metric", sql_column)
--- a/gnocchi/indexer/alembic/versions/f89ed2e3c2ec_create_last_measure_timestamp_column.py
+++ b/gnocchi/indexer/alembic/versions/f89ed2e3c2ec_create_last_measure_timestamp_column.py
@@ -13,21 +13,21 @@
 #    under the License.
 #
 
-"""Create metric truncation status column
-
-Revision ID: 18fff4509e3e
-Revises: 04eba72e4f90
+"""Create last measure push timestamp column
+Revision ID: f89ed2e3c2ec
+Revises: 18fff4509e3e
 Create Date: 2024-04-24 09:16:00
-
 """
 
 from alembic import op
 
 import sqlalchemy
 
+from sqlalchemy.sql import func
+
 # revision identifiers, used by Alembic.
-revision = '18fff4509e3e'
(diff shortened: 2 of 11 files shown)


run-upgrade-tests.sh
```diff
@@ -97,12 +97,34 @@ new_version=$(python setup.py --version)
...
 # Gnocchi 3.1 uses basic auth by default
 export OS_AUTH_TYPE=gnocchi-basic
 export GNOCCHI_USER=$GNOCCHI_USER_ID
 
+# Pifpaf configures the logs for the standard output. Therefore, depending
+# on the operating system, the standard output has some buffer size, which
+# needs to be released. Otherwise, the logs stop to be writen, and the
+# execution of the code is "frozen", due to the lack of buffer in the
+# process output. To work around that, we can read the buffer, and dump it
+# into a lof file. Then, we can cat the log file content at the end of the
```
GutoVeronezi, line 110:
```suggestion
# into a log file. Then, we can cat the log file content at the end of the
```

gnocchi/chef.py
```diff
@@ -45,9 +46,164 @@ class Chef(object):
...
+
+                    self.index.update_needs_raw_data_truncation(metric.id)
+                else:
+                    LOG.debug("No raw measures found for metric [%s] for "
+                              "cleanup.", metric.id)
+                    self.index.update_needs_raw_data_truncation(metric.id)
+
+                metric_lock.release()
+            except Exception:
+                LOG.error("Unable to lock metric [%s] for cleanup.",
```
pedro-martins, line 200: Is that an error or a warning? I mean an already locked metric is expected sometimes right? in this case It would try to run again some time after right? So if it is something expected, a warning should be ok. 
And if any error happens after getting the lock? should we release it in the except block? or removing it inside the try block, and put it in a finally block which always get executed, even in errors, what do you think?

gnocchi/chef.py
```diff
@@ -45,9 +46,164 @@ class Chef(object):
...
+
+        metrics_by_resource_id = {}
+        for metric in inactive_metrics:
+            resource_id = metric.resource_id
+            if metrics_by_resource_id.get(resource_id) is None:
+                metrics_by_resource_id[resource_id] = []
+
+            metrics_by_resource_id[resource_id].append(metric)
+
+        for resource_id in metrics_by_resource_id.keys():
```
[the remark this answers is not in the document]
rafaelweingartner: All resource types "extend" the "generic" type. That is why I hardcoded this one.

gnocchi/chef.py
```diff
@@ -45,9 +46,164 @@ class Chef(object):
...
+        )
+
+        LOG.debug("Metrics [%s] found to execute the raw data cleanup.",
+                  metrics_to_clean)
+
+        for metric in metrics_to_clean:
+            LOG.debug("Executing the raw data cleanup for metric [%s].",
+                      metric)
+            try:
+                metrid_id = metric.id
```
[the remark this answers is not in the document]
rafaelweingartner: done

gnocchi/chef.py
rafaelweingartner: I see. I agree with you.

rafaelweingartner:
Thanks @pedro-martins !

gnocchi/chef.py
```diff
@@ -45,9 +47,170 @@ class Chef(object):
...
+                        metric, ts.serialize())
+
+                    self.index.update_needs_raw_data_truncation(metric.id)
+                else:
+                    LOG.debug("No raw measures found for metric [%s] for "
+                              "cleanup.", metric.id)
+                    self.index.update_needs_raw_data_truncation(metric.id)
+
+                metric_lock.release()
+            except Exception:
```
pedro-martins, line 206: Here you are catching all exceptions, even the exceptions not related to metric lock, I think that calling the metric_lock.release() inside this except block is necessary to avoid getting resources stuck in the lock state (even though the lock has an expiration time, not sure it happens here)

rafaelweingartner: I moved to the finally. I think it would be more or less the same. If there is a lock exception, the metric_lock is None. Therefore, I am handling such situation in the code. What do you think?

pedro-martins: I agree, using a finally block and moving all the locking releases to it would be better than doing it in the end of the method and in the exception handling.

gnocchi/chef.py
```diff
@@ -51,6 +53,80 @@ def __init__(self, coord, incoming, index, storage):
...
+                          "else we can do in this process.",
+                          metrics_by_resource_id[resource_id])
+                continue
+            resource = self.index.get_resource(
+                "generic", resource_id, with_metrics=True)
+            resource_metrics = resource.metrics
+            resource_inactive_metrics = metrics_by_resource_id.get(resource_id)
+
+            all_metrics_are_inactive = True
+            for m in resource_metrics:
```
pedro-martins, line 107: it wouldnt be simple if we just compare the size of resource_metrics and resource_inactive metrics? if they are the same, then we can clean, if not, then we ignore it.

gnocchi/opts.py
```diff
@@ -73,7 +72,7 @@ def __getitem__(self, key):
   but not chunked encoding (InfluxDB)
 * http-socket/socket: support chunked encoding, but require a upstream HTTP
   Server for HTTP/1.1, keepalive and HTTP protocol correctness.
-""")
+"""),
```
pedro-martins, line 75: do we need this comma?

run-upgrade-tests.sh
```diff
@@ -107,7 +107,7 @@ export GNOCCHI_USER=$GNOCCHI_USER_ID
 # needs to be released. Otherwise, the logs stop to be writen, and the
 # execution of the code is "frozen", due to the lack of buffer in the
 # process output. To work around that, we can read the buffer, and dump it
-# into a lof file. Then, we can cat the log file content at the end of the
+# into a log file. Then, we can cat the log file content at the end of the
```
pedro-martins, line 110: :)

pedro-martins (commented):
Hi Rafael, nice feature, I just leave some comments here, but the code overall is good to me. Thanks for this patch.

gnocchi/chef.py
```diff
@@ -51,6 +53,80 @@ def __init__(self, coord, incoming, index, storage):
...
+            resource_policy_filter={"==": {"ended_at": None}}
+        )
+
+        LOG.debug("Inactive metrics found for processing: [%s].",
+                  inactive_metrics)
+
+        metrics_by_resource_id = {}
+        for metric in inactive_metrics:
+            resource_id = metric.resource_id
+            if metrics_by_resource_id.get(resource_id) is None:
```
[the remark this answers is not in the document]
rafaelweingartner: done

gnocchi/chef.py
```diff
@@ -51,6 +53,80 @@ def __init__(self, coord, incoming, index, storage):
...
+                metrics_by_resource_id[resource_id] = []
```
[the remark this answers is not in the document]
rafaelweingartner: A resource cannot have multiple metrics with the same definition. Therefore, the use of Set would not make much difference, right?

gnocchi/chef.py
rafaelweingartner: It feels risky. I find it more secure to actually check if the metrics are indeed the same.

gnocchi/opts.py
rafaelweingartner: no

rafaelweingartner:
Thanks @pedro-martins for your review!

gnocchi/chef.py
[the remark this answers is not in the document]
pedro-martins: yes, if each list have only a few elements, not hundreds, then a list is better. I just asked it to check if the number of metrics by resource is huge or not, but if it is not, then I see no problem using lists.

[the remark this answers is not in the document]
rafaelweingartner: I see. It can be huge the number of metrics, but the biggest one I have seen was about 30. Therefore, not so much.

gnocchi/chef.py
pedro-martins: I can understand your point about safety, but if you are collecting both `resource_metrics` and `inactive_metrics` from the same source (same DB table) I cant see a scenario where matching both sizes wouldn't fit this goal. Maybe I am not seeing something that you saw. 

It was just a suggestion to reduce code complexity and improve a little the runtime, nothing critical.

gnocchi/chef.py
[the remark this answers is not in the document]
pedro-martins: yea, we guess we can leave it as list then, if it is not so usual having lots of metrics by resource, it does not make sense using sets.

gnocchi/chef.py
rafaelweingartner: Yes, I load them from the DB. However, they are loaded in different moments in the code. Therefore, there is no 100% that there is no other manipulation happening in between the different loads/processing. I mean, it is very unlikely, but possible.

pedro-martins: I see, but both solutions would fail in some edge cases (like new measures for inactive_metrics while this process is running). 

The size comparing solution would fail also if someone deletes a metric and creates a new metric while this process is running (which is less likely than new measures coming), but in the end, it would only mark this resource as ended, not delete it indeed, and as it receives any new measurements, it would get back to active, right? I guess you did it to deal with such edge cases.

I see no problem with your solution, it is good to me as well, I think both solutions would work fine, I just thought that reducing complexity would be something nice. :slightly_smiling_face:

rafaelweingartner: exactly that. And even if the resource is marked as "deleted", as soon as a metric receives new datapoints, it is "undeleted".

gnocchi/chef.py
```diff
@@ -51,6 +53,80 @@ def __init__(self, coord, incoming, index, storage):
...
+                          inactive_metrics_by_resource_id[resource_id])
+                continue
+            resource = self.index.get_resource(
+                "generic", resource_id, with_metrics=True)
+            resource_metrics = resource.metrics
+            resource_inactive_metrics = inactive_metrics_by_resource_id.get(resource_id)
+
+            all_metrics_are_inactive = True
+            for m in resource_metrics:
+                if m not in resource_inactive_metrics:
```
chungg, line 108: could just make this a set comparison instead?

gnocchi/storage/__init__.py
```diff
@@ -688,6 +688,26 @@ def _map_compute_splits_operations(bound_timeserie):
             if metric.needs_raw_data_truncation:
                 indexer_driver.update_needs_raw_data_truncation(metric.id)
 
+            # Mark when the metric receives its latest measures
+            indexer_driver.update_last_measure_timestmap(metric.id)
```
chungg, line 692: i'm wondering if there's a way to do this without updating indexer db if metric_inactive_after functionality is not enabled? seems like a performance hit as we're adding additional db call

gnocchi/opts.py
```diff
@@ -172,7 +171,16 @@ def list_opts():
                        default=10000,
                        min=1,
                        help="Number of metrics that should be deleted "
-                       "simultaneously by one janitor.")
+                            "simultaneously by one janitor."),
+            cfg.IntOpt('metric_inactive_after',
```
chungg, line 175: there's some ambiguity with what time ended_at value is set to.
- the ended_at value is dependent on when the janitor runs. this will lag actual metric_inactive_after value by an unknown amount.
- we update last_measure_timestamp based on when metricd worker processes metric. this means this will also lag an unknown amount when actual last measure was captured and/or sent/received by gnocchi.

not sure but is it better to mark ended_at with latest last_measure_timestamp + metric_inactive_after or does precision not really matter?

gnocchi/chef.py
```diff
@@ -17,10 +17,12 @@
 import hashlib
 
 import daiquiri
+import datetime
```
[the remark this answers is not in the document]
rafaelweingartner: ok, done.

gnocchi/chef.py
```diff
@@ -51,6 +53,80 @@ def __init__(self, coord, incoming, index, storage):
...
+        In this process we will handle only metrics that are considered as
+        inactive, according to `metric_inactive_after` parameter. Therefore,
+        we do not need to lock these metrics while processing, as they are
+        inactive, and chances are that they will not receive measures anymore.
+        Moreover, we are only touching metadata, and not the actual data.
+        """
+
+        momment_now = utils.utcnow()
+        momment = momment_now - datetime.timedelta(
+            seconds=metric_inactive_after)
```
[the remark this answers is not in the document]
rafaelweingartner: done.

gnocchi/chef.py
rafaelweingartner: Yes, we can.  What is the gain? I mean, there would probably be some gain if the set comparison is using some sort of hash. However, we would loose the option to log the first metric we found that is not receiving measures anymore. 

If you really think that we need to use a set comparison I can change that, but right now to have the log showing the metric that make us identify the resource as not inactive, I would prefer this way.

gnocchi/storage/__init__.py
rafaelweingartner: Yes, it is an additional call. However, nothing that costly. I mean, this is an async process anyways. This extra cost is worth when we see the benefits that this is bringing to the code base.

gnocchi/opts.py
rafaelweingartner: yes, there is. I mean, it is something that does not need to be precise. What matter is not the seconds when the metric is considered as inactive, but that all of them are inactivated. 

Also, for the resource, if we want to do "last_measure_timestamp + metric_inactive_after", from which metric we take the "last_measure_timestamp"? A resource can have multiple. We can always take the last one that became inactive, and so on. However, I do not see what sort of benefits this would bring.

rafaelweingartner:
@chungg thanks for your review! I have added the code changes you suggested, and there are some remarks open, which I would like to hear your response before closing them.

gnocchi/chef.py
chungg: you're right, considering size, it'd be negligible performance impact. that said you could still log the set difference and capture all the inactive rather just first one. i'm ok with this.

gnocchi/storage/__init__.py
chungg: well there's this one and then the resource lookup call below. i'm thinking if it makes sense to move the resource logic below to janitor process maybe so this step just focuses on processing measures data.

i think that way, the janitor (or something else) could technically just make a single db query which unset ended_at for all resources where ended_at not null and has metrics with last_update_timestamp > now - ttl?

if the above works, i imagine most of logic in resource_ended_at_normalization could be done as similar db query as well?

gnocchi/opts.py
chungg: i had thought historically, gnocchi has strayed from handling time as of "now"/wall time. that said, i can't find any reference to this explicitly so i guess it doesn't matter.

if you don't care about precision, i'm ok with it. i can't imagine it'd be controversial to change this behaviour in future if that is preference.

rafaelweingartner:
Hello @chungg is there something else that needs to be addressed here?

gnocchi/storage/__init__.py
chungg: don't know if the above was considered or it makes sense? tbh. i wrote it so long ago i don't remember if it was already concluded to be dumb/invalid.

rafaelweingartner: That was considered, but to make the code cleaner and more intuitive, we decided for this approach. If you think something else is better we could try to work together with you to change a bit the logic. However, so far, this patch is in use for over 6 months already and everything is so far so good.

rafaelweingartner:
Hello guys, 
Is there something missing here to move on and merge it?

This patch is the base for some further optimizations that we are doing and that we would like to propose upstream.

gnocchi/indexer/alembic/versions/f89ed2e3c2ec_create_last_measure_timestamp_column.py
```diff
@@ -0,0 +1,39 @@
...
+down_revision = '18fff4509e3e'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "metric", sqlalchemy.Column(
+            "last_measure_timestamp", sqlalchemy.DateTime,
+            nullable=False, server_default=func.current_timestamp()))
```
chungg, line 39: i don't know if it makes sense to make this nullable. it's possible that this is disabled at first and gets a default value and is enabled later and immediately all of the resources can potentially get set to ended (until next update).

chungg (dismissed):
i'm ok with this although i think it'd be better if the update logic was in sql rather than python. something like following but in orm. could also be done later.

```
update resource r set ended_at = %s where not exists (select 1 from metric m where m.resource_id=a.id AND m.last_measure_timestamp >= %s) and r.ended_at is null;
```

gnocchi/indexer/alembic/versions/f89ed2e3c2ec_create_last_measure_timestamp_column.py
rafaelweingartner: Exactly. Thanks for the review!

rafaelweingartner:
Hello guys, 
the failing tests do not seem to be related to the patch itself. Do you have an idea on how to fix it? @chungg  and @tobias-urdin ?

When I run here in my machine, it works :smile:

Callum027:
> Hello guys, the failing tests do not seem to be related to the patch itself. Do you have an idea on how to fix it? @chungg and @tobias-urdin ?
> 
> When I run here in my machine, it works 😄

I haven't done any investigation into the cause, but it seems like there's some kind of race condition for some specific tests. It's not ideal, but I've just been force-pushing unchanged commits on my PRs until the tests pass.

rafaelweingartner:
> > Hello guys, the failing tests do not seem to be related to the patch itself. Do you have an idea on how to fix it? @chungg and @tobias-urdin ?
> > When I run here in my machine, it works 😄
> 
> I haven't done any investigation into the cause, but it seems like there's some kind of race condition for some specific tests. It's not ideal, but I've just been force-pushing unchanged commits on my PRs until the tests pass.

So it is not just me :)

Callum027:
Looking at the code, it looks like there's a separate `metricd` thread that processes incoming measures in an infinite loop with a 0.1 second delay betwen the iterations.

https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/tests/functional/fixtures.py#L264-L279

100ms is actually a decently long amount of time due to the speed at which these tests are run. The particular tests that fail are probably running their checks before the metricd threads has processed them, and just fail instead of rechecking.

rafaelweingartner:
> Looking at the code, it looks like there's a separate `metricd` thread that processes incoming measures in an infinite loop with a 0.1 second delay betwen the iterations.
> 
> https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/tests/functional/fixtures.py#L264-L279
> 
> 100ms is actually a decently long amount of time due to the speed at which these tests are run. The particular tests that fail are probably running their checks before the metricd threads has processed them, and just fail instead of rechecking.

Cool, that gave me an idea. What do you think of a config like this one then in the tests?
https://github.com/gnocchixyz/gnocchi/pull/1386/commits/8fc168f03d1418b670748c771de6029c37f61e2a

This would repeat the test if it fails, and wait 1 seconds. Therefore, we could guarantee that the MetricD will already have been executed in the next try.

Callum027:
> > Looking at the code, it looks like there's a separate `metricd` thread that processes incoming measures in an infinite loop with a 0.1 second delay betwen the iterations.
> > https://github.com/gnocchixyz/gnocchi/blob/master/gnocchi/tests/functional/fixtures.py#L264-L279
> > 100ms is actually a decently long amount of time due to the speed at which these tests are run. The particular tests that fail are probably running their checks before the metricd threads has processed them, and just fail instead of rechecking.
> 
> Cool, that gave me an idea. What do you think of a config like this one then in the tests? [8fc168f](https://github.com/gnocchixyz/gnocchi/commit/8fc168f03d1418b670748c771de6029c37f61e2a)
> 
> This would repeat the test if it fails, and wait 1 seconds. Therefore, we could guarantee that the MetricD will already have been executed in the next try.

That's a great idea. There's normally two tests that are affected by that race condition so if we could retry them when they fail, that should solve it nicely without changing the current behaviour much.

rafaelweingartner:
Thanks for the help @Callum027. @chungg and @tobias-urdin  what do you think about this patch now?

Callum027:
Hi @chungg and @tobias-urdin , any chance of getting this and the other PRs I also have proposed looked at sometime soon?

Callum027:
Hi @rafaelweingartner, I have a question about this proposal.

Can a resource be "un-ended" if we need to start supplying metrics for that resource again (e.g. from Ceilometer)?

The particular use case I'm thinking about is Swift containers. The resource ID for a Swift container in Gnocchi is composed of its project ID and container name, which of course can be reused if you create a new container with the same name. Using this change to clean up old resources is quite beneficial I think, but if you're storing metrics for Swift container in your Gnocchi deployment I'm not sure this would be usable as-is.

rafaelweingartner:
> Hi @rafaelweingartner, I have a question about this proposal.
> 
> Can a resource be "un-ended" if we need to start supplying metrics for that resource again (e.g. from Ceilometer)?
> 
> The particular use case I'm thinking about is Swift containers. The resource ID for a Swift container in Gnocchi is composed of its project ID and container name, which of course can be reused if you create a new container with the same name. Using this change to clean up old resources is quite beneficial I think, but if you're storing metrics for Swift container in your Gnocchi deployment I'm not sure this would be usable as-is.

Yes, this usecase is considered. See: https://github.com/gnocchixyz/gnocchi/pull/1386/files#diff-9c436196247ad6c3c12d0b085d6e2ae7b57c54684a298457fdcf74c9ff3ac63eR700.

We also added this process due to other situations we have seen in the past, such as bad Ceilometer configurations, which led the system to loose monitoring data for almost a month, and so on. Therefore, with an implementation like this one, we would mark the resources as "finished", but then when the collection resumes, we would mark it as "alive" again.

tobias-urdin:
I will merge this next week unless there is any objections from previous reviewers @chungg @Callum027 @pedro-martins

PR #1387, opened 2024-05-28
Add main to CLI classes

rafaelweingartner:
When developing to Gnocchi, it is handy to run Gnocchi from Pycharm. To achieve that, if we have the main methods already in the CLI modules, one can easily start Gnocchi directly from Pycharm in Debug mode. Therefore, this patch introduces such option

### DIFF
--- a/gnocchi/cli/manage.py
+++ b/gnocchi/cli/manage.py
@@ -105,3 +105,7 @@ def change_sack_size():
     s.remove_sacks()
     LOG.info("Creating new %d sacks", conf.sacks_number)
     s.upgrade(conf.sacks_number)
+
+
+if __name__ == '__main__':
+    upgrade()
--- a/gnocchi/cli/api.py
+++ b/gnocchi/cli/api.py
@@ -114,3 +114,7 @@ def api():
         args.extend(["-H", os.getenv("VIRTUAL_ENV", ".")])
 
     return os.execl(uwsgi, uwsgi, *args)
+
+
+if __name__ == '__main__':
+    api()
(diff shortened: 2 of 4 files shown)


gnocchi/cli/manage.py
```diff
@@ -105,3 +105,7 @@ def change_sack_size():
     s.remove_sacks()
     LOG.info("Creating new %d sacks", conf.sacks_number)
     s.upgrade(conf.sacks_number)
+
+
+if __name__ == '__main__':
+    upgrade()
```
chungg, line 111: i don't really have any concerns with proposal. to be nitpicky, it is a bit opaque that manage.py would upgrade. also, the commands in this file are ones you wouldn't want to trigger by accidental. if not critical to your goal, i would probably skip at least this file.

rafaelweingartner: I see. What if we create an "upgrade.py", that has the main and calls the upgrade from manage.py? Then, we avoid this kind of issue, and make the module explicit.

chungg: this is just me being paranoid but i've destroyed/corrupted the wrong db before because i wasn't thinking or i had command in my terminal history.

this is to minimise chance of someone triggering upgrade by accident. granted you still need to have a conf file and i don't know think this is that much more unsafe than having gnocchi-upgrade cmd installed. so not 100% against it but with my ops hat on, this is not idiot-proof. :)

 i don't have better suggestion if i'm honest so i will approve.

chungg (approved):

gnocchi/cli/manage.py
rafaelweingartner: ok. Thanks @chungg for the review!

pedro-martins (approved):
It seems ok to me, having these main entries to run in some IDE can be useful.

tobias-urdin (approved):

rafaelweingartner:
@tobias-urdin, thanks for the support here!

PR #1390, opened 2024-06-18
Change code to support numpy to 2.0

rafaelweingartner:


setup.cfg
```diff
@@ -29,7 +29,7 @@ scripts =
     bin/gnocchi-api
 
 install_requires =
-    numpy>=1.14.0
+    numpy>=2.0
```
chungg, line 32: might be better to keep this as 1.14  to minimise requirement  conflicts depending how people install this. pp.nan exists back then too.

rafaelweingartner: done

pedro-martins:
I see the changes you did to adapt the code to the new numpy version, but where are you changing it to 2.0?

rafaelweingartner:
> I see the changes you did to adapt the code to the new numpy version, but where are you changing it to 2.0?

I am not. As @chungg recommended there might be people using numpy < 2.0, which also supports the new syntax. The thing is that in numpy > 2.0 the old NaN is not supported anymore.

rafaelweingartner:
> I didnt see this review, github hid it from me. Now I saw the chung suggestion.
> 
> Maybe if you could change the MR title, it could help new reviewers as now you are not updating numpy anymore but making the code compatible with new versions.

Done. I changed the title. What do you think?

pedro-martins (approved):
LGTM

chungg (approved):

PR #1408, opened 2024-10-01
Use oslo.utils method to parse server format

kajinamit:
The oslo.utils library is already required via some of the current dependencies (eg. tooz). Use the utility method from it to parse host:port format instead of carrying our own method.


tobias-urdin:
Makes sense, we already depend on a lot of oslo so I don't see an issue adding utils. Please rebase and I'll merge when green, thanks!

tobias-urdin (approved):

PR #1411, opened 2024-10-03
Get rid of pkg_resources

kajinamit:
... because it was removed in Python 3.12 [1].

[1] https://docs.python.org/3/whatsnew/3.12.html#ensurepip


tobias-urdin (approved):
Thanks! We should probably add some Python 3.12 testing soon.

PR #1429, opened 2024-10-23
Remove CI image build hacks

Callum027:
Fix CI image build issues that made certain hacks necessary for things to work, and remove hacks that are no longer required with the new Ubuntu 24.04-based CI image.

* Use the Ubuntu-packaged version of Tox, Setuptools and and all other packages installed outside of the test environment. This is necessary to make dependency conflicts less likely due to setting `sitepackages` to `True` in Tox to pass through `python3-rados` to the test environment.
* Remove the hack for deleting the distro-managed version of `pyparsing`, as the new version available in Ubuntu 24.04 is compatible.
* Remove the hack for deleting the distro-packaged versions of the `jaraco.*` series of packages. The reason why this started causing problems is [Setuptools v71.0](https://setuptools.pypa.io/en/latest/history.html#v71-0-0) onwards changed behaviour to preferring external dependencies instead of the bundled versions if they are installed, even if they are not compatible. The `ceph-mgr` package installs the distro-managed version of `jaraco.text` (among others), causing the conflict.
* Remove the patch for the distro-managed version of `werkzeug`, as the new version now adds the line for deferred annotation validation.
* Don't set `VIRTUALENV_SETUPTOOLS=bundle` explicitly, as this is the default behaviour in virtualenvs created by Tox. `setuptools` was (and is) actually being installed, it was just erroring out on import due to the incompatible version of `jaraco.text` being found. `pip`'s error handling was hiding the real cause of the failure.

### DIFF
--- a/.github/workflows/gnocchi.yml
+++ b/.github/workflows/gnocchi.yml
@@ -167,12 +167,18 @@ jobs:
           push: false
           tags: ghcr.io/gnocchixyz/ci:latest
         if: steps.changes.outputs.ci_image == 'true'
+      # NOTE(callumdickinson): SETUPTOOLS_USE_DISTUTILS needs to be set to 'stdlib'
+      # for Debian-packaged setuptools to work correctly on Python 3.9 and 3.11.
+      # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1003252
       - name: Set env vars
         run: |
+          export DOCKER_ENV_OPTS=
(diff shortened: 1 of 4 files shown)


Callum027:
Our internal CI testing pipeline using a similar build image to what the public Gnocchi pipeline uses revealed the root cause of some of the problems that required hacks to be applied to work around.

I've now made a draft PR to address them upstream.

chungg (approved):

PR #1437, opened 2025-01-16
Fix running functional tests with NumPy 2.2

Callum027:
`testcase` performs an `if not` check on annotations passed in asserts.

From NumPy 2.2 this raises an exception about ambiguous falseyness [1], so convert the array to a string when passing it as an annotation to get around this.

[1]: https://numpy.org/devdocs/release/2.2.0-notes.html#expired-deprecations


tobias-urdin (approved):

tobias-urdin:

PR #1439, opened 2025-01-16
Increase string attribute max length to 256 chars

Callum027:
Increase the maximum supported length for string type resource attributes from 255 to 256 characters, to allow creating string attributes on resources that accept values of such length (e.g. Swift container names).

### DIFF
--- a/gnocchi/resource_type.py
+++ b/gnocchi/resource_type.py
@@ -141,9 +141,9 @@ class StringSchema(CommonAttributeSchema):
 
     meta_schema_ext = {
         voluptuous.Required('min_length', default=0):
-        voluptuous.All(int, voluptuous.Range(min=0, max=255)),
-        voluptuous.Required('max_length', default=255):
-        voluptuous.All(int, voluptuous.Range(min=1, max=255))
+        voluptuous.All(int, voluptuous.Range(min=0, max=256)),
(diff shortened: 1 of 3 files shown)


Callum027:
Requires https://github.com/gnocchixyz/gnocchi/pull/1437 for the build to pass.

No database migrations are required because this does not modify existing attributes. It only increases the maximum supported length for **new** string-type attributes to 256 characters.

Callum027:
Tested in a dev environment with a custom resource type.

```
(ansible) root@callumdickinson-dev-ops1:~# openstack metric resource-type show volume_backup
+-------------------------+-----------------------------------------------------------+
| Field                   | Value                                                     |
+-------------------------+-----------------------------------------------------------+
| attributes/container    | max_length=256, min_length=0, required=True, type=string  |
| attributes/display_name | max_length=255, min_length=0, required=True, type=string  |
| attributes/snapshot_id  | max_length=255, min_length=0, required=False, type=string |
| attributes/volume_id    | max_length=255, min_length=0, required=True, type=string  |
| name                    | volume_backup                                             |
| state                   | active                                                    |
+-------------------------+-----------------------------------------------------------+
```

tobias-urdin (approved):

PR #1440, opened 2025-02-11
Fix listing metrics as non-admin without creator policy defined

Callum027:
When using a minimal RBAC policy that only allows either access either to a cloud admin or a user that is part of the project a metric's resource is owned by, e.g.

```yaml
"list metric": "role:admin or project_id:%(resource.project_id)s"
```

There is an issue where `KeystoneAuthHelper.get_metric_policy_filter` will reject access to the metrics because from its perspective there are no applicable policies that the user has permission under.

This patch fixes that by adding an option to `get_metric_policy_filter` that allows an empty policy filter to be returned if an enforcement check for `resource.project_id` passes. This option is set to `False` by default, and only enabled where necessary.

This allows the caller to use `KeystoneAuthHelper.get_resource_policy_filter` to add project filters to the metric query based on the resource the metric is associated with.

### DIFF
--- a/gnocchi/rest/auth_helper.py
+++ b/gnocchi/rest/auth_helper.py
@@ -94,7 +94,9 @@ class KeystoneAuthHelper(object):
             return {"or": policy_filter}
 
     @staticmethod
-    def get_metric_policy_filter(request, rule):
+    def get_metric_policy_filter(request,
+                                 rule,
+                                 allow_resource_project_id=False):
         try:
             # Check if the policy allows the user to list any metric
             api.enforce(rule, {})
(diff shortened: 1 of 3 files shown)


gnocchi/rest/auth_helper.py
```diff
@@ -113,6 +115,26 @@ def get_metric_policy_filter(request, rule):
                 policy_filter.append(
                     {"like": {"creator": "%:" + project_id}})
 
+            resource_project_id_allowed = False
+            try:
+                # Check if the policy allows the user to list metrics linked
+                # to their project via a resource
+                api.enforce(rule, {"resource.project_id": project_id})
+            except webob.exc.HTTPForbidden:
+                pass
```
rafaelweingartner, line 124: What about a DEBUG log level here instead of just "ignoring" the exception?

Callum027: Thanks for the comment. I think there's scope here to add DEBUG logs for both cases where project-based enforcement checks fail here, so I've added them.

rafaelweingartner: Awesome!! Thanks!

It has been so many times when I am debugging OpenStack components and I have to add LOGs to enable troubleshooting. This kind of approach can really help operators that are not python coders.

rafaelweingartner (approved):

PR #1441, opened 2025-02-11
Fix metric aggregates as non-admin

Callum027:
When using the aggregates API with a metric ID, the RBAC policy `project_id:%(resource.project_id)s"` fails to authenticate a non-admin user that is part of the project the metric is also part of via its resource relationship.

This patch fixes two issues preventing this from working:

* Set `details=True` when listing specified metrics to ensure the joined load with the `Metric` and `Resource` models is done.
* Add support for processing oslo.db SQLAlchemy models in `gnocchi.rest.api.flatten_dict_to_keypairs` to make sure nested attributes are added to the list of targets passed to the policy enforcer.

### DIFF
--- a/gnocchi/rest/aggregates/api.py
+++ b/gnocchi/rest/aggregates/api.py
@@ -612,7 +612,8 @@ class AggregatesController(rest.RestController):
                                 "detail": references})
 
             metrics = pecan.request.indexer.list_metrics(
-                attribute_filter={"in": {"id": metric_ids}})
+                attribute_filter={"in": {"id": metric_ids}},
+                details=True)
             missing_metric_ids = (set(metric_ids)
                                   - set(str(m.id) for m in metrics))
             if missing_metric_ids:
--- a/gnocchi/rest/api.py
+++ b/gnocchi/rest/api.py
@@ -21,6 +21,7 @@ import logging
 import operator
 import uuid
 
+from collections import abc
 import jsonpatch
 from oslo_utils import strutils
 import pecan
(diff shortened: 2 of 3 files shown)


Callum027:
Seems like the GitHub Actions runners are having issues.

```
       Exception in thread Thread-2 (thread_compiler):
        File "/usr/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
          self.run()
        File "/usr/lib/python3.12/threading.py", line 1010, in run
          self._target(*self._args, **self._kwargs)
        File "/tmp/pip-install-nqpvtzih/uwsgi_8ef73bf3d5474636b57bf345332b6745/uwsgiconfig.py", line 116, in thread_compiler
          ret = subprocess.call(cmdline, shell=True)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/usr/lib/python3.12/subprocess.py", line 389, in call
          with Popen(*popenargs, **kwargs) as p:
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/usr/lib/python3.12/subprocess.py", line 1026, in __init__
          self._execute_child(args, executable, preexec_fn, close_fds,
        File "/usr/lib/python3.12/subprocess.py", line 1955, in _execute_child
          raise child_exception_type(errno_num, err_msg, err_filename)
      OSError: [Errno 14] Bad address: '/bin/sh'
```

rafaelweingartner (approved):

tobias-urdin (approved):

PR #1443, opened 2025-02-14
Run metricd more frequently in functional tests

Callum027:
The metricd thread in the functional tests waits for 100ms between runs.

This is actually quite a long time to wait, as the functional tests can run much faster than that. This results in a race condition where earlier tests submit measures, and they are not actually processed by the time the test that needs them is run, and fails because they are not done (and there's no waits or retries or anything).

Tighten up the metricd thread loop so that it waits only 10ms between runs. This will increase load but the more frequent runs should be good enough for the tests to pass.


Callum027:
Didn't seem to fix the issue as I'd hoped. I'll try and spend some time next week investigating this.

Callum027:
Closing, working solution provided in PR #1386

PR #1445, opened 2025-03-13
Support service role

tobias-urdin:
This adds support for the service role that is
used in the Secure RBAC community goal in OpenStack.

This is matched against what API calls that Ceilometer
and Aodh does against the Gnocchi API.


kajinamit (approved):
LGTM. This is a good step toward switching ceilometer/aodh credential from admin to service, though I expect there are number of problems we have to resolve to make sure that ALL API queries from aodh and ceilometer are allowed...

tobias-urdin:

PR #1450, opened 2025-05-07
Gnocchi deleted resources cleanup

rafaelweingartner:
After the development of improvements [1] and [2], we move with this issue to the last step of the improvement work regarding resource's metadata management in Gnocchi.

The idea is to create an option in Gnocchi to automatically clean up deleted resources after a given period of time (configured by the operator). For instance, the idea is that somebody can configure Gnocchi to automatically remove resources marked as deleted after 12 months or 24 months, and so on.

Therefore, with this feature, Gnocchi would be able to auto-clean itself; thus, this can have a good impact on the aggregates API, which manipulates the "resource" table.

[1] https://github.com/gnocchixyz/gnocchi/pull/1386
[2] https://github.com/gnocchixyz/gnocchi/pull/1385

PR #1454, opened 2025-05-13
Change upgrade test validation in Main

rafaelweingartner:
As discussed in https://github.com/gnocchixyz/gnocchi/pull/1450#issuecomment-2876210849, we will change the upgrade tests to check against stable/4.6. The idea is to remove/deprecate stable/4.5 branch.

### DIFF
--- a/gnocchi/carbonara.py
+++ b/gnocchi/carbonara.py
@@ -195,8 +195,8 @@ class GroupedTimeSeries(object):
         ordered = numpy.lexsort((self._ts['values'], self.indexes))
         min_pos = numpy.cumsum(self.counts) - self.counts
         real_pos = min_pos + (self.counts - 1) * (q / 100)
-        floor_pos = numpy.floor(real_pos).astype(numpy.integer, copy=False)
-        ceil_pos = numpy.ceil(real_pos).astype(numpy.integer, copy=False)
+        floor_pos = numpy.floor(real_pos).astype(numpy.int64, copy=False)
+        ceil_pos = numpy.ceil(real_pos).astype(numpy.int64, copy=False)
         values = (
             self._ts['values'][ordered][floor_pos] * (ceil_pos - real_pos) +
             self._ts['values'][ordered][ceil_pos] * (real_pos - floor_pos))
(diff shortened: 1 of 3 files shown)


rafaelweingartner:
@tobias-urdin and @Callum027  I did the update as we discussed in https://github.com/gnocchixyz/gnocchi/pull/1450#issuecomment-2876210849. I had to pin oslo.middleware due to the issues of Python not being able to load it, and also gabbi because in its last update, the function tests seem to be broken.

I tried understanding the issues regarding the latest versions of gabbit and oslo.middleware, but I have no clues on what we need to change to make the current code base work with them.

Callum027:
@rafaelweingartner The only issue I see with this is pinning `oslo.middleware` makes Gnocchi incompatible with OpenStack `master` upper constraints. Otherwise it looks fine to me.

https://releases.openstack.org/constraints/upper/master

rafaelweingartner:
> @rafaelweingartner The only issue I see with this is pinning `oslo.middleware` makes Gnocchi incompatible with OpenStack `master` upper constraints. Otherwise it looks fine to me.
> 
> https://releases.openstack.org/constraints/upper/master

Yes, but at the same time I did not understand why the newer version is incompatible. It is not a code issue, it seems to be an issue with the way it was packaged. Have you seen something similar?

Callum027:
> > @rafaelweingartner The only issue I see with this is pinning `oslo.middleware` makes Gnocchi incompatible with OpenStack `master` upper constraints. Otherwise it looks fine to me.
> > https://releases.openstack.org/constraints/upper/master
> 
> Yes, but at the same time I did not understand why the newer version is incompatible. It is not a code issue, it seems to be an issue with the way it was packaged. Have you seen something similar?

I haven't been able to reproduce the issue with `oslo.middleware` you're describing. What kind of error are you facing? Do you have a link to a failed pipeline I can look at?

rafaelweingartner:
This one https://github.com/gnocchixyz/gnocchi/actions/runs/15006322377/job/42165727461, but the logs are not available there because they are hidden.

rafaelweingartner:
@Callum027, @jd, and @tobias-urdin what are your opinions here?

Can we get this one merged?

jd (approved):

tobias-urdin:
If we merge this won't we risk breaking master branches in OpenStack CI? Can't we solve the underlying cause for the `oslo.middleware` and `gabbi` support for newer versions instead?

Callum027:
I build a CI container and ran the tests locally with only `gabbi` pinned as in this PR. While there are other unrelated test failures that need addressing, packages installed fine and the tests ran without `oslo.middleware` pinned, so I think we can remove that pin and thing should work fine.

Callum027:
@rafaelweingartner @tobias-urdin @jd I have created https://github.com/gnocchixyz/gnocchi/pull/1458 to fix things enough that the CI passes without pinning `oslo.middleware`.

PR #1458, opened 2025-06-26
Fix NumPy 2.3 support and test runs

Callum027:
* Change a number of dtype conversions from`np.integer` to the specific type `np.int64`. This fixes NumPy 2.3 compatibility.
* Pin Gabbi to an older version, as the Gnocchi test suite is currently incompatible with Gabbi 4.0.
* Change the "upgrade" test jobs to test upgrades from Gnocchi 4.6 instead of 4.5, as 4.6 is the current latest stable version and `master` has advanced well beyond 4.6 with new database migrations.
* Disable broken tests that cannot easily be fixed at the moment.
* Fix warnings due to use of deprecated `pip` URL package references and extra requires that do not exist.
* Add more verbose output to `run-upgrade-tests.sh` to help with troubleshooting failures.


tobias-urdin (dismissed):
LGTM, nothing but testing infrastructure is changed, looks like `numpy.int64` has been around for years so should be safe.

I've created the issue https://github.com/cdent/gabbi/issues/343 in the gabbi project to see if we can get some feedback from cdent on how to support gabbi>=4.0

tobias-urdin:
@Callum027 can you try updating the `setup_app()` function as mentioned here https://github.com/cdent/gabbi/issues/343#issuecomment-3013153822 and unpin `gabbi`, with some luck we can solve that as well in this PR, thanks! :)

kajinamit:
I'd prefer to have a separate commit to pin gabbi, so that we can easily revert it later.

Callum027:
> I'd prefer to have at least one separate commit to pin gabbi in the chain, instead of including it within the whole large commit, so that we can easily revert/backport it later. (or maybe we should separate commit to each fix item)

Hi @kajinamit, thanks for the response. I have a few reasons for why I don't think this is necessary:

* While it's unknown how the issue with Gabbi can be fixed, given the comments from the upstream developer it's likely the solution lies in fixing Gnocchi's test suite to work better.
* Even if a patch is made upstream to allow Gnocchi to work again, we would need to add excludes for the versions that don't work to the requirements (e.g. `!=4.0.0,!=4.1.0,!=4.1.1`), so that wouldn't be a straight reversion either.
* I can split these into separate commits, but unless the admin/reviewer configure the GitHub repository (and the PR) in specific ways, this results in either the commits being squashed into a single commit, or a merge commit with the commits in the PR branching off with that, both of which cause issues with reverting and cherry picking.
* I just don't think it's worth the hassle for the (IMHO low) chance that any of these will get reverted, since most of these changes are required just to get the CI to pass.

rafaelweingartner:
> I'd prefer to have at least one separate commit to pin gabbi in the chain, instead of including it within the whole large commit, so that we can easily revert/backport it later. (or maybe we should separate commit to each fix item)

I agree with @Callum027 here. I do not think that the "unpin" of gabbi will be that simple and easy. BTW, does anybody understand what was modified there that impact Gnocchi. I read the thread in their repository, but I did not understand clearly what was done there.

Callum027:
> I agree with @Callum027 here. I do not think that the "unpin" of gabbi will be that simple and easy. BTW, does anybody understand what was modified there that impact Gnocchi. I read the thread in their repository, but I did not understand clearly what was done there.

To give a short summary, with Gabbi 4.0 they switched from using `urllib3` to `httpx`, and an architecture where a subprocess is started for the HTTP server. Gnocchi's test suite is relying on some hackery with a global variable to be able start a WSGI app with some pre-configuration done, which relied on the previous behaviour and is now broken.

The issue is there appears no easy way to resolve this without some kind of patch on the Gabbi side to accommodate the behaviour that Gnocchi relies on, or refactoring the Gnocchi test suite so that it works more closely with how Gabbi 4.0 expects things to be done.

Callum027:
> @Callum027 and @tobias-urdin I am a bit confused here. I updated my patch: #1454, and now it is working without pinning oslo.middleware. Do you guys know if something changed there?

I wasn't able to reproduce your issue with oslo.middleware (nor was I actually able to see any issues regarding oslo.middleware in the other PD), but for now I think we're fine not worrying about it.

For the sake of unblocking other PRs, @tobias-urdin, would we be able to merge this soon?

PR #1460, opened 2025-06-30
Fix ended_at field normalization

rafaelweingartner:
When Gnocchi is executing the "ended_at" field normalization, which is executed if all metrics of a resource stops receiving measures, there is a bug that will cause Gnocchi to not generate properly the resource revisions, which in turn can affect the aggregates API, when filtering by resource type attributes.

This PR introduces the code changes to handle this issue.


rafaelweingartner:
@Callum027 and @tobias-urdin this one is important to get merged. Otherwise, users will have issues when using aggregates API for resources that were marked as "finished" after they cease receiving measurements in the metrics.

PR #1493, opened 2026-03-01
Get rid of pkg_resources

kajinamit:
... because it was removed in Python 3.12 [1].

[1] https://docs.python.org/3/whatsnew/3.12.html#ensurepip

(cherry picked from commit 7c90f348d7aba9155e50da4886f93c45c8d9f381)

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -10,7 +10,7 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import pkg_resources
+import importlib.metadata
 
 # -- General configuration -----------------------------------------------------
 
@@ -38,7 +38,7 @@ copyright = 'The Gnocchi Developers'
 # built documents.
 #
 # The short X.Y version.
-release = pkg_resources.get_distribution('gnocchi').version
+release = importlib.metadata.version('gnocchi')
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
--- a/gnocchi/__init__.py
+++ b/gnocchi/__init__.py
@@ -12,10 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import pkg_resources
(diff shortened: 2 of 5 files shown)


doc/source/conf.py
```diff
@@ -10,7 +10,7 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import pkg_resources
+import importlib.metadata
```
Callum027, line 13: For Python <3.10, this will need to be something like:

```python
try:
    import importlib.metadata as importlib_metadata
except ImportError:
    import importlib_metadata
```

gnocchi/__init__.py
```diff
@@ -12,10 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import pkg_resources
+import importlib.metadata
```
Callum027, line 15: For Python <3.10, this will need to be something like:

```python
try:
    import importlib.metadata as importlib_metadata
except ImportError:
    import importlib_metadata
```

doc/source/conf.py
```diff
@@ -38,7 +38,7 @@
 # built documents.
 #
 # The short X.Y version.
-release = pkg_resources.get_distribution('gnocchi').version
+release = importlib.metadata.version('gnocchi')
```
Callum027, line 41:
```python
release = importlib_metadata.version('gnocchi')
```

gnocchi/__init__.py
```diff
@@ -12,10 +12,10 @@
...
 
 try:
-    __version__ = pkg_resources.get_distribution(__name__).version
-except pkg_resources.DistributionNotFound:
+    __version__ = importlib.metadata.version('gnocchi')
```
Callum027, line 18:
````python
    __version__ = importlib_metadata.version('gnocchi')
````

Callum027 (changes requested):

PR #1495, opened 2026-03-28
Improve restoring resource log

rafaelweingartner:
While trouebleshooting a situation with a resource being "undeleted", we noticed that more logs were needed to failitate the understanding of the situation. Therefore, this PR is proposing a improve for such log message.


tobias-urdin (approved):

rafaelweingartner:
Thanks @tobias-urdin!

You have helped us a lot here in Gnocchi :)

PR #1500, opened 2026-04-27
Pin cotyledon<2.2.0 in testing

tobias-urdin:
The cotyledon 2.2.0 release introduced typing with ParamSpec but did not correcly update it's lower
Python version to 3.10 where ParamSpec was introduced.

This pins cotyledon <2.2.0 so that our Python 3.9
testing works, this can be removed when we drop
Python 3.9 testing.

We also backport some changes to testing to fix testtools changes
and a SQLAlchemy issue.

### DIFF
--- a/gnocchi/tests/base.py
+++ b/gnocchi/tests/base.py
@@ -19,6 +19,7 @@ import logging
 import os
 import subprocess
 import threading
+import unittest
 import uuid
 
 import daiquiri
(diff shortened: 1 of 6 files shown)


tobias-urdin:
Backported fixes for changes in testtools and for the SQLALchemy fix, everything is just touching testing code so shouldn't be an issue, same with the original $topic of this patch.

PR #1502, opened 2026-05-07
Only load the metrics used in the aggregates API

rafaelweingartner:
Gnocchi aggregates API can struggle to process the requests when someone is, for instance, searching for all resources in a project, and if this project has thousands of resources.

The highest cost is at the execution of "all_resources = session.scalars(q).unique().all()" in the "list_resources" method. Because of the data structure used and the SQLAlchemy joins, a single resource entry, when joined with the metrics table, can multiply the output by a factor that represents the number of metrics the resource has. Thus, we have seen a 10k resource output (of the final response in the API) become more than 100k entries in the output, which will then be instantiated as a Python object, and afterwards filtered by the "unique" operation. The problem is that this process is costly.

The proposal here is to only load the metrics used in the aggregates API request operation. By doing this, we can reduce the pressure on both the DB and Gnocchi API itself.

### DIFF
--- a/gnocchi/indexer/sqlalchemy.py
+++ b/gnocchi/indexer/sqlalchemy.py
@@ -1165,7 +1165,7 @@ class SQLAlchemyIndexer(indexer.IndexerDriver):
         return attribute_filters_to_use
 
     def _get_history_result_mapper(self, session, resource_type,
-                                   attribute_filter=None):
+                                   attribute_filter=None, metrics_to_load=None):
 
         mappers = self._resource_type_to_mappers(session, resource_type)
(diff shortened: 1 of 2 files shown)


tobias-urdin:
Does this have any API impact (behavior change) or is it just a performance improvement?

rafaelweingartner:
> Does this have any API impact (behavior change) or is it just a performance improvement?

Just the performance improvement. We are now only loading the metrics used in the API request, instead of loading all of the metrics of the resources, and then just using the metric in the API request.

tobias-urdin (approved):

PR #1503, opened 2026-07-15
Obfuscate/silence API configuration logging

Callum027:
Whenever a new API worker is started, a dump of the entire raw configuration is output to logging on the `INFO` log level, including sensitive information such as passwords.

As this is a security risk and too verbose for regular `INFO` logging, replace this with `log_opt_values` (which obfuscates secrets) outputting to the `DEBUG` log level.


Callum027:
I think the test failures are unrelated.

